PR #11, opened 2023-03-02
Add URL and URLSearchParams Polyfills

stefanlucas:
Initial implementation for the URL and URLSearchParams polyfills.


Polyfills/URLSearchParams/Source/URLSearchParams.cpp
```diff
@@ -0,0 +1,165 @@
...
+            resultString += value;
+            resultString += "&";
+        } 
+
+        // check last value of string 
+        char ch = resultString.back();
+        if (ch == '&')
+        { 
+            resultString.pop_back();
+        };
```
bghgary, line 135: I think it might be clearer if we append the first key/value outside the for loop and append the `&` first inside the first loop so that you don't need to do this odd check at the end. Or do it all in the for loop and check if it's the first element to know whether to append `?` or `&`.

Polyfills/URLSearchParams/Source/URLSearchParams.h
```diff
@@ -0,0 +1,31 @@
+#pragma once
+
+#include <Babylon/JsRuntimeScheduler.h>
+#include <napi/napi.h>
+#include <unordered_map>
+#include <vector>
+#include "../Include/Babylon/Polyfills/URLSearchParams.h"
```
bghgary, line 7: Why do we need to include this?

Tests/UnitTests/Scripts/tests.js
```diff
@@ -211,6 +211,241 @@ describe("clearTimeout", function () {
...
+            'developer.mozilla.org',          // hostname
+            'https://developer.mozilla.org',  // origin
+            '',                               // pathname
+            ''                                // search
+        );
+    });
+
+    it("should load URL with pathname (no search)", function () {
+        // Augment URL with pathname (no search)
+        url = new URL(baseUrl + '/en-US/docs');
```
bghgary, line 250: FYI: We can also use string interpolation syntax with backquotes:
```suggestion
        url = new URL(`${baseUrl}/en-US/docs`);
```

Tests/UnitTests/Scripts/tests.js
```diff
@@ -211,6 +211,241 @@ describe("clearTimeout", function () {
...
+})
+
+// URLSearchParams
+describe("URLSearchParams", function () {
+
+    // -------------------------------- URLSearchParams Get --------------------------------
+
+    it("should retrieve null from empty searchParams", function () {
+        // Get Empty
+        let params = new URLSearchParams('');
```
bghgary, line 351: Prefer `const` over `let` where possible. Multiple places.

```suggestion
        const params = new URLSearchParams('');
```

bghgary (commented):
Looks good overall. Lots of small issues. There are some TODOs in the code that we should either remove or file issues to track.

Polyfills/URL/Source/URL.cpp
```diff
@@ -0,0 +1,144 @@
...
+
+    void URL::Initialize(Napi::Env env)
+    {
+        Napi::HandleScope scope{env};
+
+        Napi::Function func = DefineClass(
+            env,
+            JS_URL_CONSTRUCTOR_NAME,
+            {
+                StaticMethod("createObjectURL", &URL::CreateObjectURL),
```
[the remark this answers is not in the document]
stefanlucas: Removed

Polyfills/URL/Source/URL.cpp
```diff
@@ -0,0 +1,144 @@
...
+    Napi::Value URL::GetSearchParams(const Napi::CallbackInfo&)
+    {
+        return m_searchParamsReference.Value();
+    }
+
+    //TODO replace stub with actual implementation
+    Napi::Value URL::CreateObjectURL(const Napi::CallbackInfo& info)
+    {
+        return Napi::Value::From(info.Env(), "");
+    }
```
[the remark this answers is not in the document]
stefanlucas: Removed, since for now we will not be using it

Polyfills/URL/Source/URL.cpp
```diff
@@ -0,0 +1,144 @@
...
+
+    URL::URL(const Napi::CallbackInfo& info)
+        : Napi::ObjectWrap<URL>{info}
+        , m_runtimeScheduler{JsRuntime::GetFromJavaScript(info.Env())}
+    {
+        if (!info.Length())
+            return;
```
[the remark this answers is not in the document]
stefanlucas: correct, we don't support base yet

Polyfills/URL/Source/URL.cpp
```diff
@@ -0,0 +1,144 @@
...
+        m_searchParams = info.Env().Global().Get(URLSearchParams::JS_URL_SEARCH_PARAMS_CONSTRUCTOR_NAME).As<Napi::Function>().New({Napi::Value::From(info.Env(), m_search) });
+        m_searchParamsReference = Napi::Persistent(m_searchParams);
+
+        // Get URL Domain
+        const size_t start = m_href.find("//");
+        const size_t originStart = m_href.find("h");
+        const size_t index = m_href.find_first_of("/", start + 2);
+
+        m_origin = m_href.substr(originStart, index);
+        m_hostname = m_origin.substr(start + 2, index);
```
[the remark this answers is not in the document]
stefanlucas: yes, Zach implemented only what was necessary for our use case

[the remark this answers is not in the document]
stefanlucas: yes, Zach implemented only what was necessary for our use case

Polyfills/URLSearchParams/Source/URLSearchParams.cpp
```diff
@@ -0,0 +1,165 @@
...
+
+        JsRuntime::NativeObject::GetFromJavaScript(env).Set(JS_URL_SEARCH_PARAMS_CONSTRUCTOR_NAME, func);
+    }
+
+    URLSearchParams::URLSearchParams(const Napi::CallbackInfo& info)
+        : Napi::ObjectWrap<URLSearchParams>{info}
+        , m_runtimeScheduler{JsRuntime::GetFromJavaScript(info.Env())}
+    {
+        std::string queryStr = info[0].As<Napi::String>();
+        parseInputQueryStr(queryStr);
```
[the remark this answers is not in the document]
stefanlucas: Yes, we are supporting

Polyfills/URLSearchParams/Source/URLSearchParams.h
stefanlucas: Removed

.gitignore
```diff
@@ -1 +1,3 @@
 /Build
+.vs
+out
```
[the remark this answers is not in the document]
stefanlucas: removed

Polyfills/URLSearchParams/Source/URLSearchParams.h
```diff
@@ -0,0 +1,31 @@
...
+
+    private:
+        void Set(const Napi::CallbackInfo& info);
+        Napi::Value Has(const Napi::CallbackInfo& info);
+        Napi::Value Get(const Napi::CallbackInfo& info);
+        void parseKeyVal(const std::string& subStr);
+        void setKeyVal(const std::string& key, const std::string& value);
+        void parseInputQueryStr(std::string& queryStr);
+
+        JsRuntimeScheduler m_runtimeScheduler;
```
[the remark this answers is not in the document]
stefanlucas: removed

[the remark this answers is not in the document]
stefanlucas: using on URLSearchParams constructor

Polyfills/URL/Source/URL.cpp
```diff
@@ -92,32 +86,27 @@ namespace Babylon::Polyfills::Internal
...
-    {
-        return Napi::Value::From(info.Env(), "");
-    }
-
     URL::URL(const Napi::CallbackInfo& info)
         : Napi::ObjectWrap<URL>{info}
         , m_runtimeScheduler{JsRuntime::GetFromJavaScript(info.Env())}
     {
-        if (!info.Length())
+        if (!info.Length()) {
```
bghgary, line 93: nit: bunch of places
```suggestion
        if (!info.Length())
        {
```

Polyfills/URL/Source/URL.cpp
```diff
@@ -92,32 +86,27 @@ namespace Babylon::Polyfills::Internal
...
+        if (qIndex != std::string::npos) {
             m_search = m_href.substr(qIndex, m_href.size() - qIndex);
-        else
-            m_search = "";
+        }
+
 
         // get UrlSearchParams object 
-        m_searchParams = info.Env().Global().Get(URLSearchParams::JS_URL_SEARCH_PARAMS_CONSTRUCTOR_NAME).As<Napi::Function>().New({Napi::Value::From(info.Env(), m_search) });
+        const Napi::Object m_searchParams = info.Env().Global().Get(URLSearchParams::JS_URL_SEARCH_PARAMS_CONSTRUCTOR_NAME).As<Napi::Function>().New({Napi::Value::From(info.Env(), m_search) });
```
bghgary, line 109: Local variables should not have `m_` prefix

```suggestion
        const Napi::Object searchParams = info.Env().Global().Get(URLSearchParams::JS_URL_SEARCH_PARAMS_CONSTRUCTOR_NAME).As<Napi::Function>().New({Napi::Value::From(info.Env(), m_search) });
```

Polyfills/URL/Source/URL.cpp
```diff
@@ -0,0 +1,133 @@
...
+    Napi::Value URL::GetHref(const Napi::CallbackInfo&)
+    {
+        std::string resultHref;
+        resultHref += m_origin;
+        resultHref += m_pathname;
+
+        std::string allParams = GetSearchQuery();  
+        resultHref += allParams;
+
+        return Napi::Value::From(Env(), resultHref);
```
bghgary, line 54: I don't see std::stringstream being used. See previous comment on this.

Polyfills/URL/Source/URL.cpp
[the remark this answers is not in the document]
bghgary: This is okay for now, but it probably makes more sense to forward this to UrlLib and use platform specific functions to parse the URL and get the parts. Please add a comment and issue to track.

Polyfills/URL/Source/URL.h
```diff
@@ -0,0 +1,49 @@
...
+        Napi::Value GetOrigin(const Napi::CallbackInfo& info);
+        Napi::Value GetPathname(const Napi::CallbackInfo& info);
+        Napi::Value GetHostname(const Napi::CallbackInfo& info);
+
+        std::string GetSearchQuery();
+        Napi::Value GetSearchParams(const Napi::CallbackInfo& info);
+
+        static Napi::Value CreateObjectURL(const Napi::CallbackInfo& info);
+
+        JsRuntimeScheduler m_runtimeScheduler;
```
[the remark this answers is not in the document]
bghgary: I'm not sure what happened. This is marked as resolved but the variable is still here?

Polyfills/URL/Source/URL.cpp
```diff
@@ -0,0 +1,133 @@
...
+    }
+
+    Napi::Value URL::GetSearchParams(const Napi::CallbackInfo&)
+    {
+        return m_searchParamsReference.Value();
+    }
+
+    URL::URL(const Napi::CallbackInfo& info)
+        : Napi::ObjectWrap<URL>{info}
+        , m_runtimeScheduler{JsRuntime::GetFromJavaScript(info.Env())}
```
bghgary, line 91: Unused?

Polyfills/URL/Source/URLSearchParams.cpp
```diff
@@ -0,0 +1,161 @@
...
+        {
+            env.Global().Set(JS_URL_SEARCH_PARAMS_CONSTRUCTOR_NAME, func);
+        }
+
+        JsRuntime::NativeObject::GetFromJavaScript(env).Set(JS_URL_SEARCH_PARAMS_CONSTRUCTOR_NAME, func);
+    }
+
+    URLSearchParams::URLSearchParams(const Napi::CallbackInfo& info)
+        : Napi::ObjectWrap<URLSearchParams>{info}
+        , m_runtimeScheduler{JsRuntime::GetFromJavaScript(info.Env())}
```
bghgary, line 31: Unused?

Polyfills/URL/Source/URLSearchParams.h
```diff
@@ -0,0 +1,29 @@
...
+        std::string GetAllParams();
+
+    private:
+        void Set(const Napi::CallbackInfo& info);
+        Napi::Value Has(const Napi::CallbackInfo& info);
+        Napi::Value Get(const Napi::CallbackInfo& info);
+        void parseKeyVal(const std::string& subStr);
+        void parseInputQueryStr(const std::string& queryStr);
+
+        JsRuntimeScheduler m_runtimeScheduler;
```
bghgary, line 25: Unused?

Polyfills/URL/Source/URLSearchParams.cpp
```diff
@@ -0,0 +1,161 @@
...
+        if (element == m_paramsMap.end())
+        {
+            return Env().Null();
+        }
+        return Napi::Value::From(Env(), element->second);
+    }
+
+    std::string URLSearchParams::GetAllParams()
+    {
+        std::stringstream resultStringStream("");
```
bghgary, line 97:
```suggestion
        std::stringstream resultStringStream;
```

Polyfills/URL/Source/URLSearchParams.cpp
```diff
@@ -0,0 +1,161 @@
...
+
+        if (m_paramsVector.empty())
+        {
+            return resultStringStream.str();
+        }
```
bghgary, line 102: This can be removed. The for loop will already exit immediately if empty.

Polyfills/URL/Source/URLSearchParams.cpp
```diff
@@ -0,0 +1,161 @@
...
+            {
+                resultStringStream << "&";
+            }
+            else 
+            {
+                resultStringStream << "?";
+            }
+
+            std::string key = m_paramsVector[i];
+            std::string value = m_paramsMap[key];
```
bghgary, line 116: Avoid copying

```suggestion
            const std::string& key = m_paramsVector[i];
            const std::string& value = m_paramsMap[key];
```

Polyfills/URL/Source/URLSearchParams.cpp
```diff
@@ -0,0 +1,161 @@
...
+        return resultStringStream.str();
+    }
+
+
+    void URLSearchParams::Set(const Napi::CallbackInfo& info)
+    {
+        if (info.Length() < 2) 
+        {
+            return;
+        }
```
bghgary, line 132: Should this throw instead of do nothing? What does the web browser do?

Polyfills/URL/Source/URLSearchParams.cpp
```diff
@@ -0,0 +1,161 @@
...
+
+        if (m_paramsMap.find(key) == m_paramsMap.end()) 
+        {
+            m_paramsVector.push_back(key);
+            m_paramsMap[key] = value;
+        }
+        else 
+        {
+            m_paramsMap[key] = value;
+        }
```
bghgary, line 145: Try using [insert_or_assign](https://en.cppreference.com/w/cpp/container/unordered_map/insert_or_assign) to eliminate two hash lookups.

Tests/UnitTests/Scripts/tests.js
```diff
@@ -211,6 +211,205 @@ describe("clearTimeout", function () {
...
+        search) {
+        expect(url).to.have.property("hostname", hostname);
+        expect(url).to.have.property("href", href);
+        expect(url).to.have.property("origin", origin);
+        expect(url).to.have.property("pathname", pathname);
+        expect(url).to.have.property("search", search);
+    }
+
+    // Base URLs:
+    const baseUrl = 'https://developer.mozilla.org';
```
bghgary, line 233: Does `httpbin.org` not work for these tests? I would prefer to use this instead of pinging some public server that isn't intended for testing.

Tests/UnitTests/Shared/Shared.cpp
```diff
@@ -33,6 +35,8 @@ int RunTests(Babylon::Polyfills::Console::CallbackT consoleCallback)
         Babylon::Polyfills::XMLHttpRequest::Initialize(env);
         Babylon::Polyfills::Console::Initialize(env, std::move(consoleCallback));
         Babylon::Polyfills::Scheduling::Initialize(env);
+        Babylon::Polyfills::URL::Initialize(env);
+        Babylon::Polyfills::URLSearchParams::Initialize(env);
```
bghgary, line 39: These should be combined. It's one polyfill.

Polyfills/URLSearchParams/Source/URLSearchParams.cpp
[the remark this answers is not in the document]
bghgary: I believe the code will throw an exception right now if there are no arguments.

Polyfills/URL/Source/URL.cpp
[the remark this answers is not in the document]
bghgary: Please add a comment in the code, add this info to the readme and file an issue.

Polyfills/URLSearchParams/Source/URLSearchParams.h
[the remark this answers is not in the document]
bghgary: Yes, but that variable is only used in the constructor. It can be removed. This class has no side effects.

Polyfills/URL/Source/URLSearchParams.cpp
stefanlucas: Removed

Polyfills/URL/Source/URLSearchParams.cpp
stefanlucas: The web browser throws an error, fixed to have the same behavior

Polyfills/URL/Source/URLSearchParams.h
stefanlucas: removed

Tests/UnitTests/Scripts/tests.js
stefanlucas: It works, for now we are making any request to the server, but I`ve changed to httpbin

Polyfills/URL/Source/URL.h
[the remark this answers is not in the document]
stefanlucas: it should be removed on this new review

Polyfills/URL/Source/URL.cpp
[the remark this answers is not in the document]
stefanlucas: [Added](https://github.com/BabylonJS/JsRuntimeHost/issues/13)

Polyfills/URLSearchParams/Source/URLSearchParams.cpp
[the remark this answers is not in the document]
stefanlucas: resolved on the latest commit

Polyfills/URLSearchParams/Source/URLSearchParams.h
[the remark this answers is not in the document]
stefanlucas: removed

.gitignore
```diff
@@ -1 +1 @@
-/Build
+/Build
```
bghgary, line 1: nit: not a big deal at all, but the red icon is here because there is no newline at the end

Polyfills/CMakeLists.txt
```diff
@@ -10,6 +10,10 @@ if(JSRUNTIMEHOST_POLYFILL_XMLHTTPREQUEST)
...
 endif()
 
+if(JSRUNTIMEHOST_POLYFILL_URL)
+    add_subdirectory(URL)
+endif()
+
 if(JSRUNTIMEHOST_POLYFILL_ABORT_CONTROLLER)
     add_subdirectory(AbortController)
-endif()
+endif()
```
bghgary, line 19: nit: ditto for newline

Polyfills/URL/CMakeLists.txt
```diff
@@ -0,0 +1,18 @@
+set(SOURCES 
+    ${SOURCES}
+    "Include/Babylon/Polyfills/URL.h"
+    "Source/URL.cpp"
+    "Source/URL.h"
+    "Source/URLSearchParams.cpp"
+    "Source/URLSearchParams.h"
+    )
```
bghgary, line 8: nit: to be consistent with other files

```suggestion
set(SOURCES 
    ${SOURCES}
    "Include/Babylon/Polyfills/URL.h"
    "Source/URL.cpp"
    "Source/URL.h"
    "Source/URLSearchParams.cpp"
    "Source/URLSearchParams.h")
```

Polyfills/URL/Readme.md
```diff
@@ -0,0 +1,27 @@
...
+
+## Supported:
+
+- [`get`](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams/get)
+- [`set`](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams/set)
+- [`has`](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams/has)
+
+## Not implemented:
+- URLSearchParams constructor is partially implemented, it does not support sequence of pairs and records
+- All other properties/methods not specified above
```
bghgary, line 27: nit: sorry I missed this before

```suggestion
## Supported

- [`get`](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams/get)
- [`set`](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams/set)
- [`has`](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams/has)

## Not implemented
- URLSearchParams constructor is partially implemented, it does not support sequence of pairs and records
- All other properties/methods not specified above
```

Polyfills/URL/Source/URL.h
```diff
@@ -0,0 +1,44 @@
...
+        static constexpr auto JS_URL_CONSTRUCTOR_NAME = "URL";
+
+    public:
+        static void Initialize(Napi::Env env);
+        static URL& GetFromJavaScript(Napi::Env env);
+
+        explicit URL(const Napi::CallbackInfo& info);
+
+    private:
+
```
bghgary, line 21: nit: extra newline (inconsistent with other code)

```suggestion
    private:
```

Polyfills/URL/Source/URL.h
```diff
@@ -0,0 +1,44 @@
...
+        std::unordered_map<std::string, std::vector<Napi::FunctionReference>> m_eventHandlerRefs;
+        std::string m_search;
+        std::string m_href;
+        std::string m_origin;
+        std::string m_pathname;
+        std::string m_hostname;
+        Napi::ObjectReference m_searchParamsReference;
+
+    };
```
bghgary, line 43: nit:

```suggestion
    };
```

Polyfills/URL/Source/URLSearchParams.cpp
```diff
@@ -0,0 +1,155 @@
...
+
+        std::string key = info[0].As<Napi::String>();
+        std::string value = info[1].ToString().Utf8Value();
+
+        if (m_paramsMap.insert_or_assign(key, value).second) 
+        {
+            m_paramsVector.push_back(key);
+        }
+       
+    }
```
bghgary, line 140: nit:

```suggestion
    }
```

Polyfills/URL/Source/URLSearchParams.h
```diff
@@ -0,0 +1,28 @@
...
+        explicit URLSearchParams(const Napi::CallbackInfo& info);
+        static constexpr auto JS_URL_SEARCH_PARAMS_CONSTRUCTOR_NAME = "URLSearchParams";
+        std::string GetAllParams();
+
+    private:
+        void Set(const Napi::CallbackInfo& info);
+        Napi::Value Has(const Napi::CallbackInfo& info);
+        Napi::Value Get(const Napi::CallbackInfo& info);
+        void parseKeyVal(const std::string& subStr);
+        void parseInputQueryStr(const std::string& queryStr);
```
bghgary, line 23: Function names should be upper camel case.

```suggestion
        void ParseKeyVal(const std::string& subStr);
        void ParseInputQueryStr(const std::string& queryStr);
```

Tests/UnitTests/Scripts/tests.js
```diff
@@ -225,6 +225,209 @@ describe("clearTimeout", function () {
...
+    it('should load URL with no pathname / search', function () {
+        // Standard URL (No pathname, no search)
+        let url = new URL(baseUrl);
+        // => 'http://httpbin.org'
+        checkURL(url,
+            'http://httpbin.org',  // href
+            'httpbin.org',          // hostname
+            'http://httpbin.org',  // origin
+            '',                               // pathname
+            ''                                // search
```
bghgary, line 258: nit: comments don't line up

bghgary: One thing you could do to avoid having to comment is to pass in an object instead of multiple parameters. Then you can write the code like this instead:
```javascript
checkUrl(url, {
    href: '...',
    hostname: '...",
    // ...
});
```

Polyfills/URL/Source/URLSearchParams.cpp
```diff
@@ -0,0 +1,155 @@
...
+}
+
+namespace Babylon::Polyfills::URLSearchParams
+{
+    void Initialize(Napi::Env env)
+    {
+        Internal::URLSearchParams::Initialize(env);
+    }
+}
```
bghgary, line 155: Unused?

bghgary (commented):
Almost all nits, getting close.

Polyfills/URL/Source/URLSearchParams.h
bghgary: I still see lower camel case?

stefanlucas: Sorry, didn't save the files before the commit. Fixed now

bghgary (approved):

PR #12, opened 2023-03-06
Lmaskati/test

lmaskati:
This PR is a draft until https://github.com/BabylonJS/UrlLib/pull/1 is merged, after which it needs to be tested with an updated CMake.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,28 +1,11 @@
 cmake_minimum_required(VERSION 3.18)
 
-include(FetchContent)
-
-if(IOS)
-    FetchContent_Declare(
-        ios-cmake
-        GIT_REPOSITORY https://github.com/leetal/ios-cmake.git
-        GIT_TAG        04d91f6675dabb3c97df346a32f6184b0a7ef845)
-
-    message(STATUS "Fetching ios-cmake")
-    FetchContent_MakeAvailable(ios-cmake)
-    message(STATUS "Fetching ios-cmake - done")
-
-    set(CMAKE_TOOLCHAIN_FILE "${ios-cmake_SOURCE_DIR}/ios.toolchain.cmake" CACHE PATH "")
-    set(PLATFORM "OS64COMBINED" CACHE STRING "")
-    set(DEPLOYMENT_TARGET "12" CACHE STRING "")
-endif()
-
-project(JsRuntimeHost)
-
 set_property(GLOBAL PROPERTY USE_FOLDERS ON)
 
 set(CMAKE_CXX_STANDARD 17)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
+
+project(JsRuntimeHost)
 
 # Options
 
@@ -40,7 +23,11 @@ option(JSRUNTIMEHOST_TESTS "Include Babylon Native Tests." ON)
 
 # Dependencies
 
+include(FetchContent)
+
 if(NOT TARGET arcana)
+    include(FetchContent)
+
     FetchContent_Declare(
         arcana
         GIT_REPOSITORY https://github.com/microsoft/arcana.cpp.git
@@ -56,8 +43,8 @@ endif()
 if(NOT TARGET UrlLib AND JSRUNTIMEHOST_POLYFILL_XMLHTTPREQUEST)
     FetchContent_Declare(
         UrlLib
-        GIT_REPOSITORY https://github.com/BabylonJS/UrlLib.git
-        GIT_TAG        6413ba699107d8114ca19e7646638d3c628eb956)
+        GIT_REPOSITORY https://github.com/lmaskati/UrlLib.git
+        GIT_TAG        origin/main)
 
     message(STATUS "Fetching UrlLib")
     FetchContent_MakeAvailable(UrlLib)
@@ -90,6 +77,12 @@ elseif(UNIX)
     set(JSRUNTIMEHOST_PLATFORM "Unix")
 else()
     message(FATAL_ERROR "Unrecognized platform: ${CMAKE_SYSTEM_NAME}")
+endif()
+
+if(APPLE)
+    set(JSRUNTIMEHOST_PLATFORM_IMPL_EXT "mm")
+else()
+    set(JSRUNTIMEHOST_PLATFORM_IMPL_EXT "cpp")
 endif()
 
 add_subdirectory(Core)
--- a/Polyfills/XMLHttpRequest/Source/XMLHttpRequest.cpp
+++ b/Polyfills/XMLHttpRequest/Source/XMLHttpRequest.cpp
@@ -39,11 +39,14 @@ namespace Babylon::Polyfills::Internal
         namespace MethodType
         {
             constexpr const char* Get = "GET";
+            constexpr const char* Post = "POST";
 
             UrlLib::UrlMethod StringToEnum(const std::string& value)
             {
                 if (value == Get)
                     return UrlLib::UrlMethod::Get;
+                else if (value == Post)
+                    return UrlLib::UrlMethod::Post;
 
                 throw std::runtime_error{"Unsupported url method: " + value};
             }
@@ -77,6 +80,8 @@ namespace Babylon::Polyfills::Internal
                 InstanceAccessor("responseType", &XMLHttpRequest::GetResponseType, &XMLHttpRequest::SetResponseType),
                 InstanceAccessor("responseURL", &XMLHttpRequest::GetResponseURL, nullptr),
                 InstanceAccessor("status", &XMLHttpRequest::GetStatus, nullptr),
+                InstanceAccessor("ok", &XMLHttpRequest::Ok, nullptr),
+                InstanceAccessor("headers", &XMLHttpRequest::GetAllResponseHeaders, nullptr),
                 InstanceMethod("getResponseHeader", &XMLHttpRequest::GetResponseHeader),
                 InstanceMethod("addEventListener", &XMLHttpRequest::AddEventListener),
                 InstanceMethod("removeEventListener", &XMLHttpRequest::RemoveEventListener),
@@ -84,6 +89,7 @@ namespace Babylon::Polyfills::Internal
                 InstanceMethod("open", &XMLHttpRequest::Open),
                 InstanceMethod("send", &XMLHttpRequest::Send),
                 InstanceMethod("setRequestHeader", &XMLHttpRequest::SetRequestHeader),
+                InstanceMethod("json", &XMLHttpRequest::Json),
             });
 
         if (env.Global().Get(JS_XML_HTTP_REQUEST_CONSTRUCTOR_NAME).IsUndefined())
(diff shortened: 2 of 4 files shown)


CMakeLists.txt
```diff
@@ -1,29 +1,12 @@
...
-    FetchContent_MakeAvailable(ios-cmake)
-    message(STATUS "Fetching ios-cmake - done")
-
-    set(CMAKE_TOOLCHAIN_FILE "${ios-cmake_SOURCE_DIR}/ios.toolchain.cmake" CACHE PATH "")
-    set(PLATFORM "OS64COMBINED" CACHE STRING "")
-    set(DEPLOYMENT_TARGET "12" CACHE STRING "")
-endif()
-
-project(JsRuntimeHost)
-
```
bghgary, line 21: Bad merge? This shouldn't be deleted or iOS won't work.

CMakeLists.txt
```diff
@@ -92,6 +79,12 @@ else()
     message(FATAL_ERROR "Unrecognized platform: ${CMAKE_SYSTEM_NAME}")
 endif()
 
+if(APPLE)
+    set(JSRUNTIMEHOST_PLATFORM_IMPL_EXT "mm")
+else()
+    set(JSRUNTIMEHOST_PLATFORM_IMPL_EXT "cpp")
+endif()
+
```
bghgary, line 87: Probably bad merge also?

Polyfills/XMLHttpRequest/Source/XMLHttpRequest.cpp
```diff
@@ -77,12 +80,16 @@ namespace Babylon::Polyfills::Internal
                 InstanceAccessor("responseType", &XMLHttpRequest::GetResponseType, &XMLHttpRequest::SetResponseType),
                 InstanceAccessor("responseURL", &XMLHttpRequest::GetResponseURL, nullptr),
                 InstanceAccessor("status", &XMLHttpRequest::GetStatus, nullptr),
+                InstanceAccessor("ok", &XMLHttpRequest::Ok, nullptr),
```
bghgary, line 83: I don't see this in the spec. Why is it here?

Polyfills/XMLHttpRequest/Source/XMLHttpRequest.cpp
```diff
@@ -77,12 +80,16 @@ namespace Babylon::Polyfills::Internal
...
+                InstanceAccessor("headers", &XMLHttpRequest::GetAllResponseHeaders, nullptr),
```
bghgary, line 84: I'm confused by this. The spec is a function call [getAllResponseHeaders](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/getAllResponseHeaders). It's not a property?

Polyfills/XMLHttpRequest/Source/XMLHttpRequest.cpp
```diff
@@ -77,12 +80,16 @@ namespace Babylon::Polyfills::Internal
...
                 InstanceMethod("getResponseHeader", &XMLHttpRequest::GetResponseHeader),
                 InstanceMethod("addEventListener", &XMLHttpRequest::AddEventListener),
                 InstanceMethod("removeEventListener", &XMLHttpRequest::RemoveEventListener),
                 InstanceMethod("abort", &XMLHttpRequest::Abort),
                 InstanceMethod("open", &XMLHttpRequest::Open),
                 InstanceMethod("send", &XMLHttpRequest::Send),
+                InstanceMethod("setRequestHeader", &XMLHttpRequest::SetRequestHeader),
+                InstanceMethod("json", &XMLHttpRequest::Json),
```
bghgary, line 92: Again, not in the spec. Are you seeing this somewhere? I'm not sure why these non-spec things are being added.

Polyfills/XMLHttpRequest/Source/XMLHttpRequest.cpp
```diff
@@ -106,10 +113,19 @@ namespace Babylon::Polyfills::Internal
...
+        // check response type
+        if (m_request.ResponseType() == UrlLib::UrlResponseType::String)
+        {
+            return Napi::Value::From(Env(), m_request.ResponseString().data());
+        }
+        else
+        {
+            // if it is text, return responseText
+            gsl::span<const std::byte> responseBuffer{ m_request.ResponseBuffer() };
+            auto arrayBuffer{ Napi::ArrayBuffer::New(Env(), responseBuffer.size()) };
```
bghgary, line 125: There is an overload that takes an external data pointer. Assuming m_request doesn't get deleted, this should be better than copying.

Polyfills/XMLHttpRequest/Source/XMLHttpRequest.cpp
```diff
@@ -144,6 +160,44 @@ namespace Babylon::Polyfills::Internal
...
+        }
+        return Napi::Value::From(Env(), false);
+    }
+
+    Napi::Value XMLHttpRequest::GetAllResponseHeaders(const Napi::CallbackInfo&)
+    {
+        auto responseHeaders = m_request.GetAllResponseHeaders();
+        Napi::Object responseHeadersObject = Napi::Object::New(Env());
+
+        for (auto&& iter : responseHeaders)
```
bghgary, line 177: I'm not sure I've seen this before. Why use rvalue reference for the iterator?

Polyfills/XMLHttpRequest/Source/XMLHttpRequest.cpp
```diff
@@ -212,6 +266,9 @@ namespace Babylon::Polyfills::Internal
             throw Napi::Error::New(info.Env(), "XMLHttpRequest must be opened before it can be sent");
         }
 
+        std::string requestBody = info[0].IsString() ? info[0].As<Napi::String>().Utf8Value() : std::string();
+        m_request.SetRequestBody(requestBody);
```
bghgary, line 270: I think it would make more sense like this.
```suggestion
        if (info.Length() > 0)
        {
            if (!info[0].IsString())
            {
                throw Napi::Error:New(info.Env(), "Only strings are supported in XMLHttpRequest body");
            }

            std::string requestBody = info[0].As<Napi::String>().Utf8Value();
            m_request.SetRequestBody(requestBody);
        }
```

Tests/UnitTests/Scripts/tests.js
```diff
@@ -20,12 +20,22 @@ mocha.setup({ ui: "bdd", reporter: BabylonReporter });
...
 
 describe("XMLHTTPRequest", function () {
-    function createRequest(method, url) {
+    function createRequest(method, url, body) {
         return new Promise((resolve) => {
             const xhr = new XMLHttpRequest();
             xhr.open(method, url);
             xhr.addEventListener("loadend", () => resolve(xhr));
-            xhr.send();
+            !body ? xhr.send() : xhr.send(body);
```
bghgary, line 28: Should be equivalent to this. JavaScript shouldn't know the difference.
```suggestion
            xhr.send(body);
```

Tests/UnitTests/Scripts/tests.js
```diff
@@ -20,12 +20,22 @@ mocha.setup({ ui: "bdd", reporter: BabylonReporter });
...
+        });
+    }
+
+    function createRequestWithHeaders(method, url, headers, body) {
+        return new Promise((resolve) => {
+            const xhr = new XMLHttpRequest();
+            xhr.open(method, url);
+            headers.forEach((value, key) => xhr.setRequestHeader(key, value));
+            xhr.addEventListener("loadend", () => resolve(xhr));
+            !body ? xhr.send() : xhr.send(body);
```
bghgary, line 38: ditto

bghgary (commented):
First pass. Will review again after updates.

bghgary:
PR title could be updated here too :-)

Polyfills/XMLHttpRequest/Source/XMLHttpRequest.cpp
lmaskati: Are you referring to the ` return Napi::Value::From(Env(), m_request.ResponseString().data());` line? If so, I was trying to keep it the same as what was already being done in `XMLHttpRequest::GetResponseText(const Napi::CallbackInfo&)`

Polyfills/XMLHttpRequest/Source/XMLHttpRequest.cpp
lmaskati: Instead of doing `if (!info[0].IsString())` I did `if (!info[0].IsString() && !info[0].IsUndefined())`, to allow undefined arguments (based on this discussion, the tests now pass an undefined arguments into `xhr.send()`: https://github.com/BabylonJS/JsRuntimeHost/pull/12#discussion_r1127174569)

lmaskati: Additionally, since `info[0]` can be undefined, I've reverted back to ` info[0].IsString() ? info[0].As<Napi::String>().Utf8Value() : std::string()` from `info[0].As<Napi::String>().Utf8Value();` otherwise windows throws an error.

Polyfills/XMLHttpRequest/Source/XMLHttpRequest.cpp
```diff
@@ -212,6 +246,17 @@ namespace Babylon::Polyfills::Internal
...
 
+        if (info.Length() > 0)
+        {
+            if (!info[0].IsString() && !info[0].IsUndefined())
+            {
+                throw Napi::Error::New(info.Env(), "Only strings are supported in XMLHttpRequest body");
+            }
+
+            std::string requestBody = info[0].IsString() ? info[0].As<Napi::String>().Utf8Value() : std::string();
+            m_request.SetRequestBody(requestBody);
```
bghgary, line 257: Is this what happens on the web (i.e., passing in an undefined will cause the body to be set to empty string)? Also null is not the same as undefined in JS. Probably should check for that also. Looking at the spec, I would expect passing in undefined or null will simply not set the body at all.

bghgary (commented):
Some small things. Looking good otherwise.

Polyfills/XMLHttpRequest/Source/XMLHttpRequest.cpp
bghgary: Yes, my code doesn't account for undefined or null.

Polyfills/XMLHttpRequest/Source/XMLHttpRequest.cpp
bghgary: Discussed offline. Let's copy for now since it's safer.

Polyfills/XMLHttpRequest/Source/XMLHttpRequest.cpp
lmaskati: Reviewed spec and not setting body for undefined and null

bghgary (approved):

PR #19, opened 2023-03-23
Cleanup Initialization Logic for Polyfills

karisme:
This PR places the polyfill initialization logic inside a conditional check. This prevents it from running when the class being polyfilled already exists.

### DIFF
--- a/Polyfills/AbortController/Source/AbortController.h
+++ b/Polyfills/AbortController/Source/AbortController.h
@@ -10,7 +10,6 @@ namespace Babylon::Polyfills::Internal
     class AbortController final : public Napi::ObjectWrap<AbortController>
     {
     public:
-        static constexpr auto JS_ABORT_CONTROLLER_CONSTRUCTOR_NAME = "AbortController";
         static void Initialize(Napi::Env env);
 
         explicit AbortController(const Napi::CallbackInfo& info);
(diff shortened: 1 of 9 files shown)


Polyfills/AbortController/Source/AbortController.h
```diff
@@ -10,6 +10,7 @@ namespace Babylon::Polyfills::Internal
     class AbortController final : public Napi::ObjectWrap<AbortController>
     {
     public:
+        static constexpr auto JS_ABORT_CONTROLLER_CONSTRUCTOR_NAME = "AbortController";
```
bghgary, line 13: Why does this need to be public?

bghgary (commented):
Looks good except for one comment.

Polyfills/AbortController/Source/AbortController.h
karisme: It doesn't I can make it private or move inside the cpp file.

bghgary: I would recommend inside the cpp file.

karisme: I moved some of the other definitions into their respective cpp file as well, but AbortSignal and URLSearchParam are used outside so they needed to remain public.

bghgary (approved):

PR #20, opened 2023-03-25
v8 debugging support

bghgary:
- Add support for V8 debugging via the V8 inspector protocol, moved from Babylon Native repo, updated with simpler build steps for `llhttp` dependencies.
- Fix issue with large JS files not working. `asio` usage was incorrect in the v8 inspector agent.
- Add some JS logic to wait for debugger when requested.

There is a known issue where the inspector agent crashes after the unit tests end after attaching a debugger. #21


Tests/UnitTests/Scripts/tests.js
```diff
@@ -1,5 +1,5 @@
 ﻿// Set this to true to make attaching a debugging easier.
-const waitForDebugger = true;
+const waitForDebugger = false;
```
lmaskati, line 2: Why do we need to set this flag to false? Will this impact debugging on other platforms?

Tests/UnitTests/Scripts/tests.js
```diff
@@ -1,4 +1,7 @@
-﻿// TODO: use premade reporter (once Console Polyfill is fixed)
+﻿// Set this to true to make attaching a debugging easier.
```
lmaskati, line 1: Comment probably needs to be updated

CMakeLists.txt
```diff
@@ -74,7 +74,7 @@ if(NOT TARGET UrlLib AND JSRUNTIMEHOST_POLYFILL_XMLHTTPREQUEST)
     set_property(TARGET UrlLib PROPERTY FOLDER Dependencies)
 endif()
 
-if(JSRUNTIMEHOST_CORE_APPRUNTIME_V8_INSPECTOR)
+if(NAPI_JAVASCRIPT_ENGINE STREQUAL "V8" AND JSRUNTIMEHOST_CORE_APPRUNTIME_V8_INSPECTOR)
```
lmaskati, line 77: Why did we need this additional check?

Tests/UnitTests/Scripts/tests.js
bghgary: Setting this to true will wait for the debugger. We don't want that for normal, non-debugging cases (e.g., running in the CI).

Tests/UnitTests/Scripts/tests.js
bghgary: This comment is here to tell people to set this flag to true when debugging this file. Maybe I can reword it if it's not clear?

CMakeLists.txt
bghgary: `JSRUNTIMEHOST_CORE_APPRUNTIME_V8_INSPECTOR` is set to true by default even if the javascript engine is not V8. Without this check, it will try to compile the v8inspector even for non-V8 engines.

bghgary: Production code should explicitly set this to false to prevent debugging and reduce code size.

SergioRZMasson (approved):
LGTM

Tests/UnitTests/Scripts/tests.js
lmaskati: Got it, that makes sense.

Core/AppRuntime/V8Inspector/Source/V8InspectorSocket.cpp
```diff
@@ -642,7 +639,7 @@ void TcpHolder::SetHandler(ProtocolHandler* handler) {
   handler_ = handler;
 }
 
-int TcpHolder::WriteRaw(const std::vector<char>& buffer/*, uv_write_cb write_cb*/) {
```
lmaskati: Just for my understanding - why can't we pass buffer in as a const reference?

lmaskati (approved):

Core/AppRuntime/V8Inspector/Source/V8InspectorSocket.cpp
bghgary: This is an optimization. Passing as const reference will eventually cause the code to copy the buffer. Either we use const reference from beginning to end or use move semantics all the way through. Both will work but a mixture will cause an additional copy for no reason.

bghgary: Actually, I'm wrong about both will work. If we use const references everywhere, it will still result in an unnecessary copy when we need to use it with `asio`.

PR #22, opened 2023-04-04
AbortController/Signal Callback Fix

zacary-brown:
### AbortController/Signal
- AbortController/Signal was not checking for a null callback and would crash if the callback was called while null. Added checks to avoid this
- Properly calling eventListeners using RaiseEvent
- Added unit tests to check for this situation


Documentation/DebugJavascriptDevTools.md
```diff
@@ -0,0 +1,43 @@
...
+
+## DevTool Settings
+
+If you have issues loading some of your Javascript files double check that this setting is enabled. Go into the settings by clicking on the gear in the top right corner and check this box in the preferences tab.
+
+![DevTools window](Images/DevTools/devtools-settings.png)
+
+## Debugging on Android
+
+Download and unzip ADB and run the following commands in that folder while your android device is plugged in (or emulator is on in Android Studio)
```
lmaskati, line 37: Not a big deal, but might be useful to have a link to ADB.

Polyfills/AbortController/Source/AbortSignal.cpp
```diff
@@ -93,6 +105,19 @@ namespace Babylon::Polyfills::Internal
...
+
+    void AbortSignal::RaiseEvent(const char* eventType)
+    {
+        auto it = m_eventHandlerRefs.find(eventType);
+        if (it != m_eventHandlerRefs.end())
+        {
+            const auto& eventHandlerRefs = it->second;
+            for (const auto& eventHandlerRef : eventHandlerRefs)
+            {
+                eventHandlerRef.Call({});
```
lmaskati, line 117: Do we need to check that `eventHandlerRef` is not null / undefined?

Documentation/DebugJavascriptDevTools.md
```diff
@@ -0,0 +1,43 @@
+# JsRuntimeHost Script Debugging
+You can use Chromium DevTools to debug your JavaScript while running with the V8 JavaScript engine (Default Engine for Android).
+
+## Add the Remote Target URL
+Depending on which Chromium-based browser you're using (Edge or Chrome), navigate to `edge://inspect` or `chrome://inspect`.
```
bghgary, line 5: I know you copied this but a cross-browser way to do this is to use `about://inspect`.

Polyfills/AbortController/Source/AbortController.cpp
```diff
@@ -26,8 +26,12 @@ namespace Babylon::Polyfills::Internal
...
     void AbortController::Abort(const Napi::CallbackInfo&)
     {
-        m_signal.Set("aborted", true);
-        m_signal.Get("onabort").As<Napi::Function>().Call({});
+        AbortSignal* sig = AbortSignal::Unwrap(m_signal.Value());
+
+        if (sig != nullptr)
+        {
+            sig->Abort();
+        }
```
lmaskati, line 34: do we want to throw an error if `sig == nullptr`?

Documentation/DebugJavascriptDevTools.md
```diff
@@ -0,0 +1,43 @@
...
+
+![Target discovery settings](Images/DevTools/chrome-targets.png)
+
+## Setting Up To Debug
+Make sure to change the line in [tests.js](https://github.com/BabylonJS/JsRuntimeHost/blob/f487c7b3f89b407e95a53543a06a34f1a1fbb860/Tests/UnitTests/Scripts/tests.js#L2) to true, or else the DevTools won't have enough time to attach to the Javascript Instance. If you are debugging your own script, you can also use [setTimeout()](https://developer.mozilla.org/en-US/docs/web/api/settimeout) to delay the execution of the code you are trying to debug while you wait to attach the debugger.
+
+Also, currently a bug, but line 574 in file [V8InspectorAgent.cpp](https://github.com/BabylonJS/JsRuntimeHost/blob/f487c7b3f89b407e95a53543a06a34f1a1fbb860/Core/AppRuntime/V8Inspector/Source/V8InspectorAgent.cpp#L574) needs to be commented out if it hasn't been already.
+```
+CHECK_EQ(State::kAccepting, state_);
```
bghgary, line 15: I already removed this in the code.

Documentation/DebugJavascriptDevTools.md
```diff
@@ -0,0 +1,43 @@
...
+Once you have the DevTools window open, you can use it in the same way as if you were inspecting a web page.
+For more information, see this documentation from Google on [how to debug JavaScript using Chrome DevTools](https://developer.chrome.com/docs/devtools/javascript/).
+
+![DevTools window](Images/DevTools/chrome-debugger.png)
+
+## DevTool Settings
+
+If you have issues loading some of your Javascript files double check that this setting is enabled. Go into the settings by clicking on the gear in the top right corner and check this box in the preferences tab.
+
+![DevTools window](Images/DevTools/devtools-settings.png)
```
bghgary, line 33: It sounds like this is optional, but it really isn't (at least not for default usage). Maybe it would be good to include as one of a main step right after clicking inspect.

Documentation/DebugJavascriptDevTools.md
```diff
@@ -0,0 +1,43 @@
...
+
+## Debugging on Android
+
+Download and unzip ADB and run the following commands in that folder while your android device is plugged in (or emulator is on in Android Studio)
+```
+adb forward tcp:5643 tcp:5643
+adb reverse tcp:8000 tcp:8000
```
bghgary, line 40: This one isn't necessary. Also, if `adb forward` isn't run, inspect won't show up. I think this instruction should be earlier.

Tests/UnitTests/Scripts/tests.js
```diff
@@ -23,16 +23,48 @@ mocha.setup({ ui: "bdd", reporter: BabylonReporter });
...
+
+    it("should not throw while aborting and correctly trigger both callbacks", function (done) {
         const controller = new AbortController();
         expect(controller.signal.aborted).to.equal(false);
 
         // Expect aborted to be true after abort()
         controller.signal.onabort = () => {
             expect(controller.signal.aborted).to.equal(true);
-            done();
+            // no done(); .onabort is always called first
```
bghgary, line 43: This test is relying on implementation details which is not that great.

Polyfills/AbortController/Source/AbortSignal.cpp
zacary-brown: This was a function already existing from XMLHttpRequest. I can't imagine that someone is going to .addEventListener('abort', null) but I suppose it might be a possibility?

Tests/UnitTests/Scripts/tests.js
zacary-brown: I'm not super proficient in the testing language, I was just introduced to the done() functionality the other day. Should the done() just be removed? Or should I separate the .onabort and .addEventListener callback tests?

bghgary: Separating would be one option. Another option is to set flags (one for each callback) and call done when all the flags are set.

Polyfills/AbortController/Source/AbortController.cpp
zacary-brown: Doesn't seem like it's possible for the user to set the signal as null, so the only way for that to happen is for napi to fail the construction of it. For extra safety I could add throws for this and inside of the controller construction.

Documentation/DebugJavascriptDevTools.md
bghgary, line 1: nit: I know you copied this from Babylon Native, but can you rename the file to `DebugJavaScriptDevTools.md` with capital S? Also in other places in the readme, please use JavaScript and not Javascript.

Documentation/DebugJavascriptDevTools.md
```diff
@@ -0,0 +1,37 @@
+# JsRuntimeHost Script Debugging
+You can use Chromium DevTools to debug your JavaScript while running with the V8 JavaScript engine (Default Engine for Android).
+
+## Add the Remote Target URL
+On whichever Chromium-based browser you're using (Edge or Chrome), navigate to `about://inspect`.
+Click `Configure...` and in the target discovery settings add an entry `localhost:{port}`, where `{port}` is consistent with the one supplied to the call in [AppRuntime_V8.cpp](https://github.com/BabylonJS/JsRuntimeHost/blob/f487c7b3f89b407e95a53543a06a34f1a1fbb860/Core/AppRuntime/Source/AppRuntime_V8.cpp#L85) and [AppRuntime_JSI.cpp](https://github.com/BabylonJS/JsRuntimeHost/blob/09c5acc49a6a64dec0e62075aebac6ad0a32aa93/Core/AppRuntime/Source/AppRuntime_JSI.cpp#L39).
```
bghgary, line 6: The JSI one isn't hooked up yet, so maybe don't put that here yet. Maybe just put a hard coded 5643 for the port for now. If we decide that this port needs to be configurable, we can update the code and docs.

Documentation/DebugJavascriptDevTools.md
```diff
@@ -0,0 +1,37 @@
...
+
+![Target discovery settings](Images/DevTools/chrome-targets.png)
+
+## Setting Up To Debug
+Make sure to change the line in [tests.js](https://github.com/BabylonJS/JsRuntimeHost/blob/f487c7b3f89b407e95a53543a06a34f1a1fbb860/Tests/UnitTests/Scripts/tests.js#L2) to true, or else the DevTools won't have enough time to attach to the Javascript Instance. If you are debugging your own script, you can also use [setTimeout()](https://developer.mozilla.org/en-US/docs/web/api/settimeout) to delay the execution of the code you are trying to debug while you wait to attach the debugger.
+
+## Debugging on Android/Emulator
+
+Download and unzip [ADB](https://developer.android.com/tools/releases/platform-tools) and run the following commands in that folder while your android device is plugged in (or emulator is on in Android Studio)
```
bghgary, line 15: nit:
```suggestion
Install [ADB](https://developer.android.com/tools/releases/platform-tools) and run the following command with an Android device plugged in or Android emulator running.
```

Documentation/DebugJavascriptDevTools.md
```diff
@@ -0,0 +1,37 @@
...
+```
+adb forward tcp:5643 tcp:5643
+```
+
+Note, every time you unplug and re-plug your device these ADB commands need to be run again.
```
bghgary, line 20:
```suggestion
Note, every time the Android device is unpluggeg and re-plugged, this ADB command need to be run again.
```

Documentation/DebugJavascriptDevTools.md
```diff
@@ -0,0 +1,37 @@
...
+
+## Click Inspect
+Run the UnitTests application and wait for the DevTools page to recognize the inspector server spun up by the app. Once you see the app pop up in the list of Remote Targets, you may click inspect to open up a DevTools window.
+
+![Ready to inspect](Images/DevTools/chrome-inspect.png)
+
+## DevTool Settings
+
+There is also a setting that needs to be enabled in the DevTools. Go into the settings by clicking on the gear in the top right corner and check this box in the preferences tab.
```
bghgary, line 29:
```suggestion
There is also a setting that needs to be enabled in the DevTools. Go into the settings by clicking on the gear in the top right corner and check this box in the preferences tab. This only needs to be done once.
```

Documentation/Readme.md
```diff
@@ -0,0 +1,4 @@
+# JsRuntimeHost Documention
+
+## Debugging Javascript on Windows and Android
```
bghgary, line 3:
```suggestion
## Debugging JavaScript on Android
```

Polyfills/AbortController/Source/AbortController.cpp
bghgary: If it's not supposed to be possible, I would recommend replacing the if/throw with an assert.

Polyfills/AbortController/Source/AbortController.cpp
```diff
@@ -26,14 +26,29 @@ namespace Babylon::Polyfills::Internal
...
         : Napi::ObjectWrap<AbortController>{info}
     {
         m_signal = Napi::Persistent(info.Env().Global().Get(AbortSignal::JS_ABORT_SIGNAL_CONSTRUCTOR_NAME).As<Napi::Function>().New({}));
+
+        const auto& sig = m_signal.Value();
+
+        if (sig.IsNull() || sig.IsUndefined())
+        {
+            throw std::runtime_error("AbortSignal should not be null/undefined");
+        }
```
bghgary, line 51: Why is this necessary?

Polyfills/AbortController/Source/AbortSignal.cpp
```diff
@@ -26,6 +26,19 @@ namespace Babylon::Polyfills::Internal
...
 
+    void AbortSignal::Abort()
+    {
+        m_aborted = true;
+
+        auto onabort = m_onabort.Value();
+        if (!onabort.IsNull() && !onabort.IsUndefined())
+        {
+            onabort.Call({});
+        }
```
bghgary, line 37: Does this work if the reference is empty? It seems like you have unit test for this, but I'm not sure why you don't need to check if a reference is empty.

Tests/UnitTests/Scripts/tests.js
```diff
@@ -23,16 +23,59 @@ mocha.setup({ ui: "bdd", reporter: BabylonReporter });
...
+        const controller = new AbortController();
+        expect(controller.signal.aborted).to.equal(false);
+
+        // If this function is unsuccessfully removed it will assert when called
+        function Throw() {
+            expect(controller.signal.aborted).to.equal(false);
         }
+
+        controller.signal.addEventListener('abort', Throw);
+        controller.signal.removeEventListener('abort', Throw);
```
bghgary, line 74:
```suggestion
        const onAbort = () => {
            expect(controller.signal.aborted).to.equal(false);
        };

        controller.signal.addEventListener('abort', onAbort);
        controller.signal.removeEventListener('abort', onAbort);
```

bghgary (commented):
Mostly small stuff. I would recommend splitting this into two PRs since the docs and abort are not related to each other.

Polyfills/AbortController/Source/AbortController.cpp
zacary-brown: It probably isn't. It was brought up the other day asking if m_signal could be null/undefined but if anything in it's construction fails Napi would probably throw something first.

Polyfills/AbortController/Source/AbortSignal.cpp
zacary-brown: It looks like when checking .IsUndefined(), it calls Type() == napi_undefined which Type() then calls IsEmpty(). IsEmpty() returns true, and because IsEmpty() returned true Type() returns napi_undefined. IsUndefined() seems to catch the empty reference case.

Polyfills/AbortController/Source/AbortController.cpp
bghgary: I don't think it's possible to persist a null or undefined.

Polyfills/AbortController/Source/AbortSignal.cpp
bghgary: Interesting. I'm pretty sure this is implemented differently in the napi/jsi adaptor. I might need to fix that.

bghgary (approved):

bghgary (approved):

PR #39, opened 2023-06-30
CMake fetch standard

CedricGuillemet:
Doing this PR as a test so we are on the same page. I'll do the same work for BN and others once we settle down with this 'standard'.

- moved every `FetchContent_Declare` at the top of the file
- added `CONTENT_TO_FETCH` variable to hold content to really fetch
- a fetch for ios toolchain is separated. It doesn't look possible to fetch the toolchain at the same place. I guess it's too late in CMake pipeline
- set `TARGET` properties after fetch for obvious reason. This means some tests are duplicated. IDK if it's possible to improve that (like `if(NOT TARGET UrlLib AND JSRUNTIMEHOST_POLYFILL_XMLHTTPREQUEST)`)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -29,7 +29,10 @@ FetchContent_Declare(llhttp
 set(CONTENT_TO_FETCH cmake-extensions)
 
 if(IOS)
-    set(CONTENT_TO_FETCH ${CONTENT_TO_FETCH} ios-cmake)
+    FetchContent_MakeAvailable(ios-cmake)
+    set(CMAKE_TOOLCHAIN_FILE "${ios-cmake_SOURCE_DIR}/ios.toolchain.cmake" CACHE PATH "")
+    set(PLATFORM "OS64COMBINED" CACHE STRING "")
+    set(DEPLOYMENT_TARGET "12" CACHE STRING "")
 endif()
 
 # Project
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -63,56 +67,46 @@ option(NAPI_BUILD_ABI "Build the ABI layer." ON)
...
 
-if(NOT TARGET UrlLib AND JSRUNTIMEHOST_POLYFILL_XMLHTTPREQUEST)
-    FetchContent_Declare(
-        UrlLib
-        GIT_REPOSITORY https://github.com/BabylonJS/UrlLib.git
-        GIT_TAG 4182996cbe3b85a960e38bd8a61892688ed55213)
+# Fetching content
+message(STATUS "Fetching for ${CMAKE_PROJECT_NAME} : ${CONTENT_TO_FETCH}")
+FetchContent_MakeAvailable(${CONTENT_TO_FETCH})
+message(STATUS "${CMAKE_PROJECT_NAME} content fetched.")
```
bghgary, line 92: See https://stackoverflow.com/questions/38938315/difference-between-cmake-project-name-and-project-name. I would also suggest we format it differently.

```suggestion
message(STATUS "Fetching dependencies for ${PROJECT_NAME} (${CONTENT_TO_FETCH})")
FetchContent_MakeAvailable(${CONTENT_TO_FETCH})
message(STATUS "Fetching dependencies for ${PROJECT_NAME} - done")
```

bghgary (approved):

PR #44, opened 2023-08-02
Fix unittests

CedricGuillemet:
- reorder CI win32 job to fix the artifact publish
- Fix websocket cancellation with this related PR : https://github.com/BabylonJS/UrlLib/pull/8
- removed setimeout workaround for Mocha exitCode
- changed the reporter from tap to spec. The tap reporter makes extra calls to settimeout that cause extra async operation after setting exit code. This does not happen with spec

I kicked 5 pipeline runs and 1 Android JSC job failed because it was not able to start android emulator, so issue seams to be gone.

### DIFF
--- a/Polyfills/WebSocket/Source/WebSocket.cpp
+++ b/Polyfills/WebSocket/Source/WebSocket.cpp
@@ -172,10 +172,14 @@ namespace Babylon::Polyfills::Internal
     void WebSocket::OpenCallback()
     {
         m_runtimeScheduler([this, cancellationSource{m_cancellationSource}]() {
+            if (cancellationSource->cancelled())
+            {
+                return;
+            }
             m_readyState = ReadyState::Open;
             try
             {
-                if (!cancellationSource->cancelled() && !m_onopen.IsEmpty())
+                if (!m_onopen.IsEmpty())
                 {
                     m_onopen.Call({});
                 }
--- a/Polyfills/WebSocket/Source/WebSocket.cpp
+++ b/Polyfills/WebSocket/Source/WebSocket.cpp
@@ -41,8 +41,14 @@ namespace Babylon::Polyfills::Internal
               [this](const std::string& message) { MessageCallback(message); },
               [this](const std::string& message) { ErrorCallback(message); })
         , m_url(info[0].As<Napi::String>())
+        , m_cancellationSource{std::make_shared<arcana::cancellation_source>()}
     {
         m_webSocket.Open();
+    }
+
+    WebSocket::~WebSocket()
+    {
+        m_cancellationSource->cancel();
     }
 
     void WebSocket::Close(const Napi::CallbackInfo& info)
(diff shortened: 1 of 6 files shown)


Polyfills/WebSocket/Source/WebSocket.cpp
```diff
@@ -165,11 +171,11 @@ namespace Babylon::Polyfills::Internal
 
     void WebSocket::OpenCallback()
     {
-        m_runtimeScheduler([this]() {
+        m_runtimeScheduler([this, cancellationSource{m_cancellationSource}]() {
             m_readyState = ReadyState::Open;
             try
             {
-                if (!m_onopen.IsEmpty())
+                if (!cancellationSource->cancelled() && !m_onopen.IsEmpty())
```
bghgary, line 178: This check should be at the beginning of this scope. `m_readyState` might be dead also. I think you should try removing these fixes for WebSocket for now and see if the problem still persists.

CedricGuillemet: As said offline, this change is mandatory (at least) for Chakra. I've changed the scope here.

Polyfills/WebSocket/Source/WebSocket.cpp
```diff
@@ -211,14 +225,17 @@ namespace Babylon::Polyfills::Internal
...
-        m_runtimeScheduler([this, message]() {
+        m_runtimeScheduler([this, message, cancellationSource{m_cancellationSource}]() {
             try
             {
                 if (!m_onmessage.IsEmpty())
                 {
                     Napi::Object messageEvent = Napi::Object::New(Env());
                     messageEvent.Set("data", message);
-                    m_onmessage.Call({messageEvent});
+                    if (!cancellationSource->cancelled())
```
bghgary, line 235: Do the same thing as earlier and check at the beginning of the lambda.

Polyfills/WebSocket/Source/WebSocket.cpp
```diff
@@ -231,14 +248,17 @@ namespace Babylon::Polyfills::Internal
...
-        m_runtimeScheduler([this, message]() {
+        m_runtimeScheduler([this, message, cancellationSource{m_cancellationSource}]() {
             try
             {
                 if (!m_onerror.IsEmpty())
                 {
                     Napi::Object errorEvent = Napi::Object::New(Env());
                     errorEvent.Set("message", message);
-                    m_onerror.Call({errorEvent});
+                    if (!cancellationSource->cancelled())
```
bghgary, line 258: Do the same thing as earlier and check at the beginning of the lambda.

bghgary (approved):

Polyfills/WebSocket/Source/WebSocket.cpp
CedricGuillemet: Fixed

PR #48, opened 2023-09-05
fix v8inspector crash at shutdown

CedricGuillemet:
`Stop ` method is called from 2 threads. the main when app stops or by the thread itself. A mutex is added to guard concurrency. Then, when Stop is called from the main thread, unique pointer is reset just after but the thread itself is still using it. reset is removed from the main thread. Thread is responsible for creating unique pointer and deleting it.

Also removed `out_` because unused.

### DIFF
--- a/Core/AppRuntime/V8Inspector/Source/V8InspectorSocketServer.cpp
+++ b/Core/AppRuntime/V8Inspector/Source/V8InspectorSocketServer.cpp
@@ -351,6 +351,7 @@ namespace Babylon
 
     void InspectorSocketServer::Stop()
     {
+        std::lock_guard<std::mutex> guard{m_mutex};
         if (state_ == ServerState::kStopped)
             return;
         CHECK_EQ(state_, ServerState::kRunning);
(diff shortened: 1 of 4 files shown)


Core/AppRuntime/V8Inspector/Source/V8InspectorSocketServer.cpp
```diff
@@ -351,6 +354,7 @@ namespace Babylon
 
     void InspectorSocketServer::Stop()
     {
+        std::lock_guard<std::mutex> guard{m_mutex};
```
bghgary, line 357: What's this for? Is Stop being called from multiple threads?

bghgary: Ah, I missed the PR description. Is this the only function that is called from two threads?

CedricGuillemet: It's the only one I've seen so far.

bghgary (approved):

PR #49, opened 2023-09-06
Resolve symlinks for app:///[...] and file:///[...] URLs on Windows

docEdub:
On macOS and Linux, app:/// and file:/// URLs that point to symlinks automatically resolve to the target file, but not on Windows.

This change updates UrlLib to resolve symlinks programmatically on Windows so the behavior is consistent across all platforms, and adds unit tests. Note that unit tests are only done on macOS, Unix, and Win32 platforms because we cannot create symlinks on the CI machines for other platforms without admin rights.

I will update the UrlLib Git tag when https://github.com/BabylonJS/UrlLib/pull/10 is merged.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -19,7 +19,7 @@ FetchContent_Declare(arcana
     GIT_TAG 10d167ffe0f86b2ddaccd93fd78ff2240cbb0c49)
 FetchContent_Declare(UrlLib
     GIT_REPOSITORY https://github.com/BabylonJS/UrlLib.git
-    GIT_TAG f09f431f24cac7ffb0f17e5a83ce6c2f11d29e31)
+    GIT_TAG e519cd6c05d28d6888661d9ffc901882179e5818)
 FetchContent_Declare(asio
     GIT_REPOSITORY https://github.com/chriskohlhoff/asio.git
     GIT_TAG f693a3eb7fe72a5f19b975289afc4f437d373d9c)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -19,7 +19,7 @@ FetchContent_Declare(arcana
     GIT_TAG 10d167ffe0f86b2ddaccd93fd78ff2240cbb0c49)
 FetchContent_Declare(UrlLib
     GIT_REPOSITORY https://github.com/BabylonJS/UrlLib.git
-    GIT_TAG e519cd6c05d28d6888661d9ffc901882179e5818)
+    GIT_TAG b4da79c52829be88588ce1bb194b855d05ae008d)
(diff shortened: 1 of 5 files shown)


CMakeLists.txt
```diff
@@ -19,7 +19,7 @@ FetchContent_Declare(arcana
     GIT_TAG 10d167ffe0f86b2ddaccd93fd78ff2240cbb0c49)
 FetchContent_Declare(UrlLib
     GIT_REPOSITORY https://github.com/BabylonJS/UrlLib.git
-    GIT_TAG f09f431f24cac7ffb0f17e5a83ce6c2f11d29e31)
+    GIT_TAG d18aa71dd1a15f948344cf32059536e38e8cd935)
```
CedricGuillemet, line 22: git hash will change with a squash and merge of the PR. Or, to say it differently, a new commit will be pushed to urllib main branch when the PR will be merged. so, this commit number will not be valid.

Tests/UnitTests/CMakeLists.txt
```diff
@@ -5,6 +5,11 @@ endif()
 set(SCRIPTS
     "Scripts/tests.js")
 
+if(APPLE OR UNIX OR WIN32)
```
CedricGuillemet, line 8: Android is Unix. so, if I'm not wrong this can be simplified to (NOT WINDOW_STORE). Or is Android expected to not run this code block?

CMakeLists.txt
docEdub: Ya, I will update it when the UrlLib PR is merged.

Tests/UnitTests/CMakeLists.txt
docEdub: I was not able to create the symlinks needed for testing on Android (and iOS), so I left those platforms out of the unit tests.

Tests/UnitTests/Android/app/src/main/cpp/CMakeLists.txt
```diff
@@ -17,7 +17,7 @@ add_library(UnitTestsJNI SHARED
     JNI.cpp
     ${UNIT_TESTS_DIR}/Shared/Shared.h
     ${UNIT_TESTS_DIR}/Shared/Shared.cpp)
-target_compile_definitions(UnitTestsJNI PRIVATE JSRUNTIMEHOST_PLATFORM="${JSRUNTIMEHOST_PLATFORM}")
+target_compile_definitions(UnitTestsJNI PRIVATE JSRUNTIMEHOST_PLATFORM="Android")
```
docEdub, line 20: CMake variable `JSRUNTIMEHOST_PLATFORM` isn't set in this scope.

PR #50, opened 2023-09-11
Console log params

carolhmj:
This PR adds string substitutions https://developer.mozilla.org/en-US/docs/Web/API/console#using_string_substitutions to the console.log polyfill. For now, the %o substitution (Objects) just prints [object Object], but this capability can be added later if needed.


CMakeLists.txt
```diff
@@ -25,8 +25,11 @@ FetchContent_Declare(asio
     GIT_TAG f693a3eb7fe72a5f19b975289afc4f437d373d9c)
 FetchContent_Declare(llhttp
     URL "https://github.com/nodejs/llhttp/archive/refs/tags/release/v8.1.0.tar.gz")
+FetchContent_Declare(fmt
+    GIT_REPOSITORY https://github.com/fmtlib/fmt.git
+    GIT_TAG a0b8a92e3d1532361c2f7feb63babc5c18d00ef2)
```
bghgary, line 30: Adding a new OSS dependency requires some more work. We should chat offline.

Polyfills/Console/Source/Console.cpp
```diff
@@ -36,18 +39,59 @@ namespace
...
-        {
-            if (i > 0)
-            {
-                ss << " ";
+        
+        std::string formattedString = "";
+        if (info.Length() > 0) {
+            formattedString = info[0].ToString().Utf8Value();
+            // check if this string has substitutions or not
+            std::regex toSub("(%[oOs])|(%(\\d*\\.\\d*)?[dif])");
```
CedricGuillemet, line 47: Regex are slow, like really slow. There should be a faster test before to check for `%` character before.

CMakeLists.txt
CedricGuillemet: I'm not sure fmt is needed here in the first place actually. only `fmt::sprintf` is used. These 2 calls can be replaced with ìtoa` and `ftoa` or `sprintf`

Tests/UnitTests/Scripts/tests.js
```diff
@@ -594,6 +594,33 @@ describe("URLSearchParams", function () {
     });
 });
 
+
+describe("Console", function () {
```
CedricGuillemet, line 598: I would also add : `console.log('%s');` and `console.log('%a %b %c', 123);`

CMakeLists.txt
bghgary: If that works, it is much better than adding a new dependency.

carolhmj: using sprintf_s instead 😄

bghgary: We have to be very careful using printf-like functions with a user supplied format string. There are lots of possible security issues here.

CedricGuillemet:
sprintf_s is windows specific but snprintf is not https://cplusplus.com/reference/cstdio/snprintf/

Polyfills/Console/Source/Console.cpp
```diff
@@ -36,18 +38,64 @@ namespace
...
```
bghgary, line 46: You don't want to create a new `regex` object each time. This can be const and put into an anonymous namespace in this file to make it only created once.

Polyfills/Console/Source/Console.cpp
```diff
@@ -36,18 +38,64 @@ namespace
...
+            std::smatch matches;
+            
+            // for each argument beyond the first (which is the string itself, try to find a substitution string)
+            for (size_t i = 1; i < info.Length(); i++) {
+                Napi::Value v = info[i];
+
+                // check if there's a corresponding match to this argument
+                if (formattedString.find("%") != std::string::npos && std::regex_search(formattedString, matches, toSub)) {
+                    std::string match = matches[0].str();
```
bghgary, line 55: This copies the string for no reason. Do this instead:
```suggestion
                    const std::string& match = matches[0].str();
```

Polyfills/Console/Source/Console.cpp
```diff
@@ -36,18 +38,64 @@ namespace
...
+                            written = sprintf_s(buffer, size, match.c_str(), number.Int64Value());
+                        }
+                        converted = std::string(buffer);
+                    }
+                    else {
+                        converted = "NaN";
+                    }
+                    
+                    // replace converted on the original match place
+                    size_t start_pos = formattedString.find(match);
```
bghgary, line 85: Why match this again? You should already have the position from the original match?

bghgary (commented):
This new algorithm is probably much slower than the original stringstream method. Is there a better way to handle this instead of using string replacement/concatenation which will result in multiple string allocations?

Polyfills/Console/Source/Console.cpp
```diff
@@ -36,18 +38,64 @@ namespace
...
+                        // object: for now just turn into [Object object]
+                        converted = v.ToString().Utf8Value();
+                    }
+                    else if (v.IsNumber() && v.ToString().Utf8Value() != "NaN") {
+                        // number formatting
+                        // if we have the float specified, force convert to a float
+                        Napi::Number number = v.ToNumber();
+                        // question: is there a "good" limit here?
+                        char buffer[100];
+                        int size = 100;
```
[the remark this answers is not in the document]
CedricGuillemet: sizeof will work as well and is part of the language.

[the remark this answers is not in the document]
bghgary: Yes, but sizeof is intended for getting the number of bytes. It will work because this happens to be bytes because it's char. Generally, it is better to use `std::size` to get the number of elements and show intent.

carolhmj:
> This new algorithm is probably much slower than the original stringstream method. Is there a better way to handle this instead of using string replacement/concatenation which will result in multiple string allocations?

@bghgary At first I thought about just using the sprintf/snprintf/etc method, but that fails for some cases where console.log doesn't. For example, console.log("%s", 1) fails for sprintf, but works on JS. Because of this I'm checking each match individually to convert to a type that won't fail in the sprintf. Maybe we could first check if the console string doesn't have any substitutions on its first argument, and if so, fall back to the previous faster method?

bghgary:
> > This new algorithm is probably much slower than the original stringstream method. Is there a better way to handle this instead of using string replacement/concatenation which will result in multiple string allocations?
> 
> @bghgary At first I thought about just using the sprintf/snprintf/etc method, but that fails for some cases where console.log doesn't. For example, console.log("%s", 1) fails for sprintf, but works on JS. Because of this I'm checking each match individually to convert to a type that won't fail in the sprintf. Maybe we could first check if the console string doesn't have any substitutions on its first argument, and if so, fall back to the previous faster method?

I don't know what is meant by sprintf method. If this means passing the user input directly into sprintf, this is really bad in terms of security unless we do a lot of checking of the format string which probably negates the benefits of using it. Let's find some time to discuss offline with @CedricGuillemet.

Polyfills/Console/Source/Console.cpp
```diff
@@ -14,28 +14,14 @@
 
 namespace
 {
+    // Struct used for splitting string into parts
     struct STRING_PART
```
CedricGuillemet, line 18: name should be `StringPart`

Tests/UnitTests/Scripts/tests.js
```diff
@@ -616,16 +615,16 @@ describe("URLSearchParams", function () {
...
+  it("should log sequence of strings", function () {
         expect(() => console.log("I", "am", "a", "test", "string", 2, 2.345, { foo: 'bar' })).to.not.throw();
     });
     it("Should log string substitutions", function () {
-        expect(() => console.log("String sub: %s, float sub: %.2f, int sub: %2.d", "string", 3.1457, 3.1457)).to.not.throw();
+      expect(() => console.log("String sub: %s, float sub: %.2f, int sub: %2.d", "string", 3.1457, 3.1457)).to.not.throw();
+        //expect(() => console.log("float sub: %.2f, int sub: %2.d", 3.1457, 3.1457)).to.not.throw();
     });
     it("Should allow numbers to substitute strings", function () {
         expect(() => console.log("Print these numbers! %s %s", 1.2345, 1)).to.not.throw();
```
CedricGuillemet, line 630: nit: does it support emojis? not needed for this feqture but I'm curious.

Polyfills/Console/Source/Console.cpp
```diff
@@ -148,7 +86,45 @@ namespace
         if (info.Length() > 0)
         {
             std::string firstArg = info[0].ToString();
-            std::vector<STRING_PART> parts = SplitString(firstArg);
+            std::vector<STRING_PART> parts = std::vector<STRING_PART>();
```
CedricGuillemet, line 89:
```suggestion
            std::vector<StringPart> parts{};
```

Polyfills/Console/Source/Console.cpp
```diff
@@ -34,20 +71,137 @@ namespace
         }
     }
 
+    bool EndsWith(std::string_view s, char c) {
+        if (s.size() > 0)
+        {
+            return s.at(s.size() - 1) == c;
+        }
```
bghgary, line 78:
```suggestion
        if (!s.empty())
        {
            return s.back() == c;
        }
```

Polyfills/Console/Source/Console.cpp
```diff
@@ -71,7 +225,7 @@ namespace Babylon::Polyfills::Console
 {
     void Initialize(Napi::Env env, CallbackT callback)
     {
-        Napi::HandleScope scope{env};
+        Napi::HandleScope scope{ env };
```
bghgary, line 228: Revert pls. This is not the right style. Use clang-format.

Polyfills/Console/Source/Console.cpp
```diff
@@ -56,13 +210,13 @@ namespace
...
-                    if (!existingFunction->Value().IsUndefined())
-                    {
-                        Call(existingFunction->Value(), info);
-                    }
-                },
+            if (!existingFunction->Value().IsUndefined())
+            {
+                Call(existingFunction->Value(), info);
+            }
+        },
```
bghgary, line 219: I don't know what is going on with the spacing here. Please use clang-format.

Polyfills/Console/Source/Console.cpp
```diff
@@ -34,20 +71,137 @@ namespace
...
+            for (j = 0; j < firstArg.size(); j++)
+            {
+                const char currChar = firstArg.at(j);
+                if (!matching && currChar == '%')
+                {
+                    matching = true;
+                    if (j > 0)
+                    {
+                        std::string_view prefix(firstArg.data() + start, j - start);
+                        parts.push_back({prefix, false});
```
bghgary, line 104: This is allocations again. We need to try to not allocate inside the algorithm as much as possible.

bghgary (changes requested):
I didn't finish the review, but there are enough issues that we should sync offline and discuss.

Tests/UnitTests/Scripts/tests.js
carolhmj: I'm not sure, is there a way to test for this?

carolhmj:
I tried updating the algorithm to avoid using that vector and doing less allocations. I still had to create a string, but if there's a way to not do it then I'll change it 🙂

Polyfills/Console/Source/Console.cpp
```diff
@@ -34,16 +38,136 @@ namespace
...
+    // based on: https://stackoverflow.com/a/26221725
+    template<typename T>
+    std::string FormatValue(std::string& formatString, T value)
+    {
+        // get the size of what will be written (+1 for null terminator)
+        int size_s = snprintf(nullptr, 0, formatString.c_str(), value) + 1;
+        // if size is 0 or less, we had a formatting error
+        if (size_s <= 0)
+        {
+            throw std::runtime_error("Error during console.log formatting.");
```
bghgary, line 59: You need to throw a Napi::Error instead of std::runtime_error which requires that you pass in Napi::Env. Or put a try/catch outside and convert to Napi::Error outside.

Polyfills/Console/Source/Console.cpp
```diff
@@ -34,16 +38,136 @@ namespace
...
+        }
+        return false;
+    }
+
+    // based on: https://stackoverflow.com/a/26221725
+    template<typename T>
+    std::string FormatValue(std::string& formatString, T value)
+    {
+        // get the size of what will be written (+1 for null terminator)
+        int size_s = snprintf(nullptr, 0, formatString.c_str(), value) + 1;
```
bghgary, line 55: I know this is copied from stackoverflow, but I don't think you need +1 and then a subsequent -1 later. It used to be warranted for the buffer, but that's been replaced.

Polyfills/Console/Source/Console.cpp
```diff
@@ -34,16 +38,136 @@ namespace
...
+        // if size is 0 or less, we had a formatting error
+        if (size_s <= 0)
+        {
+            throw std::runtime_error("Error during console.log formatting.");
+        }
+        size_t size = static_cast<size_t>(size_s);
+        // create a string to hold that size (-1 as we don't want the null terminator)
+        std::string buf(size - 1, '0');
```
bghgary, line 63: nit: Modern C++ uses brace initialization. Many places.
```suggestion
        std::string buf{size - 1, '0'};
```

Polyfills/Console/Source/Console.cpp
```diff
@@ -34,16 +38,136 @@ namespace
...
+        if (info.Length() > 0)
         {
-            if (i > 0)
+            std::string firstArg = info[0].ToString();
+            size_t currArgIndex = 1;
+
+            // Split the first string into parts limited by substitution characters
+            size_t start = 0;
+            bool matching = false;
+            std::size_t j;
```
bghgary, line 80: nit: inconsistent usage of `std::` for `size_t`

Polyfills/Console/Source/Console.cpp
```diff
@@ -34,16 +38,136 @@ namespace
...
+            {
+                const char currChar = firstArg.at(j);
+                if (!matching && currChar == '%')
+                {
+                    matching = true;
+                    if (j > 0)
+                    {
+                        std::string_view prefix(firstArg.data() + start, j - start);
+                        ss << prefix;
+                    }
```
bghgary, line 91: Would it work to add to `ss` when not matching instead of this? Also, I don't know what `prefix` means here.

Polyfills/Console/Source/Console.cpp
```diff
@@ -34,16 +38,136 @@ namespace
...
+                    start = j;
+                }
+                else if (matching && (currChar == 's' || currChar == 'o' || currChar == 'O' || currChar == 'd' || currChar == 'i' || currChar == 'f'))
```
bghgary, line 94: I don't think this works. We can't assume the last character of a substitution string will always terminate in one of these characters. For example, `console.log("%l%d", 10)` should result in `%l10`.

Polyfills/Console/Source/Console.cpp
```diff
@@ -34,16 +38,136 @@ namespace
...
+                        }
+                        currArgIndex++;
+                    }
+                    else
+                    {
+                        ss << prefix;
+                    }
+                    start = j + 1;
+                }
+                else if (matching && (currChar == ' '))
```
bghgary, line 150: Similar to the previous comment, it is not guaranteed that there will be space and it doesn't make much sense to explicitly search for space.

Polyfills/Console/Source/Console.cpp
```diff
@@ -34,16 +38,136 @@ namespace
...
+                                ss << "NaN";
+                            }
+                            else
+                            {
+                                // I had to copy the string view to a string here, because the data pointer of the string view
+                                // returns a view of the entire character sequence, which caused an issue on the following
+                                // step when formatting.
+                                std::string copiedSubString(prefix);
+                                std::string formatted = FormatValue(copiedSubString, d);
+                                ss << formatted;
```
bghgary, line 118: I'm guessing this is because you need a null-terminated string, which makes sense, but we can probably safely limit the format string to a certain size and allocate the string on the stack instead of the heap.

Polyfills/Console/Source/Console.cpp
```diff
@@ -34,16 +38,136 @@ namespace
...
+                                ss << "NaN";
+                            }
+                            else
+                            {
+                                int64_t n = currArg.ToNumber().Int64Value();
+                                // Some explanation as above to why copy to a string
+                                std::string copiedSubString(prefix);
+                                std::string formatted = FormatValue(copiedSubString, n);
+                                ss << formatted;
+                            }
```
bghgary, line 136: Some relevant info:
- JS has no concept of integers. All numbers are doubles. Node-API will coerce the JS double value to a C integer value in `Int64Value()`.
- NaN is a floating-point concept. Integers don't have NaN.

Probably what should happen here is call `DoubleValue()` first to check for NaN like you are doing and then cast the result to integer if the currChar is `i`. Right now, this code will always format as integer even if it's double.

Polyfills/Console/Source/Console.cpp
```diff
@@ -34,16 +38,136 @@ namespace
...
+                {
+                    matching = false;
+                    std::string_view prefix(firstArg.data() + start, j - start + 1);
+                    // This is a substitution argument, so if we have any remaining arguments to substitute, we should.
+                    // Try to use the next arg for subsitution
+                    if (currArgIndex < info.Length())
+                    {
+                        Napi::Value currArg = info[currArgIndex];
+                        // Check the type of the sub string
+                        if (EndsWith(prefix, 'f'))
```
bghgary, line 104: Can you use `currChar` instead of `EndsWith`?

bghgary (changes requested):

Polyfills/Console/Source/Console.cpp
carolhmj: I tried removing the +1/-1 when I noticed I didn't need the null terminator, but this gives me wrong results, of:
result1
Opposed to:
result2
With the +1/-1

CedricGuillemet: According to this doc https://cplusplus.com/reference/cstdio/snprintf/ it's not counting the null character.

Polyfills/Console/Source/Console.cpp
carolhmj: That's a good catch, I didn't think of this case.

Polyfills/Console/Source/Console.cpp
```diff
@@ -34,16 +34,82 @@ namespace
...
 
+    inline bool isNumberEnding(char c)
+    {
+        return c == 'd' || c == 'i' || c == 'f';
+    }
+
+    inline bool isStringOrObjectEnding(char c)
+    {
+        return c == 'o' || c == 'O' || c == 's';
+    }
```
bghgary, line 45: nit: I'm not sure this makes the code clearer. You already have to explain exactly what is happening when you call these functions and they are only called once. I think it's clearer without these functions.

Polyfills/Console/Source/Console.cpp
```diff
@@ -34,16 +34,82 @@ namespace
...
+        if (info.Length() > 0)
         {
-            if (i > 0)
+            std::string firstArg = info[0].ToString();
+            size_t currArgIndex = 1;
+
+            std::size_t j = 0;
+            while (j < firstArg.size())
+            {
+                const char currChar = firstArg.at(j);
```
bghgary, line 58: Use `[]` here since you are already doing a bounds check. `at` will perform bounds check that is unnecessary. https://stackoverflow.com/questions/14699060/what-is-the-difference-between-stringat-and-stringoperator

Polyfills/Console/Source/Console.cpp
```diff
@@ -34,16 +34,82 @@ namespace
...
+                // When a '%' is encountered, check the next character to determine the type of string we have
+                if (currChar == '%' && j < firstArg.size() - 1 && currArgIndex < info.Length())
+                {
+                    char nextChar = firstArg.at(j + 1);
```
bghgary, line 62: Ditto on `at` usage.

Polyfills/Console/Source/Console.cpp
```diff
@@ -34,16 +34,82 @@ namespace
...
+                    if (isStringOrObjectEnding(nextChar))
+                    {
+                        ss << currArg.ToString().Utf8Value();
+                        currArgIndex++;
+                    }
+                    // or [dif], when it specifies a number
+                    else if (isNumberEnding(nextChar))
+                    {
+                        double d = currArg.ToNumber().DoubleValue();
+                        // in IEEE standard, comparing NaN is always false. this avoids using isnan
```
bghgary, line 74: Why do we not want to use `std::isnan`?

Polyfills/Console/Source/Console.cpp
```diff
@@ -34,16 +34,82 @@ namespace
...
+                    // the next character can be one of: [soO], when the substitution string specifies a string
+                    if (isStringOrObjectEnding(nextChar))
+                    {
+                        ss << currArg.ToString().Utf8Value();
+                        currArgIndex++;
+                    }
+                    // or [dif], when it specifies a number
+                    else if (isNumberEnding(nextChar))
+                    {
+                        double d = currArg.ToNumber().DoubleValue();
```
bghgary, line 73: Do we need to check if `currArg` is actually a number before forcing it to a number?

bghgary (commented):
Looks much better! Just some small comments.

Polyfills/Console/Source/Console.cpp
carolhmj: The Apple builds were failing if I used std::isnan, while the ubuntu builds fails if I use just isnan 😅 so @CedricGuillemet suggested me this as an alternative

Polyfills/Console/Source/Console.cpp
carolhmj: If currArg is not a number, it's converted to NaN which is what we want

Polyfills/Console/Source/Console.cpp
bghgary: Did you include `cmath`?

carolhmj: Oooh, I had taken cmath out because it worked in Windows without it, but adding it back solves that issue! Thanks!

Tests/UnitTests/Scripts/tests.js
carolhmj: It does seem to work on the pipelines 😄 
image

bghgary (approved):

PR #52, opened 2023-09-25
Disable deprecated api warnings

ryantrem:
These warnings show up as build errors in Xcode. Disabling them until their is a suitable replacement (future API, custom code, or some OSS lib).


bghgary (approved):

CedricGuillemet:
Actually, I'm not sure these 2 methods are used. I think we can remove them.

ryantrem:
> Actually, I'm not sure these 2 methods are used. I think we can remove them.

If I just comment them out, then I get a linker error. So they seem used? @bghgary

bghgary:
I would expect they are used or I would have implemented them :)

PR #55, opened 2023-09-28
Build fixes

bghgary:
- Only include tests by default when project is top level
- Fix typos with arcana and UrlLib target folders
- Fix npm install command to work properly


docEdub (approved):
👍 LGTM

PR #57, opened 2023-10-03
Update UrlLib dependency and add symlink unit tests

docEdub:


Tests/UnitTests/Scripts/tests.js
```diff
@@ -165,6 +165,20 @@ describe("XMLHTTPRequest", function () {
         expect(xhr).to.have.property('readyState', 4);
         expect(xhr).to.have.property('status', 200);
     });
+
+    if (hostPlatform === "macOS" || hostPlatform === "Unix" || hostPlatform === "Win32") {
+        it("should load URL pointing to symlink", async function () {
+            this.timeout(1000);
```
bghgary, line 171: what's this for?

bghgary (approved):

Tests/UnitTests/Scripts/tests.js
docEdub: I added this a while back to get the test working early on, but maybe it works without it now. Let me check.

docEdub: ya it works without it

PR #58, opened 2023-10-04
Add UrlLib response buffer unit test

docEdub:
Needs https://github.com/BabylonJS/UrlLib/pull/13 to pass.

### DIFF
--- a/Tests/UnitTests/Scripts/tests.js
+++ b/Tests/UnitTests/Scripts/tests.js
@@ -183,7 +183,7 @@ describe("XMLHTTPRequest", function () {
 
     it("should load URL as array buffer", async function () {
         const xhr = await createRequest("GET", "app:///Scripts/symlink_target.js", undefined, "arraybuffer");
-        var expected = Uint8Array.from("var symlink_target_js = true;".split(""), x => x.charCodeAt(0));
+        var expected = new Uint8Array("var symlink_target_js = true;".split("").map(x => x.charCodeAt(0)));
         var response = new Uint8Array(xhr.response);
         expect(response).to.eql(expected);
     });
--- a/Tests/UnitTests/Scripts/tests.js
+++ b/Tests/UnitTests/Scripts/tests.js
@@ -63,10 +63,13 @@ describe("AbortController", function () {
 });
 
 describe("XMLHTTPRequest", function () {
-    function createRequest(method, url, body) {
+    function createRequest(method, url, body, responseType) {
         return new Promise((resolve) => {
             const xhr = new XMLHttpRequest();
             xhr.open(method, url);
+            if (responseType !== undefined) {
+                xhr.responseType = responseType;
+            }
             xhr.addEventListener("loadend", () => resolve(xhr));
             xhr.send(body);
         });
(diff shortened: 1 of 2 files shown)


Tests/UnitTests/Scripts/tests.js
```diff
@@ -177,6 +180,13 @@ describe("XMLHTTPRequest", function () {
             expect(xhr).to.have.property('responseText', 'var symlink_target_js = true;');
         });
     }
+
+    it("should load URL as array buffer", async function () {
+        const xhr = await createRequest("GET", "app:///Scripts/symlink_target.js", undefined, "arraybuffer");
+        var expected = Uint8Array.from("var symlink_target_js = true;".split(""), x => x.charCodeAt(0));
```
bghgary, line 186: What's the `split` for?

bghgary (approved):

Tests/UnitTests/Scripts/tests.js
docEdub: Splits the string into individual characters fed as `x` into `x.charCodeAt(0)`. `Uint8Array.from("...")` without the character code conversion doesn't match.

bghgary: I'm not sure this is supported in Chakra, but if so, you can use https://developer.mozilla.org/en-US/docs/Web/API/TextEncoder/encode.

docEdub: That gives `ReferenceError: TextEncoder is not defined` on Chakra and V8. I haven't tried it on JSC.

bghgary: Ah, yes, it does appear to be a DOM feature, so not part of JavaScript. Never mind.

bghgary: I guess another option is to add another file that is actually binary with a few bytes that you can just hard code here.

bghgary:
Build is failing for Android JSC :-/

> 1 XMLHTTPRequest
> 10-04 19:10:29.717  3194  3212 I JsRuntimeHost:        should load URL as array buffer TypedArray.from requires its this argument subclass a TypedArray constructor   from@[native code]
> 10-04 19:10:29.717  3194  3212 I JsRuntimeHost:   app:///Scripts/tests.js:186:39

docEdub:
> Build is failing for Android JSC :-/
> 
> > 1 XMLHTTPRequest
> > 10-04 19:10:29.717  3194  3212 I JsRuntimeHost:        should load URL as array buffer TypedArray.from requires its this argument subclass a TypedArray constructor   from@[native code]
> > 10-04 19:10:29.717  3194  3212 I JsRuntimeHost:   app:///Scripts/tests.js:186:39

Ya, I'm looking at it now. Works on Android API 33 but not the CI's API 27.

bghgary:
That's weird. I don't see how API level has to do with the JS.

docEdub:
> That's weird. I don't see how Android API level has to do with the JS.

Works on my local machine with API 27, too, so I'm not sure what's going on now.

> Are you testing with V8 or JSC?

I don't know. Whatever the default is.

bghgary:
Default is V8 for Android. If you don't change to JSC, it probably won't repro.

bghgary:
Put this in the `gradle.properties` file:
> jsEngine=JavaScriptCore

PR #69, opened 2023-12-14
Save line/column/stack trace information about exceptions

### LINKED ISSUE #42: Reported error has no details when script loading fails
When loading a script, if the loading fails due to a parse error or some other error, the error output is not detailed enough to determine what is wrong. Here is an example when running Chakra:

```
Microsoft C++ exception: UnifiedRegex::ParseError at memory location 0x0000005DEC96B6B0.
Microsoft C++ exception: ParseExceptionObject at memory location 0x0000005DEC968ED0.
Uncaught Error: Unknown failure
```

It is possible to figure out the problem by stopping at the first exception, but it would be much better if the error message included the line number and what failed.

carolhmj:
fixes #42


Core/AppRuntime/Source/AppRuntime.cpp
```diff
@@ -36,6 +37,26 @@ namespace Babylon
...
+    int32_t getNumberPropertyFromError(Napi::Error error, const char* propertyName)
+    {
+        Napi::Value value = error.Get(propertyName);
+        if (value.IsUndefined())
+        {
+            return -1;
+        }
+        return value.ToNumber().Int32Value();
+    }
+
```
bghgary, line 59: These should be functions in an anonymous namespace at the top since they don't need access to class members. Also should be UpperCamelCase.

Core/AppRuntime/Source/AppRuntime.cpp
```diff
@@ -44,6 +65,24 @@ namespace Babylon
...
+                    std::string url = getStringPropertyFromError(error, "url");
+                    int32_t line = getNumberPropertyFromError(error, "line");
+                    int32_t column = getNumberPropertyFromError(error, "column");
+                    int32_t length = getNumberPropertyFromError(error, "length");
+                    std::string stack = getStringPropertyFromError(error, "stack");
+
+                    ss << "Error on line " << line << " and column " << column
+                       << ": " << msg << ". Length: " << length << ". Source: " << source << ". URL: " << url << ". Stack:" << std::endl << stack << std::endl;
+                    
+                    Napi::Error newError = Napi::Error::New(env, ss.str());
```
bghgary, line 83: This doesn't belong here. They should go in the `DefaultUnhandledExceptionHandler` implementation instead.

Core/Node-API/Source/js_native_api_v8.h
bghgary, line 1: Node-API uses a different style than we do. It would be best to keep the same style (i.e., two spaces indent, etc.)

Core/Node-API/Include/Shared/napi/napi-inl.h
bghgary, line 1: We should update Node-API to the latest before merging this PR and then this change will be included.

bghgary (changes requested):

CedricGuillemet:
fixes #69

Core/AppRuntime/Source/AppRuntime.cpp
CedricGuillemet: Fixed.

Core/Node-API/Include/Shared/napi/napi-inl.h
CedricGuillemet: I think this is done :)

Core/Node-API/Source/js_native_api_v8.h
CedricGuillemet: Fixed.

Core/AppRuntime/Source/AppRuntime.cpp
CedricGuillemet: I'm not sure to follow. `DefaultUnhandledExceptionHandler` is platform specific. this code is not.

CedricGuillemet: Done with rethrow. JSC is ok btw.

Core/AppRuntime/Source/AppRuntime_Android.cpp
```diff
@@ -14,6 +14,16 @@ namespace Babylon
     {
         std::stringstream ss{};
         ss << "Uncaught Error: " << error.what() << std::endl;
+
+        try
+        {
+            throw;
```
bghgary, line 20: This is assuming we are being called in a catch and ignoring the argument. Are we not allowed to do this?
```suggestion
            throw error;
```

Core/Node-API/Source/js_native_api_v8.h
```diff
@@ -404,13 +404,67 @@ class TryCatch : public v8::TryCatch {
...
+      v8::Local<v8::Value> exception = Exception();
+      EnhanceV8Exception(exception);
+      _env->last_exception.Reset(_env->isolate, exception);
     }
+    // [BABYLON-NATIVE-ADDITION]
   }
 
  private:
   napi_env _env;
+  // [BABYLON-NATIVE-ADDITION]
```
bghgary, line 418: nit: newline before this would be good

Core/AppRuntime/Source/AppRuntime_Android.cpp
```diff
@@ -14,6 +14,16 @@ namespace Babylon
...
+        }
+        catch (const Napi::Error& error)
+        {
+            ss << GetErrorInfos(error) << std::endl;
+        }
```
bghgary, line 25: Seems like this is all the same between platforms. Factor out into a shared function?

Core/AppRuntime/Include/Babylon/AppRuntime.h
```diff
@@ -46,5 +46,8 @@ namespace Babylon
 
         std::unique_ptr<WorkQueue> m_workQueue{};
         std::function<void(const std::exception&)> m_unhandledExceptionHandler{};
+        static std::string GetStringPropertyFromError(Napi::Error error, const char* propertyName);
+        static int32_t GetNumberPropertyFromError(Napi::Error error, const char* propertyName);
+        static std::string GetErrorInfos(Napi::Error error);
```
bghgary, line 51: These should probably be anonymous local functions in the cpp instead of the header (assuming you add a new shared function for the try/catch). We just need one function that is callable by the `DefaultUnhandledExceptionHandler` implementation.

bghgary (approved):

Core/Node-API/Source/js_native_api_v8.h
CedricGuillemet: fixed

Core/AppRuntime/Source/AppRuntime_Android.cpp
CedricGuillemet: `throw error;` crashes the app.

Core/AppRuntime/Source/AppRuntime_Android.cpp
CedricGuillemet: makes sense. fixed.

Core/AppRuntime/Include/Babylon/AppRuntime.h
CedricGuillemet: fixed

Core/AppRuntime/Source/AppRuntime_Android.cpp
bghgary: I think we have to capture the exception_ptr instead and pass that along.

PR #59, opened 2023-10-04
XCode15, NDK 25, up v8 and jsc

CedricGuillemet:
fixes #51

PR #70, opened 2024-01-05
Update Node-API to latest from node.js

carolhmj:


CedricGuillemet:
`napi_create_external_arraybuffer` checks the type of the array just created but it appears to not be a typed array.

CedricGuillemet:
Regarding performance with https://github.com/BabylonJS/BabylonNative/pull/1350 I get 2.7 seconds before this update 2.9-3 seconds with.

CedricGuillemet:
It's actually not a good idea to update node-api without updating v8/jsc. So, it's better to mix https://github.com/BabylonJS/JsRuntimeHost/pull/59 with this PR and changes in js_native_api_v8.cc are then not needed: v8 api is in sync with node-api.

But there are issues:
* sandbox. When disabled, I get this error on win32
```
# Fatal error in , line 0
# Embedder-vs-V8 build configuration mismatch. On embedder side sandbox is DISABLED while on V8 side it's ENABLED.
```
But, there is this check in node-api:
https://github.com/nodejs/node/blob/4ab63db9e207b5dbb6c046946716a49b6e2c3e53/src/node_api.cc#L1053

* this brings the 2nd issue regarding external buffer. It's currently disabled in this PR (see other comment) and it's a bit painful to put back in the code. I did try but I end up copying many files from node-api and it gets messy quite fast. Is there a way to just add what's needed here?

Core/Node-API/Source/js_native_api_v8.cc
```diff
@@ -2563,68 +2931,142 @@ napi_status napi_create_arraybuffer(napi_env env,
...
+    return napi_set_last_error(env, napi_no_external_buffers_allowed);
+#endif
+
+    v8::Isolate* isolate = env->isolate;
+
+    // The finalizer object will delete itself after invoking the callback.
+    v8impl::BufferFinalizer* finalizer =
+        v8impl::BufferFinalizer::New(env, finalize_cb, nullptr, finalize_hint);
+
+    v8::MaybeLocal<v8::Object> maybe = {};
```
CedricGuillemet, line 3009: internal buffer is not used in JSRuntimeHost, only used in BabylonNative. So, not an issue to get CI tests to pass. Finding the minimal set of code to get `node::Buffer::New` to work is difficult.

CedricGuillemet: Is it something that can be fixed with NODE_SPECIFIC ?

CedricGuillemet:
@bghgary 
I did another try at fixing the external buffer, with a copy this time. When testing ValidationTests, it crashes after 3 or 4 tests with this callstack :
image

Same crash happens with or without the self-deleting weak reference.

CedricGuillemet:
The crash I saw was because of async shaders. When running in release, everything is fine with Validation tests. Integration PR is here : https://github.com/BabylonJS/BabylonNative/pull/1353

Core/Node-API/Source/js_native_api_v8.cc
```diff
@@ -1,74 +1,228 @@
...
+inline napi_status V8NameFromPropertyDescriptor(
+    napi_env env,
+    const napi_property_descriptor* p,
+    v8::Local<v8::Name>* result) {
   if (p->utf8name != nullptr) {
     CHECK_NEW_FROM_UTF8(env, *result, p->utf8name);
   } else {
     v8::Local<v8::Value> property_value =
-      v8impl::V8LocalValueFromJsValue(p->name);
+        v8impl::V8LocalValueFromJsValue(p->name);
```
bghgary, line 225: There seems to be spacing issues in this file. This file uses two spaces for indent. Not sure if this is from node or did we do it?

Core/Node-API/Source/js_native_api_v8.cc
```diff
@@ -2208,59 +2498,130 @@ GEN_COERCE_FUNCTION(STRING, String, string)
...
 
   return napi_clear_last_error(env);
 }
 
-napi_status napi_get_value_external(napi_env env,
-                                    napi_value value,
-                                    void** result) {
-  CHECK_ENV(env);
+// Need NAPI_VERSION >= 8
+/*napi_status NAPI_CDECL napi_type_tag_object(napi_env env,
```
bghgary, line 2553: Did we comment this out or is this part of the node code?

Core/Node-API/Source/js_native_api_v8_internals.h
```diff
@@ -85,4 +59,7 @@ class PersistentToLocal {
 #define CHECK_LE(a, b) CHECK((a) <= (b))
 #endif
 
+#define NAPI_PRIVATE_KEY(context)                                      \
+  (v8::Private::New(context->GetIsolate()))
```
bghgary, line 63: Did you look into the perf? Are we okay with the private key perf?

.github/jobs/ios.yml
```diff
@@ -1,14 +1,20 @@
+parameters:
+  name: ''
+  vmImage: ''
+  xCodeVersion: ''
+  simulator: ''
```
PolygonalSun, line 5: Nit: With these parameters, it might be a good idea to format things closer to the [schema definition](https://learn.microsoft.com/en-us/azure/devops/pipelines/yaml-schema/parameters-parameter?view=azure-pipelines) for parameters:
```
# Example
parameters:
- name: name
   type: string
   default: ''
```
Also, since `name` is a property of a parameter, it might help to rename it to something like `jobName`.

.github/azure-pipelines.yml
```diff
@@ -71,6 +75,18 @@ jobs:
...
+
+  - template: jobs/ios.yml
+    parameters:
+      name: 'iOS_Xcode150'
+      vmImage: 'macOS-13'
+      xCodeVersion: 15.0
+      simulator: 'iPhone 14'
 
   # Linux
   - template: jobs/linux.yml
```
PolygonalSun, line 92: Nit: newline

Core/Node-API/packages.config
```diff
@@ -1,9 +1,9 @@
...
-  <package id="v8.symbols-v142-x86" version="8.4.371.15" targetFramework="native" />
-  <package id="v8-v142-x64" version="8.4.371.15" targetFramework="native" />
-  <package id="v8-v142-x86" version="8.4.371.15" targetFramework="native" />
+  <package id="v8.redist-v143-x64" version="11.9.169.4" targetFramework="native" />
+  <package id="v8.redist-v143-x86" version="11.9.169.4" targetFramework="native" />
+  <package id="v8.symbols-v143-x64" version="11.9.169.4" targetFramework="native" />
+  <package id="v8.symbols-v143-x86" version="11.9.169.4" targetFramework="native" />
+  <package id="v8-v143-x64" version="11.9.169.4" targetFramework="native" />
+  <package id="v8-v143-x86" version="11.9.169.4" targetFramework="native" />
 </packages>
```
PolygonalSun, line 9: Nit: newline

CedricGuillemet: fixed

Core/Node-API/Include/Shared/napi/napi-inl.h
```diff
@@ -463,6 +735,8 @@ inline bool Value::IsDataView() const {
   NAPI_THROW_IF_FAILED(_env, status, false);
   return result;
 }
+
+// [BABYLON-NATIVE-ADDITION]
```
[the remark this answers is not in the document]
CedricGuillemet: I didn't do any addition. I tried to remove some addition marked with `// [BABYLON-NATIVE-ADDITION]` with moderate success. I keep them unchanged.

Core/Node-API/Include/Shared/napi/node_api.h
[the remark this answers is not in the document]
CedricGuillemet: I could move the file in Source!

Core/Node-API/Source/js_native_api_v8.cc
CedricGuillemet: we did because of missing napi_type_tag.

Core/Node-API/Source/env.cc
```diff
@@ -5,7 +5,7 @@ namespace Napi
     Napi::Value Eval(Napi::Env env, const char* source, const char* sourceUrl)
     {
         napi_value result;
-        NAPI_THROW_IF_FAILED(env, napi_run_script(env, Napi::String::New(env, source), sourceUrl, &result));
+        NAPI_THROW_IF_FAILED(env, napi_run_script(env, Napi::String::New(env, source), &result));
```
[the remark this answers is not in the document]
CedricGuillemet: Indeed! It's back in.

Core/AppRuntime/V8Inspector/Source/V8InspectorAgent.cpp
```diff
@@ -234,7 +234,11 @@ namespace Babylon
         void connectFrontend()
         {
             session_ = inspector_->connect(
-                1, new ChannelImpl(agent_), v8_inspector::StringView());
+                1, new ChannelImpl(agent_), v8_inspector::StringView()
+#ifndef ANDROID
+                , v8_inspector::V8Inspector::kFullyTrusted
+#endif
```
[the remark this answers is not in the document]
CedricGuillemet: Comment added.

Core/Node-API/Include/Shared/napi/napi-inl.h
```diff
@@ -1947,119 +2726,200 @@ inline Buffer<T> Buffer<T>::New(napi_env env,
...
   }
-  return Buffer(env, value, length, data);
+  return Buffer(env, value);
 }
+#endif  // NODE_API_NO_EXTERNAL_BUFFERS_ALLOWED
 
 template <typename T>
-inline Buffer<T> Buffer<T>::Copy(napi_env env, const T* data, size_t length) {
+inline Buffer<T> Buffer<T>::NewOrCopy(napi_env env, T* data, size_t length) {
+#ifndef NODE_API_NO_EXTERNAL_BUFFERS_ALLOWED
```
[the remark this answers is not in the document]
CedricGuillemet: yes. with a different type of finalizer vs napi_finalize.

Core/Node-API/CMakeLists.txt
```diff
@@ -132,18 +139,18 @@ if(NAPI_BUILD_ABI)
...
-                PRIVATE v8::zlib)
-            
+                PRIVATE v8::third_party_zlib)
+
             if(CPU_ARCH STREQUAL "x64")
                 # Enable V8 Pointer Compression
                 # https://v8.dev/blog/pointer-compression
                 # https://stackoverflow.com/q/62921373
-                set(NAPI_DEFINITIONS PUBLIC V8_COMPRESS_POINTERS)
+                set(NAPI_DEFINITIONS PUBLIC V8_COMPRESS_POINTERS V8_ENABLE_SANDBOX)
```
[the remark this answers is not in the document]
CedricGuillemet: Spheres test gives 2.9seconds VS 2.7seconds without this change. But, idk the impact of private keys vs buffer copy. If I'm not wrong, buffer copy happens mostly when scene is setup, not at rendering time.

Core/Node-API/Source/js_native_api_v8.cc
CedricGuillemet: I don't remember doing a change for this particular function.

Core/Node-API/Include/Shared/napi/napi-inl.h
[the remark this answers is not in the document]
bghgary: Anything marked with `// [BABYLON-NATIVE-ADDITION]` is required. Don't remove them.

Core/Node-API/Include/Shared/napi/napi-inl.h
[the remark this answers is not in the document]
bghgary: Is this from node or did you add it?

Core/Node-API/Include/Shared/napi/node_api.h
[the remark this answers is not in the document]
bghgary: But it is still required? I don't understand why we want this file at all.

Core/Node-API/Include/Shared/napi/node_api_types.h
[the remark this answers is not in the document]
bghgary: These files should not be included at all.

Core/Node-API/Source/js_native_api_v8_internals.h
bghgary: But there was a perf regression right? So we have to do something.

Core/AppRuntime/V8Inspector/Source/V8InspectorAgent.cpp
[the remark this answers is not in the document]
bghgary: Can you file an issue on this to change this once the nuget package is available?

Core/Node-API/CMakeLists.txt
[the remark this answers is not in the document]
bghgary: We need to capture a performance profile.

Core/Node-API/Include/Shared/napi/node_api.h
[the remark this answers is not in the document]
CedricGuillemet: Indeed. Removed.

Core/Node-API/Include/Shared/napi/node_api_types.h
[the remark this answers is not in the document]
CedricGuillemet: Removed

Core/Node-API/Include/Shared/napi/napi-inl.h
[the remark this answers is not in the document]
CedricGuillemet: No. It's been here with Carol's first commit and here : https://github.com/nodejs/node-addon-api/blob/ea712094e3cfd0ce973631683a1c3275e451fa32/napi-inl.h#L2696

Core/Node-API/CMakeLists.txt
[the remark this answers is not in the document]
CedricGuillemet: Let's talk offline

Core/AppRuntime/V8Inspector/Source/V8InspectorAgent.cpp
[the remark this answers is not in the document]
CedricGuillemet: This will not compile when the API is updated with new nuget. I don't think it's worth adding an issue for that case.

Core/Node-API/Include/Shared/napi/napi-inl.h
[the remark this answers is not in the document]
CedricGuillemet: in `#ifndef NODE_ADDON_API_DISABLE_NODE_SPECIFIC` block -> not used in our case.

Core/Node-API/Source/js_native_api_v8.cc
CedricGuillemet: It looks like the NAPI_VERSION >= 8 check was absent from this file. The functions were commented out in previous commits. I've changed it to a #if to be consistent with function declaration.

Core/Node-API/Include/Shared/napi/js_native_api.h
```diff
@@ -528,8 +529,40 @@ node_api_post_finalizer(napi_env env,
...
 #endif  // NAPI_EXPERIMENTAL
 
+// [BABYLON NATIVE]: Promote this function from experimental as a stopgap
+NAPI_EXTERN napi_status napi_add_finalizer(napi_env env,
+                                           napi_value js_object,
+                                           void* native_object,
+                                           napi_finalize finalize_cb,
+                                           void* finalize_hint,
+                                           napi_ref* result);
+
```
bghgary, line 565: There is already another version in this file. We need to consolidate it somehow.

CedricGuillemet: I'm removing the definition with the `[BABYLON NATIVE]` tag as the function is part of napi update as well.

bghgary: Yes, but it's under a version define. We are currently only defining V3.

CedricGuillemet: It builds and unittests are ok with `NAPI_VERSION=5`  and  `NAPI_HAS_THREADS=0`

Core/Node-API/CMakeLists.txt
```diff
@@ -165,7 +165,8 @@ add_library(napi ${SOURCES})
 target_compile_definitions(napi
     PUBLIC NODE_ADDON_API_DISABLE_DEPRECATED
     PUBLIC NODE_ADDON_API_DISABLE_NODE_SPECIFIC
-    PUBLIC NAPI_VERSION=3
+    PUBLIC NAPI_VERSION=5
```
bghgary, line 168: I'm guessing we don't have these `NAPI_VERSION >= 5` functions implemented for all platforms?

```c++
// Dates
NAPI_EXTERN napi_status NAPI_CDECL napi_create_date(napi_env env,
                                                    double time,
                                                    napi_value* result);

NAPI_EXTERN napi_status NAPI_CDECL napi_is_date(napi_env env,
                                                napi_value value,
                                                bool* is_date);

NAPI_EXTERN napi_status NAPI_CDECL napi_get_date_value(napi_env env,
                                                       napi_value value,
                                                       double* result);

// Add finalizer for pointer
NAPI_EXTERN napi_status NAPI_CDECL napi_add_finalizer(napi_env env,
                                                      napi_value js_object,
                                                      void* finalize_data,
                                                      napi_finalize finalize_cb,
                                                      void* finalize_hint,
                                                      napi_ref* result);
```

Can you file an issue to figure this out?

Core/Node-API/Include/Shared/napi/napi.h
```diff
@@ -56,2055 +72,3178 @@ static_assert(sizeof(char16_t) == sizeof(wchar_t), "Size mismatch between char16
...
-    /// `sign_bit` must be provided and will be set to 1 if this BigInt is negative.
-    /// `*word_count` has to be initialized to the length of the `words` array.
-    /// Upon return, it will be set to the actual number of words that would
-    /// be needed to store this BigInt (i.e. the return value of `WordCount()`).
-    void ToWords(int* sign_bit, size_t* word_count, uint64_t* words);
-  };
-#endif  // NAPI_EXPERIMENTAL
+  MaybeOrValue<Value> RunScript(const char* utf8script) const;
+  MaybeOrValue<Value> RunScript(const std::string& utf8script) const;
+  MaybeOrValue<Value> RunScript(String script) const;
```
bghgary, line 321: This is new. It can maybe replace the Eval function we have? It just needs the source parameter.

Core/Node-API/CMakeLists.txt
CedricGuillemet: Correct! https://github.com/BabylonJS/JsRuntimeHost/issues/81

Core/Node-API/Include/Shared/napi/napi.h
CedricGuillemet: not really, `RunSript` calls `napi_run_script` with empty `source_url`.

bghgary: No need to block the PR for this, just asking. Maybe file an issue.

CedricGuillemet: Ho! I see! you mean replace the eval code in env.cc ?

CedricGuillemet: I did the change.

bghgary: This is good. Maybe even remove the Eval function completely, but then it will be breaking, so maybe not. :-)

Core/Node-API/Source/js_native_api_v8_internals.h
bghgary: I pushed a fix: https://github.com/BabylonJS/JsRuntimeHost/pull/70/commits/2b67aedc2f8d487b7521c1e58cd8835d05049315. This makes the perf better than main now.

bghgary (approved):

PR #80, opened 2024-02-21
Update CMakeLists.txt for consistency with recent changes

bghgary:


CMakeLists.txt
```diff
@@ -1,120 +1,111 @@
...
-        target_include_directories(asio INTERFACE "${asio_SOURCE_DIR}/asio/include")
-        set_property(TARGET asio PROPERTY FOLDER Dependencies)
-    endif()
-    if(NOT TARGET llhttp_static)
-        set_property(TARGET llhttp_static PROPERTY FOLDER Dependencies)
-    endif()
+    set_property(TARGET gmock PROPERTY FOLDER Dependencies/GoogleTest)
+    set_property(TARGET gmock_main PROPERTY FOLDER Dependencies/GoogleTest)
+    set_property(TARGET gtest PROPERTY FOLDER Dependencies/GoogleTest)
+    set_property(TARGET gtest_main PROPERTY FOLDER Dependencies/GoogleTest)
```
PolygonalSun, line 106: I think that Android is also using these so this may not be the right spot for this.

bghgary: I need to fix the condition.

bghgary: I'm not sure why Babylon Native's if condition is different than this.

CedricGuillemet (approved):

PR #85, opened 2024-03-21
Minor fixes

bghgary:
- Remove UWP tests to avoid confusion since they don't work
- Clean up the shared test code a bit
- Add SetConsoleOutputCP to Win32 test to get Unicode console output


PolygonalSun (approved):
LGTM

PR #86, opened 2024-04-03
Add `BABYLON_API` calling convention macro for Core and Polyfills classes

docEdub:
Without explicit calling convention modifers, consumers are required to compile with the same calling convention used in the released binaries. If a consumer compiles with the `__stdcall` calling convention, the linker won't resolve the function calls because the released binaries are built with the default `__cdecl` calling convention.

This change fixes the issue using an overridable `BABYLON_API` macro defined as `__cdecl` by default on Windows platforms or defined as empty on non-Windows platforms.


bghgary (approved):

SergioRZMasson (approved):
LGTM

PR #88, opened 2024-04-05
A bunch of miscellaneous fixes

bghgary:
- Move Node-API extensions from Babylon Native into JsRuntimeHost.
- Add helper method to Node-API extensions to get error string from either callstack or message if callstack is not available. Use this method where appropriate (e.g., DefaultUnhandledExceptionHandler) to avoid crash when callstack is not available.
- Add EnableDebugger flag to AppRuntime to allow user configuration.
- Add Node-API defines required for JsRuntimeHost directly into the headers instead in CMake to avoid having all consumers add defines to the compile options.
- Fix an issue in Napi::Error::New to allow Napi::Error to propogate properly.
- Don't throw in Chakra's napi_add_finalizer to avoid compiler warning.
- Fix an issue in XMLHttpRequest where some errors are being dropped.


Polyfills/XMLHttpRequest/Source/XMLHttpRequest.cpp
```diff
@@ -255,19 +254,18 @@ namespace Babylon::Polyfills::Internal
...
             SetReadyState(ReadyState::Done);
             RaiseEvent(EventType::LoadEnd);
 
             // Assume the XMLHttpRequest will only be used for a single request and clear the event handlers.
             // Single use seems to be the standard pattern, and we need to release our strong refs to event handlers.
             m_eventHandlerRefs.clear();
+        }).then(arcana::inline_scheduler, arcana::cancellation::none(), [env = info.Env()](arcana::expected<void, std::exception_ptr> result) {
+            if (result.has_error())
+            {
+                Napi::Error::New(env, result.error()).ThrowAsJavaScriptException();
```
ryantrem, line 267: I know you are only moving this code in this PR (not introducing it), but can you remind me why `ThrowAsJavaScriptException` is being used in this case? Since contractually XMLHttpRequest already has a way of communicating errors (the `error` event), I would have thought we'd see that being used here. I thought `ThrowAsJavaScriptException` was only used if you aren't using C++ exceptions at all or if there is effectively a fire and forget async call from JS to native (so there is no other way to communicate the error back to JS). Am I remembering that wrong? I also don't remember when on the JS side this thrown error will be observed. 🤔

bghgary: Hmm, I don't know. Ideally, we should match the browser's behavior.

ryantrem (approved):

PR #90, opened 2024-05-24
Debug traces + CB

CedricGuillemet:
Documentation will be part of BabylonNative.

2 levels of logging
# Code logging
Works are a console log. Enable Debug Trace (off by default) and set an outputter(none set by default) then call Trace.
```
SetTraceOutput([](const char *trace){
printf("%s\n", trace);
});
SetDebugTrace(true);
...
Trace("Error detected.");
```

# Babylon code instrumentation
Babylon code is instrumented and will use Trace function described earlier IF `JSRUNTIMEHOST_DEBUG_TRACE` preprocessor is defined. If it's not (default), no log will happen and no performance impact.

To summarize, user can log in his own code with Babylon logging API. Babylon Native and JSRuntimeHost code is instrumented but will only log if  `JSRUNTIMEHOST_DEBUG_TRACE` preprocessor is defined and logging is enabled and an outputer function is set.

Thread safety in the logging output is responsability of the user.

### DIFF
--- a/Core/Foundation/Include/Babylon/DebugTrace.h
+++ b/Core/Foundation/Include/Babylon/DebugTrace.h
@@ -1,16 +1,14 @@
 #pragma once
 #include <functional>
 
-#define DEBUG_TRACE_NOOP(...)
-
 #define _STRINGIFY(x) #x
 #define _TOSTRING(x) _STRINGIFY(x)
-#define _DEBUG_TRACE(_format, ...) Babylon::DebugTrace::Trace(__FILE__ "(" _TOSTRING(__LINE__) "): " _format "\n", ##__VA_ARGS__); 
+#define _DEBUG_TRACE(format, ...) Babylon::DebugTrace::Trace(__FILE__ "(" _TOSTRING(__LINE__) "): " format "\n", ##__VA_ARGS__); 
 
 #if JSRUNTIMEHOST_DEBUG_TRACE
 #   define DEBUG_TRACE _DEBUG_TRACE
 #else
-#   define DEBUG_TRACE(...) DEBUG_TRACE_NOOP()
+#   define DEBUG_TRACE(...)
 #endif // JSRUNTIMEHOST_DEBUG_TRACE
 
 namespace Babylon
@@ -18,9 +16,9 @@ namespace Babylon
     namespace DebugTrace
     {
         // off by default.
-        void SetDebugTrace(bool enabled);
+        void EnableDebugTrace(bool enabled);
         // equivalent of a printf that will use TraceOutput function to log. Need DebugTrace to be enabled.
-        void Trace(const char* _format, ...);
+        void Trace(const char* format, ...);
         // set the function used to output.
         void SetTraceOutput(std::function<void(const char* trace)> traceOutputFunction);
     }
--- a/Core/Foundation/Include/Babylon/DebugTrace.h
+++ b/Core/Foundation/Include/Babylon/DebugTrace.h
@@ -1,14 +1,17 @@
 #pragma once
 #include <functional>
 
+#if JSRUNTIMEHOST_DEBUG_TRACE
 #define _STRINGIFY(x) #x
 #define _TOSTRING(x) _STRINGIFY(x)
 #define _DEBUG_TRACE(format, ...) Babylon::DebugTrace::Trace(__FILE__ "(" _TOSTRING(__LINE__) "): " format "\n", ##__VA_ARGS__); 
+#define _DEBUG_RUN(lambda) lambda();
 
-#if JSRUNTIMEHOST_DEBUG_TRACE
-#   define DEBUG_TRACE _DEBUG_TRACE
-#else
-#   define DEBUG_TRACE(...)
+#define DEBUG_TRACE _DEBUG_TRACE
+#define DEBUG_RUN _DEBUG_RUN
+#else // No debugging trace enabled
+#define DEBUG_TRACE(...)
+#define DEBUG_RUN(...)
 #endif // JSRUNTIMEHOST_DEBUG_TRACE
 
 namespace Babylon
--- a/Core/Foundation/Include/Babylon/DebugTrace.h
+++ b/Core/Foundation/Include/Babylon/DebugTrace.h
@@ -1,18 +1,15 @@
 #pragma once
 #include <functional>
 
-#if JSRUNTIMEHOST_DEBUG_TRACE
+#ifdef BABYLON_DEBUG_TRACE
 #define _STRINGIFY(x) #x
 #define _TOSTRING(x) _STRINGIFY(x)
 #define _DEBUG_TRACE(format, ...) Babylon::DebugTrace::Trace(__FILE__ "(" _TOSTRING(__LINE__) "): " format "\n", ##__VA_ARGS__); 
-#define _DEBUG_RUN(lambda) lambda();
 
 #define DEBUG_TRACE _DEBUG_TRACE
-#define DEBUG_RUN _DEBUG_RUN
 #else // No debugging trace enabled
 #define DEBUG_TRACE(...)
-#define DEBUG_RUN(...)
-#endif // JSRUNTIMEHOST_DEBUG_TRACE
+#endif // BABYLON_DEBUG_TRACE
 
 namespace Babylon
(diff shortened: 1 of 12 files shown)


Core/Foundation/Include/Babylon/DebugTrace.h
```diff
@@ -0,0 +1,27 @@
+#pragma once
+#include <functional>
+
+#define DEBUG_TRACE_NOOP(...)
```
bghgary, line 4: Why do you need this?

Core/Foundation/Include/Babylon/DebugTrace.h
```diff
@@ -0,0 +1,27 @@
...
+
+#define _STRINGIFY(x) #x
+#define _TOSTRING(x) _STRINGIFY(x)
+#define _DEBUG_TRACE(_format, ...) Babylon::DebugTrace::Trace(__FILE__ "(" _TOSTRING(__LINE__) "): " _format "\n", ##__VA_ARGS__);
```
bghgary, line 8: nit: Why does `_format` have an underscore?

Core/Foundation/Include/Babylon/DebugTrace.h
```diff
@@ -0,0 +1,27 @@
...
+#endif // JSRUNTIMEHOST_DEBUG_TRACE
+
+namespace Babylon
+{
+    namespace DebugTrace
+    {
+        // off by default.
+        void SetDebugTrace(bool enabled);
+        // equivalent of a printf that will use TraceOutput function to log. Need DebugTrace to be enabled.
+        void Trace(const char* _format, ...);
```
bghgary, line 23: Ditto on `_format`

Core/Foundation/Include/Babylon/DebugTrace.h
```diff
@@ -0,0 +1,27 @@
...
+#else
+#   define DEBUG_TRACE(...) DEBUG_TRACE_NOOP()
+#endif // JSRUNTIMEHOST_DEBUG_TRACE
+
+namespace Babylon
+{
+    namespace DebugTrace
+    {
+        // off by default.
+        void SetDebugTrace(bool enabled);
```
bghgary, line 21: Maybe a better name?
```suggestion
        void EnableDebugTrace(bool enabled);
```

Core/Foundation/Include/Babylon/DebugTrace.h
```diff
@@ -0,0 +1,27 @@
...
+        // equivalent of a printf that will use TraceOutput function to log. Need DebugTrace to be enabled.
+        void Trace(const char* _format, ...);
+        // set the function used to output.
+        void SetTraceOutput(std::function<void(const char* trace)> traceOutputFunction);
```
bghgary, line 25: Do we need two functions? It seems like we can use one function for enabling/setting the callback. If someone wants to enable, they have to provide a callback?

Core/Foundation/Source/DebugTrace.cpp
```diff
@@ -0,0 +1,45 @@
...
+    {
+        static bool s_debugTraceEnabled{};
+        static std::function<void(const char* output)> s_traceOutputFunction;
+        static void TraceArgs(const char* format, va_list args)
+        {
+            char temp[8192];
+            auto len = vsnprintf(temp, sizeof(temp), format, args);
+            temp[len] = '\0';
+            s_traceOutputFunction(temp);
+        }
```
bghgary, line 18: Anonymous namespace doesn't need to be under the Babylon namespace.

Also, `static` is redundant inside an anonymous namespace.

Core/Foundation/Include/Babylon/DebugTrace.h
```diff
@@ -0,0 +1,27 @@
...
+
+#define _STRINGIFY(x) #x
+#define _TOSTRING(x) _STRINGIFY(x)
+#define _DEBUG_TRACE(_format, ...) Babylon::DebugTrace::Trace(__FILE__ "(" _TOSTRING(__LINE__) "): " _format "\n", ##__VA_ARGS__); 
+
+#if JSRUNTIMEHOST_DEBUG_TRACE
+#   define DEBUG_TRACE _DEBUG_TRACE
+#else
+#   define DEBUG_TRACE(...) DEBUG_TRACE_NOOP()
+#endif // JSRUNTIMEHOST_DEBUG_TRACE
```
bghgary, line 14: I don't think this is enough. We probably need something that allows us to check if debug trace is enabled so that people can wrap more complex debug code in an `#ifdef`. I guess we can check `JSRUNTIMEHOST_DEBUG_TRACE`, but maybe this name isn't good?

Core/Foundation/Source/DebugTrace.cpp
```diff
@@ -0,0 +1,45 @@
...
+{
+    namespace
+    {
+        static bool s_debugTraceEnabled{};
+        static std::function<void(const char* output)> s_traceOutputFunction;
+        static void TraceArgs(const char* format, va_list args)
+        {
+            char temp[8192];
+            auto len = vsnprintf(temp, sizeof(temp), format, args);
+            temp[len] = '\0';
```
bghgary, line 16: `vsnprintf` may return negative values if an error occurs. Maybe should check that.

CedricGuillemet: Fixed.

Core/Foundation/Include/Babylon/DebugTrace.h
CedricGuillemet: I've added DEBUG_RUN that takes a lambda. It works. Isn't it over kill compared to something like `assert` ? Do you have a use case for that? `JSRUNTIMEHOST_DEBUG_TRACE` might indeed not be a good name :) Naming is hard!

CedricGuillemet: using BABYLON_DEBUG_TRACE  instead. macro with lambda removed.

Core/Foundation/Source/DebugTrace.cpp
```diff
@@ -0,0 +1,52 @@
+#include <stdarg.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <functional>
+
+namespace
+{
+    bool s_debugTraceEnabled{};
```
bghgary, line 8: nit: `s_` is for static members of a class. This should probably be `g_` for global variables.

Tests/UnitTests/Android/app/src/main/cpp/JNI.cpp
```diff
@@ -21,6 +21,9 @@ Java_com_jsruntimehost_unittests_Native_javaScriptTests(JNIEnv* env, jclass claz
 
     android::global::Initialize(javaVM, context);
 
+    Babylon::DebugTrace::EnableDebugTrace(true);
+    Babylon::DebugTrace::SetTraceOutput([](const char* trace) { printf("%s\n", trace); });
```
bghgary, line 25: Ideally should flush after printf to avoid buffering. Maybe use std::cout instead which has a flush function.

Tests/UnitTests/Linux/App.cpp
```diff
@@ -1,6 +1,11 @@
 #include "../Shared/Shared.h"
+#include "Babylon/DebugTrace.h"
+#include <cstdio>
 
 int main()
 {
+    Babylon::DebugTrace::EnableDebugTrace(true);
+    Babylon::DebugTrace::SetTraceOutput([](const char* trace) { printf("%s\n", trace); });
```
bghgary, line 8: Ditto

bghgary (approved):

Core/Foundation/Source/DebugTrace.cpp
CedricGuillemet: Fixed.

Tests/UnitTests/Android/app/src/main/cpp/JNI.cpp
CedricGuillemet: Fixed.

Tests/UnitTests/Linux/App.cpp
CedricGuillemet: Fixed.

PR #91, opened 2024-05-24
V8 build fixes for x86

bghgary:
Pointer compression, etc. cannot be on for x86 builds when targeting V8.

V8 compile will currently fail if pointer compression is turned on:
> Pointer compression can be enabled only for 64-bit architectures

See https://github.com/v8/v8/blob/main/include/v8-internal.h#L165-L167

This change just makes it such that the macros are defined only for x64.


ryantrem:
Can you explain this a bit more?

bghgary:
I updated the description with more text.

ryantrem (approved):

PR #96, opened 2024-09-19
switch ArrayBuffer constructor

kmatzen:
I'd like to be able to incorporate [emnapi](https://emnapi-docs.vercel.app/) as an emscripten-based backend for JsRuntimeHost, but it has a limitation in that it can't properly expose the backing memory of an allocated ArrayBuffer as writable without an extra memory synchronization call after it has been constructed [[issue]](https://github.com/toyobayashi/emnapi/issues/128).  This doesn't seem to be an issue if the memory given to the ArrayBuffer constructor is fully initialized.  Would you mind if this were switched to pass the initialized memory as input rather than attempt to modify the backing buffer after construction?  Existing tests pass and BabylonNative on emscripten runs correctly.

### DIFF
--- a/Core/Node-API-Extensions/include/napi/pointer.h
+++ b/Core/Node-API-Extensions/include/napi/pointer.h
@@ -32,8 +32,7 @@ namespace Napi
         static Napi::Value Create(Napi::Env env, ReturnT (ClassT::*functionPtr)(ArgsT...))
         {
             using PointerT = decltype(functionPtr);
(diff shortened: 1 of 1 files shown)


bghgary (approved):
Should be okay, but this it not the only place we do this kind of array buffer management.

kmatzen:
No longer needed as it makes more sense to keep babylon in the js layer and wrap it with emscripten rather than to compile babylonnative down to wasm.

PR #97, opened 2024-09-19
emscripten backend

kmatzen:
First pass on integrating an emscripten backend.  The .a and .js files from emnapi are not yet integrated into cmake.


kmatzen:
No longer needed as it makes more sense to keep babylon in the js layer and wrap it with emscripten rather than to compile babylonnative down to wasm.

PR #100, opened 2025-01-09
AppRuntime iOS EnableDebugger

julapy:
hi, in order to run `JSGlobalContextSetInspectable` there needs to be check around it for iOS 16.4
````
        if (@available(iOS 16.4, *)) {
            JSGlobalContextSetInspectable(globalContext, m_options.EnableDebugger);
        }
````

i noticed that `AppRuntime_iOS.mm` was calling `RunEnvironmentTier` in `AppRuntime_JavaScriptCore.cpp`
however because its inside a cpp file, @available syntax was not allowed.

ive duplicated the `RunEnvironmentTier` function and moved it to `AppRuntime_iOS.mm` where @available check can run.

### DIFF
--- a/Core/AppRuntime/Source/AppRuntime_iOS.mm
+++ b/Core/AppRuntime/Source/AppRuntime_iOS.mm
@@ -13,6 +13,24 @@ namespace Babylon
         RunEnvironmentTier();
     }
 
+    void AppRuntime::RunEnvironmentTier(const char*)
+    {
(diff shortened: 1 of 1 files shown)


julapy:
on another note, might also be good show how AppRuntime Options can be enabled inside LibNativeBridge, so people can easily enable debugger.
````
    auto runtimeOptions = Babylon::AppRuntime::Options();
    runtimeOptions.EnableDebugger = true;
    runtime.emplace(runtimeOptions);
````

PR #101, opened 2025-01-17
Fix issues with reference tracking

ryantrem:
Background refresher: We want to be able to track weak references to arbitrary JS objects (created either from JavaScript or from native). When this is a weak reference, we need to know when the JS object has been GC'd such that the reference evaluates to null. To track when an object is GC'd, we want to register a finalizer callback. However, the only way to know when an object is finalized with JSC is to create a new class and provide a `finalize` callback. Given this, what we do is create a "sentinel" JS object that is attached/linked to the actual JS object of interest, and only referenced by that JS object. This way when the JS object of interest is GC'd, the attached/linked sentinel JS object with a finalizer callback is also GC'd and our finalizer is called and we know the JS object of interest has been GC'd.

There are two main issues being addressed in this PR:
1. Since we are relying on the finalizer of the sentinel object (not the actual object of interest), it is possible that the JS object of interest is GC'd, then a new JS object is instantiated at the same memory address, then we finally get the finalizer callback for the sentinel object. When this happens, and we also create a reference to the new JS object, our reference tracking thinks we are creating a reference to an existing JS object, then we finally get the finalizer for the sentinel object, and remove our reference tracking for the new object. Effectively this happens because we are using the memory address of the JS object as a globally and temporally unique identifier, when in fact that same identifier can refer to two different objects at different times, and the sentinel object finalizer can be called after this transition happens. To address this, we augment the reference tracking to use the `NativeInfo` instance memory address as a unique identifier (tracked in our active reference list, and already attached to the sentinel JS object), which is safe because that object is not deleted until after the sentinel is GC'd, so it will definitely be valid (unique) for at least as long as the JS object being tracked. When adding a ref as well as when accessing the underlying value of the ref, we check this new unique identifier to make sure the JS object (at the stored memory address) is actually still the same one.
2. Previously the way we were attaching/linking the sentinel object to the actual JS object was through the prototype chain. This has the benefit of being a well hidden implementation detail (e.g. it won't show up in a JS debugger, etc.), but if the prototype chain of an object is ever assigned to another object, it totally breaks the reference tracking. This could happen if code explicitly sets the prototype of one object to the prototype of another object, but it could also happen if you take a ref to a class and then instantiate the class (the class's prototype chain will implicitly be copied to the object instance). To make this more robust, we now instead attach/link the sentinel by setting it as a property on the object being tracked. We use a symbol as the key, and make it non-enumerable, readonly, and non-deletable. This should make it very hard to accidentally attach/link a sentinel object to a another object.

Since there is now an object id check in `napi_ref__::value`, this also exposed another napi bug, where we try to read properties off of a GC'd JS object. To fix this, I'm also taking this napi fix: https://github.com/nodejs/node-addon-api/pull/1607/files

### DIFF
--- a/Core/Node-API/Include/Shared/napi/napi-inl.h
+++ b/Core/Node-API/Include/Shared/napi/napi-inl.h
@@ -4513,7 +4513,7 @@ template <typename T>
 inline ObjectWrap<T>::~ObjectWrap() {
   // If the JS object still exists at this point, remove the finalizer added
   // through `napi_wrap()`.
-  if (!IsEmpty()) {
+  if (!IsEmpty() && !_finalized) {
     Object object = Value();
     // It is not valid to call `napi_remove_wrap()` with an empty `object`.
     // This happens e.g. during garbage collection.
@@ -4955,6 +4955,7 @@ inline void ObjectWrap<T>::FinalizeCallback(napi_env env,
                                             void* /*hint*/) {
   HandleScope scope(env);
   T* instance = static_cast<T*>(data);
+  instance->_finalized = true;
   instance->Finalize(Napi::Env(env));
   delete instance;
 }
--- a/Core/Node-API/Source/js_native_api_javascriptcore.cc
+++ b/Core/Node-API/Source/js_native_api_javascriptcore.cc
@@ -139,7 +139,7 @@ namespace {
     return reinterpret_cast<napi_value*>(const_cast<OpaqueJSValue**>(values));
   }
 
-  napi_env ToNapi(const JSContextRef context) {
+  napi_env ToNapi(const JSGlobalContextRef context) {
     return napi_env__::get(context);
   }
 
@@ -310,7 +310,7 @@ namespace {
                                          size_t argumentCount,
                                          const JSValueRef arguments[],
                                          JSValueRef* exception) {
-      napi_env env{ToNapi(ctx)};
+      napi_env env{ToNapi(JSContextGetGlobalContext(ctx))};
       ConstructorInfo* info = NativeInfo::Query<ConstructorInfo>(env, constructor, exception);
       if (*exception) {
         return nullptr;
@@ -405,7 +405,7 @@ namespace {
                                      size_t argumentCount,
                                      const JSValueRef arguments[],
                                      JSValueRef* exception) {
-      napi_env env{ToNapi(ctx)};
+      napi_env env{ToNapi(JSContextGetGlobalContext(ctx))};
       FunctionInfo* info = NativeInfo::Query<FunctionInfo>(env, function, exception);
       if (*exception) {
         return nullptr;
--- a/Core/Node-API/Source/js_native_api_javascriptcore.cc
+++ b/Core/Node-API/Source/js_native_api_javascriptcore.cc
@@ -644,7 +644,7 @@ namespace {
 }
 
 struct napi_ref__ {
-    void log(std::string action)
+    void log(std::string action) const
     {
         auto entry = ref_log.find(_value);
         if (entry == ref_log.end())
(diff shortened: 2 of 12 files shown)


Core/Node-API/Include/Shared/napi/napi-inl.h
```diff
@@ -4513,7 +4513,7 @@ template <typename T>
 inline ObjectWrap<T>::~ObjectWrap() {
   // If the JS object still exists at this point, remove the finalizer added
   // through `napi_wrap()`.
-  if (!IsEmpty()) {
+  if (!IsEmpty() && !_finalized) {
```
Allan121, line 4516: Did this extra check end up being relevant? Or just extra defensiveness just in case?

Core/Node-API/Source/js_native_api_javascriptcore.cc
```diff
@@ -577,18 +646,42 @@ namespace {
...
-  }
+  napi_ref__() = default;
+
+  // Copy semantics
+  napi_ref__(const napi_ref__&) = delete;
+  napi_ref__& operator=(const napi_ref__&) = delete;
+
+  // Move semantics
+  napi_ref__(napi_ref__&&) = delete;
+  napi_ref__& operator=(napi_ref__&&) = delete;
```
Allan121, line 657: Rookie question, but what are are the default copy/move semantics if they're not defined explicitly?

Core/Node-API/Include/Shared/napi/napi-inl.h
ryantrem: Yes, because the `Value()` function calls `napi_ref__::value()` which now accesses the object (to get the "object id").

Core/Node-API/Source/js_native_api_javascriptcore.cc
ryantrem: It's complicated... I believe it depends on what constructors have been explicitly included or deleted. @bghgary's recommendation was to just be explicit with all of them to avoid confusion.

bghgary, line 657:
```suggestion
  napi_ref__(napi_ref__&&) noexcept = delete;
  napi_ref__& operator=(napi_ref__&&) noexcept = delete;
```

Core/Node-API/Source/js_native_api_javascriptcore.cc
```diff
@@ -577,18 +646,42 @@ namespace {
...
+          auto entry{info->Env()->active_ref_values.find(value)};
+          // NOTE: The finalizer callback is actually on a "sentinel" JS object that is linked to the
+          // actual JS object we are trying to track. This means it is possible for the tracked object
+          // to be garbage collected and a new object created at the same memory address before we get
+          // the callback for the sentinel object finalizer. Guard against this by checking that the
+          // tracked object still has the same unique object id.
+          if (entry != info->Env()->active_ref_values.end() && entry->second == info->GetObjectId())
+          {
+            info->Env()->active_ref_values.erase(entry);
+          }
```
bghgary, line 677: nit: spacing
```suggestion
        auto entry{info->Env()->active_ref_values.find(value)};
        // NOTE: The finalizer callback is actually on a "sentinel" JS object that is linked to the
        // actual JS object we are trying to track. This means it is possible for the tracked object
        // to be garbage collected and a new object created at the same memory address before we get
        // the callback for the sentinel object finalizer. Guard against this by checking that the
        // tracked object still has the same unique object id.
        if (entry != info->Env()->active_ref_values.end() && entry->second == info->GetObjectId())
        {
          info->Env()->active_ref_values.erase(entry);
        }
```

Core/Node-API/Source/js_native_api_javascriptcore.cc
```diff
@@ -477,20 +529,34 @@ namespace {
...
-      JSObjectSetPrototype(env->context, ToJSObject(env, object), prototype);
-
+      JSObjectRef sentinel{JSObjectMake(env->context, info->_class, info)};
+      CHECK_NAPI(NativeInfo::Link<ReferenceInfo>(env, ToJSObject(env, object), sentinel));
       info->AddFinalizer(finalizer);
       return napi_ok;
     }
 
+    std::uintptr_t GetObjectId()
+    {
```
bghgary, line 556:
```suggestion
    std::uintptr_t GetObjectId() {
```

Core/Node-API/Source/js_native_api_javascriptcore.cc
```diff
@@ -623,12 +719,20 @@ struct napi_ref__ {
     return _count;
   }
 
-  napi_value value(napi_env env) const {
-    if (env->active_ref_values.find(_value) == env->active_ref_values.end()) {
-      return nullptr;
+  napi_status value(napi_env env, napi_value* result) const {
+    assert(_value);
+    if (env->active_ref_values.find(_value) != env->active_ref_values.end())
+    {
```
bghgary, line 725:
```suggestion
    if (env->active_ref_values.find(_value) != env->active_ref_values.end()) {
```

bghgary (approved):

PR #109, opened 2025-06-23
Small inclusion fix for <exception>

SergioRZMasson:
The file napi.h uses std::exception and std::exception_ptr. However, it does not include the <exception> header. This normally get taken care by other includes, but it can cause compiler errors depending on include order for consuming applications.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,5 +1,0 @@
-{
-    "files.associations": {
-        "exception": "cpp"
-    }
-}


Core/Node-API/Include/Shared/napi/napi.h
```diff
@@ -36,6 +36,7 @@
 #endif  // NAPI_HAS_THREADS
 #include <string>
 #include <vector>
+#include <exception>
```
ryantrem, line 39: Should this be conditional behind NAPI_CPP_EXCEPTIONS?

SergioRZMasson: Yes, it should. We also need to guard both ```static Error New(napi_env env, const std::exception& exception)``` and ```static Error New(napi_env env, const std::exception_ptr& exception_ptr)``` using NAPI_CPP_EXCEPTIONS  in the napi.h. We are already doing that for the implementation in the cpp file but I think we forgot to do it for the header.

PR #110, opened 2025-07-10
Fixed exception include

SergioRZMasson:
The include for exception need to be done after NAPI_CPP_EXCEPTIONS gets defined.

### DIFF
--- a/Core/Node-API/Include/Shared/napi/napi.h
+++ b/Core/Node-API/Include/Shared/napi/napi.h
@@ -36,9 +36,6 @@
 #endif  // NAPI_HAS_THREADS
 #include <string>
 #include <vector>
-#ifdef NAPI_CPP_EXCEPTIONS
-#include <exception>
-#endif  // NAPI_CPP_EXCEPTIONS
 
 // VS2015 RTM has bugs with constexpr, so require min of VS2015 Update 3 (known
 // good version)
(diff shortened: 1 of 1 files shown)


bghgary:
Is this a problem in the napi.h from node?

SergioRZMasson:
> Is this a problem in the napi.h from node?

I don't think so. We added: 

```
#ifdef NAPI_CPP_EXCEPTIONS
  // [BABYLON-NATIVE-ADDITION]
  static Error New(napi_env env, const std::exception& exception);
  // [BABYLON-NATIVE-ADDITION]
  static Error New(napi_env env, const std::exception_ptr& exception_ptr);
#endif  // NAPI_CPP_EXCEPTIONS
```

This brings std::exception and std::exception_ptr to the file. Therefore we need to include <expcetion>

bghgary:
> > Is this a problem in the napi.h from node?
> 
> I don't think so. We added:
> 
> ```
> #ifdef NAPI_CPP_EXCEPTIONS
>   // [BABYLON-NATIVE-ADDITION]
>   static Error New(napi_env env, const std::exception& exception);
>   // [BABYLON-NATIVE-ADDITION]
>   static Error New(napi_env env, const std::exception_ptr& exception_ptr);
> #endif  // NAPI_CPP_EXCEPTIONS
> ```
> 
> This brings std::exception and std::exception_ptr to the file. Therefore we need to include

We should wrap any changes to this file with `// [BABYLON-NATIVE-ADDITION]`

PR #112, opened 2025-07-16
TS + ES

CedricGuillemet:


Tests/package.json
```diff
@@ -2,8 +2,29 @@
...
+    "v8-android": "^7.8.2"
+  },
+  "devDependencies": {
+    "@babel/cli": "^7.27.0",
+    "@babel/core": "^7.26.10",
+    "@babel/preset-env": "^7.26.9",
+    "@types/chai": "^5.2.2",
+    "@types/mocha": "^10.0.10",
+    "chalk": "^5.4.1",
+    "esbuild": "^0.25.6",
```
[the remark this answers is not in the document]
CedricGuillemet: I can use webpack here instead. I have no opinion on which bundler to use.

CedricGuillemet:
> So the idea here is to write unit tests in TypeScript with es module imports, bundle it, and execute the bundled unit test code in JsRuntimeHost?

yes, exactly that.

Tests/package.json
```diff
@@ -2,8 +2,37 @@
...
+  "scripts": {
+    "build": "webpack",
+    "postinstall": "npm run build"
+  },
   "dependencies": {
-    "chai": "^4.3.4",
-    "mocha": "^9.2.2"
+    "chai": "5.2.0",
+    "jsc-android": "^241213.1.0",
+    "mocha": "11.1.0",
```
CedricGuillemet, line 12: Using this specific version as more recent ones (at least the one I've tested) errors:
```
ERROR in node:path
Module build failed: UnhandledSchemeError: Reading from "node:path" is not handled by plugins (Unhandled scheme).
Webpack supports "data:" and "file:" URIs by default.
You may need an additional plugin to handle "node:" URIs.
```

Tests/UnitTests/Scripts/tests.ts
```diff
@@ -1,6 +1,14 @@
-﻿mocha.setup({ ui: "bdd", reporter: "spec", retries: 5 });
+﻿import MochaDefault from "mocha";
+const Mocha = MochaDefault as typeof import("mocha");
```
ryantrem, line 2: Is this equivalent to:
```suggestion
﻿import * as MochaDefault from "mocha";
```

Tests/UnitTests/Scripts/tests.ts
```diff
@@ -60,7 +68,7 @@ describe("AbortController", function () {
 });
 
 describe("XMLHTTPRequest", function () {
-    function createRequest(method, url, body, responseType) {
+    function createRequest(method: string, url: string, body: any = undefined, responseType: any = undefined): Promise<XMLHttpRequest> {
```
ryantrem, line 71: Can you use `unknown` instead of `any`?
```suggestion
    function createRequest(method: string, url: string, body: uknown = undefined, responseType: unknown = undefined): Promise<XMLHttpRequest> {
```

Tests/UnitTests/Scripts/tests.ts
```diff
@@ -199,7 +207,7 @@ describe("setTimeout", function () {
     });
 
     it("should return an id greater than zero when given an undefined function", function () {
-        const id = setTimeout(undefined, 0);
+        const id = setTimeout(undefined as any, 0);
```
ryantrem, line 210: Maybe pass in an empty lambda?

Tests/UnitTests/Scripts/tests.ts
```diff
@@ -567,7 +583,7 @@ describe("URL", function () {
     it("should update href after URLSearchParams are changed", function () {
         // Augment URL with pathname and search
         const url = new URL("https://httpbin.org/en-US/docs?foo=1&bar=2");
-        url.searchParams.set("foo", 999);
+        url.searchParams.set("foo", 999 as any);
```
ryantrem, line 586: Why this cast? Should it just be
```suggestion
        url.searchParams.set("foo", "999");
```

Tests/UnitTests/Scripts/tests.ts
```diff
@@ -617,12 +633,13 @@ describe("URLSearchParams", function () {
     const paramsSet = new URLSearchParams("");
 
     it("should throw exception when trying to set with less than 2 parameters", function () {
+        // @ts-expect-error
```
ryantrem, line 636: Maybe add a comment on what the expected error is and why we need this?

Tests/babel-plugin-replace-bigint.js
```diff
@@ -0,0 +1,13 @@
+module.exports = function() {
```
ryantrem, line 1: Maybe some comments about why this is needed.

Tests/babel.config.json
ryantrem, line 1: How does this file relate to the .babelrc at the root?

Tests/package.json
```diff
@@ -2,8 +2,35 @@
   "name": "jsruntimehost",
   "version": "1.0.0",
   "private": true,
+  "scripts": {
+    "build": "webpack",
```
ryantrem, line 6: Maybe add a `"watch": "webpack -watch",` to make it easier for local dev?

Tests/webpack.config.js
```diff
@@ -0,0 +1,72 @@
...
+    rules: [
+        {
+          test: /\.(js|ts)$/,
+          exclude: (modulePath) => {
+            return (
+              /node_modules/.test(modulePath) &&
+              !/node_modules[\\/](?:@babylonjs|mocha|chai)/.test(modulePath)
+            );
+          },
+          use: 'babel-loader',
```
ryantrem, line 67: I think you can put Babel configuration inline here, rather than separate babel specific files. Would this make more sense?

If you prefer to keep it in a separate file, maybe `babel.config.js` (instead of json or rc) since it is more flexible?

Tests/babel-plugin-replace-bigint.js
CedricGuillemet: Fixed

Tests/package.json
CedricGuillemet: Fixed

Tests/UnitTests/Scripts/tests.ts
CedricGuillemet: Fixed.

Tests/UnitTests/Scripts/tests.ts
CedricGuillemet: It's done in the test just before.

Tests/UnitTests/Scripts/tests.ts
```diff
@@ -567,7 +582,7 @@ describe("URL", function () {
     it("should update href after URLSearchParams are changed", function () {
         // Augment URL with pathname and search
         const url = new URL("https://httpbin.org/en-US/docs?foo=1&bar=2");
-        url.searchParams.set("foo", 999);
+        url.searchParams.set("foo", "999");
```
bghgary, line 585: Should this be the same as the others where you cast to `any`? This is semantically different by passing in a string.

Tests/UnitTests/Scripts/tests.ts
```diff
@@ -416,7 +423,7 @@ if (hostPlatform !== "Unix") {
             };
 
             ws.onerror = (ev) => {
-                done(new Error(ev.message));
+                done(new Error(String(ev)));
```
bghgary, line 426: I tested this in a browser. It seems the event is just a generic event, so `ev.message` will be undefined. `String(ev)` will return `[object Event]` which isn't much better. The event itself does not seem to give more information. The console has a message when something fails. We should probably just say the websocket failed here.

```suggestion
                done(new Error("WebSocket failed"));
```

PR #123, opened 2025-11-05
Revert "Blob: Add factory function"

alexchuber:
Reverts BabylonJS/JsRuntimeHost#120

### DIFF
--- a/Polyfills/Blob/Source/Blob.cpp
+++ b/Polyfills/Blob/Source/Blob.cpp
@@ -4,10 +4,9 @@
 
 namespace Babylon::Polyfills::Internal
 {
-    static constexpr auto JS_BLOB_CONSTRUCTOR_NAME = "Blob";
-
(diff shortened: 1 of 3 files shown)


Polyfills/Blob/Source/Blob.cpp
```diff
@@ -18,30 +17,13 @@ namespace Babylon::Polyfills::Internal
                     InstanceAccessor("type", &Blob::GetType, nullptr),
                     InstanceMethod("text", &Blob::Text),
                     InstanceMethod("arrayBuffer", &Blob::ArrayBuffer),
-                    InstanceMethod("bytes", &Blob::Bytes),
+                    InstanceMethod("bytes", &Blob::Bytes)
```
alexchuber:
```suggestion
                    InstanceMethod("bytes", &Blob::Bytes),
```

bghgary (approved):

Issue #82, opened 2024-03-05
Node-API napi_remove_wrap is not being handled correctly


shial4:
If a finalize callback was associated with the wrapping, it will no longer be called when the JavaScript object becomes garbage-collected.

this is the bit from the Node-API docs that is not being handled correctly, from napi_remove_wrap (https://nodejs.org/api/n-api.html#napi_remove_wrap)

bghgary:
This is the case in JSC but we should check Chakra also. It should not be an issue for V8.

PR #130, opened 2025-12-09
Flesh out URL polyfill and add performance.now polyfill

ryantrem:
- Flesh out the `URL` polyfill (moslty by Claude).
- Add a bunch more unit tests for `URL` (by Claude).
- Add a polyfill for `performance.now()` (by Claude).
- Add some unit tests for `performance.now()` (by Claude).
- Uncomment `JSGlobalContextSetInspectable` and add platform and os version checks.
- Fix a bug where finalizers should be removed when `napi_remove_wrap` is called (per https://nodejs.org/api/n-api.html#napi_remove_wrap). Otherwise, a crash occurs if a constructor throws an error (which can happen with URL when the passed in string is not a valid url). This also addresses #82.
- Fix a bug where we try to assign to the result in `napi_remove_wrap` when the `void** result` is null. We just need to check `result` before assigning to it. `result` can be null in certain situations, such as if a constructor throws (same scenario as above).
- Fix a small issue where the webpack watch command does not work (at least on MacOS).

### DIFF
--- a/Polyfills/URL/Source/URL.cpp
+++ b/Polyfills/URL/Source/URL.cpp
@@ -3,6 +3,9 @@
 #include <regex>
 #include <optional>
 
+// NOTE: This is a platform agnostic implementation created with a lot of help from AI :)
+//       In the future, we may want to consider using platform-specific URL parsing APIs instead.
+
 namespace
 {
     // Parsed URL components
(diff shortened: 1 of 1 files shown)


Core/Node-API/Source/js_native_api_javascriptcore.cc
bghgary: Hmm, maybe we should just remove the wrapper completely?

PR #132, opened 2026-01-15
Quickjs

CedricGuillemet:


Core/AppRuntime/Source/WorkQueue.cpp
```diff
@@ -42,6 +54,15 @@ namespace Babylon
         while (!m_cancelSource.cancelled())
         {
             m_dispatcher.blocking_tick(m_cancelSource);
+#ifdef USE_QUICKJS
```
CedricGuillemet: How to do it properly?

.github/workflows/ci.yml
ryantrem, line 1: Do we want some Apple coverage with QuickJS?

Core/AppRuntime/Include/Babylon/AppRuntime.h
```diff
@@ -62,6 +62,7 @@ namespace Babylon
         void RunPlatformTier();
         void RunEnvironmentTier(const char* executablePath = ".");
         void Run(Napi::Env);
+        void SetPostTickCallback(std::function<void()> callback);
```
ryantrem: 1. I don't really understand why this is a callback setter, rather than just a private function that each implementation provides (like DrainMicroTasks).
2. I don't understand why we need both this and DrainMicroTasks. Couldn't we have just a single private function that runs after each tick, and Hermes uses that hook to do its drain micro task work, and quickjs uses it to do its execute pending job work?

Core/Node-API/Source/env_quickjs.cc
```diff
@@ -0,0 +1,141 @@
...
+    Env Attach(JSContext* context)
+    {
+        napi_env env_ptr{new napi_env__};
+        env_ptr->context = context;
+        env_ptr->current_context = env_ptr->context; 
+
+        // Get Object.prototype.hasOwnProperty
+        JSValue global = JS_GetGlobalObject(context);
+        JSValue object = JS_GetPropertyStr(context, global, "Object");
+        if (!JS_IsException(object) && JS_IsObject(object))
```
ryantrem: Shouldn't we throw if any of these fail? It seems like that's what we do for other js engines.

Polyfills/Performance/Source/Performance.cpp
```diff
@@ -33,6 +33,10 @@ namespace Babylon::Polyfills::Performance
             performance = Napi::Object::New(env);
             env.Global().Set(JS_INSTANCE_NAME, performance);
         }
+        else
+        {
+            return; // already defined (might be QuickJS built-in)
+        }
 
         performance.Set("now", Napi::Function::New(env, Now, "now"));
```
ryantrem, line 39: Why don't we just move this into the if block and skip the else?

Polyfills/WebSocket/Source/WebSocket.cpp
```diff
@@ -43,6 +43,18 @@ namespace Babylon::Polyfills::Internal
         , m_url(info[0].As<Napi::String>())
         , m_cancellationSource{std::make_shared<arcana::cancellation_source>()}
     {
+        // Keep the JS wrapper alive for the lifetime of the connection. The
```
ryantrem: Is this how it behaves in the browser? Why is this only just now needed, and not in the past?

.github/workflows/ci.yml
CedricGuillemet: Not sure if it's useful but build is quick so we can keep it.

Core/AppRuntime/Include/Babylon/AppRuntime.h
CedricGuillemet: Indeed, it can be simpler.

Polyfills/WebSocket/Source/WebSocket.cpp
CedricGuillemet: QuickJS is deterministic with ref counting: it's freed immediately when it reaches 0. Other engines with GC gave it more time to finish its async work.

ryantrem: Doesn't this mean the JS object is leaked though if it doesn't explicitly get closed?

CedricGuillemet: no, close is called when closed explicitely or when there is a failure.

ryantrem: But if you just do `new WebSocket("ws://localhost:8080")` will it leak? Nothing will cause close or a failure right?

CedricGuillemet: I just tried and the close callback gets called with a failure code and self reference is freed.

ryantrem: Sorry I'm confused... if you just do `new WebSocket("ws://localhost:8080")` and the close callback gets called, isn't that a different bug? Why would a freshly instantiated `WebSocket` get automatically closed? I think I must be missing something 🙈.

ryantrem (approved):

Polyfills/WebSocket/Source/WebSocket.cpp
ryantrem: @CedricGuillemet will see if there is a clean impl against https://websockets.spec.whatwg.org/#garbage-collection, otherwise it's probably ok for now.

PR #144, opened 2026-04-02
[TEST] Stress test for AppRuntime destructor deadlock race

bghgary:
**DO NOT MERGE** — This branch has an intentionally broken destructor to test whether CI can reproduce the deadlock race condition.

## What this tests

The AppRuntime destructor has a race condition where \cancel() + notify_all()\ from the main thread can lose the notification if it fires between \locking_tick\'s condition check and \condition_variable::wait()\. This causes a deadlock on \m_thread.join()\.

This branch:
- Merges WorkQueue into AppRuntime (from the fix branch)
- Reverts the destructor to the **old broken** cancel-from-outside pattern
- Adds a stress test that creates/destroys AppRuntime 10,000 times with a 5s timeout per iteration

The race window is ~nanoseconds, so it may not trigger on every run. CI VMs with shared resources and hypervisor preemption have a better chance of hitting it.

## Expected result
- **If the race triggers**: Test fails with \Deadlock detected on iteration N\
- **If the race doesn't trigger**: All tests pass (the bug exists but wasn't hit)


bghgary:
Superseded by a new PR with a deterministic test approach.

PR #145, opened 2026-04-02
Deterministic test proving AppRuntime destructor deadlock

bghgary:
Deterministic test proving the AppRuntime destructor has a deadlock race condition.

## The Bug

The destructor cancels from the main thread via cancel() + notify_all(). The notify fires **without holding the queue mutex**, so if the worker thread hasn't entered condition_variable::wait() yet, the signal is lost and join() hangs forever.

## How This Test Works

Uses an arcana testing hook (ARCANA_TESTING_HOOKS) that sleeps **while holding the queue mutex**, right before wait(). This guarantees the worker is between the condition check and wait() when the destructor fires.

- **Old code**: cancel() + notify_all() don't need the mutex, so they fire and the notification is lost while the worker sleeps. Worker enters wait() with no pending signal. **Deadlock.**
- **[Fixed code](https://github.com/bghgary/JsRuntimeHost/tree/fix/merge-workqueue-into-appruntime)**: Append(cancel) calls push() which needs the mutex, so it blocks until the worker enters wait(). Push completes, notifies, worker wakes. **No deadlock.**

## Expected Result

**This PR should FAIL CI** -- it intentionally has the old broken destructor to demonstrate the deadlock. The test will timeout after 5 seconds on iteration 0.

## Dependencies

- microsoft/arcana.cpp#59 (testing hooks)


bghgary:
Superseded — was based on the merged WorkQueue branch, not main.

Issue #61, opened 2023-10-10
When using UWP and V8, V8 debugging doesn't work


PolygonalSun:
When building against UWP, while using V8, V8 debugging will not work.  The projects will still build as long as the `JSRUNTIMEHOST_CORE_APPRUNTIME_V8_INSPECTOR` is disabled when the initial `cmake` project files are built.

PR #148, opened 2026-04-07
Add ThreadSanitizer CI jobs (Linux + macOS)

bghgary:
Add ThreadSanitizer (TSan) support for detecting data races at runtime.

## Changes

- **CMakeLists.txt**: New `ENABLE_THREAD_SANITIZER` option with `-fsanitize=thread`. Includes a mutual exclusion check with `ENABLE_SANITIZERS` since TSan and ASan cannot be combined.
- **.github/workflows/build-linux.yml**: New `enable-thread-sanitizer` input, passed as `ENABLE_THREAD_SANITIZER` to CMake. Wires `TSAN_OPTIONS` with suppression file and sets `JSC_useConcurrentGC=0` for the Run Tests step (see below).
- **.github/workflows/build-macos.yml**: New `enable-thread-sanitizer` input; wires `TSAN_OPTIONS`.
- **.github/workflows/ci.yml**: New `Ubuntu_ThreadSanitizer_clang` and `macOS_Xcode164_ThreadSanitizer` jobs.
- **.github/tsan_suppressions.txt**: Suppresses JSC-internal data races on Ubuntu (`called_from_lib:libjavascriptcoregtk`). These are allocator-level races in JSC's JIT worker threads that we cannot fix.

## Linux TSan: Concurrent GC Workaround

JSC on Linux uses `pthread_kill(tid, SIGUSR1)` + `sem_wait` in `WTF::Thread::suspend()` to stop mutator threads at GC safepoints. TSan's signal interception defers SIGUSR1 delivery indefinitely when the target is inside instrumented code; the handler never runs and the Collector Thread's `sem_wait` deadlocks. macOS JSC uses Mach `thread_suspend()` (no Unix signals) and is unaffected.

Setting `JSC_useConcurrentGC=0` on the Linux TSan job removes the dedicated Collector Thread; GC runs on the mutator without cross-thread signals. Reproduced locally in WSL Ubuntu: default configuration hung 9/20 runs; with `JSC_useConcurrentGC=0`, 0/30 runs hung.

## Dependency Updates

- **arcana.cpp**: Points to upstream `microsoft/arcana.cpp` (includes #61 — TSan-safe test hook callback mutex).
- **UrlLib**: Points to upstream `BabylonJS/UrlLib` at `d251ad44` (includes BabylonJS/UrlLib#27 — Apple WebSocket `@synchronized` fixes).

## Platform Support

TSan is supported on Linux and macOS with Clang/GCC. MSVC does not support TSan and Clang targeting Windows does not have a TSan runtime library.

## CI Impact

The TSan jobs run in parallel with other jobs and do not increase overall pipeline time.

## Status

- macOS TSan: passes clean
- Ubuntu TSan: JSC-internal races suppressed via `called_from_lib:libjavascriptcoregtk`; concurrent GC disabled to avoid TSan/signal deadlock


.github/azure-pipelines.yml
```diff
@@ -124,3 +124,10 @@ jobs:
       enableSanitizers: true
       CC: clang
       CXX: clang++
+
+  - template: jobs/linux.yml
+    parameters:
+      name: Ubuntu_ThreadSanitizer_clang
+      enableThreadSanitizer: true
+      CC: clang
+      CXX: clang++
```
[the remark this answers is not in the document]
bghgary: [Responded by Copilot on behalf of @bghgary] Agreed. This PR will stay draft until the pre-existing data races are fixed. Once TSan passes clean, it can be merged as a blocking check.

.github/jobs/linux.yml
```diff
@@ -34,3 +39,6 @@ jobs:
   - script: ./UnitTests
     workingDirectory: 'Build/ubuntu/Tests/UnitTests'
     displayName: 'Run Tests'
+    ${{if eq(parameters.enableThreadSanitizer, true)}}:
+      env:
+        TSAN_OPTIONS: suppressions=$(Build.SourcesDirectory)/.github/tsan_suppressions.txt
```
bghgary: Double check if the reason why we are getting errors is because of an old version of JSC. Check if there is a newer version of JSC and see if it's fixed. Mac version seems to be fine.

ryantrem (approved):

.github/jobs/linux.yml
bghgary: [Responded by Copilot on behalf of @bghgary]

Checked the CI logs for build 51994 (Ubuntu_ThreadSanitizer_clang):

- The job is already installing the newest JSC available on `ubuntu-latest` (Ubuntu 24.04 Noble): `libjavascriptcoregtk-4.1-0 2.50.4-0ubuntu0.24.04.1`. This is the latest update in Noble's apt repos.
- Upstream WebKitGTK is further ahead — current stable is **2.52.3** (April 2026). The 2.52.x series has additional JSC fixes, including [CVE-2025-43425](https://security.snyk.io/vuln/SNYK-UNMANAGED-WEBKITGTK-14101692) which is explicitly documented as a "Race Condition within a Thread" in WebKitGTK (DFG node-cloning race). Noble hasn't backported to 2.52 yet.
- This is consistent with macOS passing clean — Xcode 16.4's bundled JSC is much newer and includes these fixes.

So we're already on the latest apt-available JSC. To test whether 2.52.x eliminates the races we'd need either a WebKitGTK PPA or a source build on the CI, both of which add significant time/complexity to the job. I think the suppressions are still the right call for now, with a comment noting that newer upstream JSC likely fixes these and the suppressions can be revisited once Noble picks up 2.52.

bghgary:
[Posted by Copilot on behalf of @bghgary]

Converting to draft while we investigate a Linux-specific hang in the `Ubuntu_ThreadSanitizer_clang` job.

## Symptom

The `Run Tests` step hangs silently and the job is killed at the 30-min timeout. Recent history of the job on this PR:

| Commit | Duration | Outcome |
|---|---|---|
| 2df3c81 | 1.9 min | failure (TSan race) |
| 1201a3c | 1.8 min | failure (TSan race) |
| f1d4dd1 | 1.9 min | failure (TSan race) |
| e895125 | 14.8 min | cancelled |
| 7a963f5 | 18.4 min | cancelled |
| d28ca80 | 30.3 min | cancelled (timeout) |

When TSan trips on a race early, the run fails in ~2 min. When it doesn't, the tests hang. As the race fixes land, the hang is no longer masked.

## Step timing on d28ca80

| Step | Duration |
|---|---|
| Install packages | 18s |
| Configure CMake | 23s |
| Build Solution | 60s |
| **Run Tests** | **28+ min with no output → cancelled** |

Build + configure are ~100s. The whole job cost is in the hung test run.

## Last log lines

```
[log]     ✅ should load URL as array buffer
##[error]The Operation will be canceled.
```

No output for ~28 minutes after the last `XMLHTTPRequest` test. The next test in the suite is `should load a PLY file and parse vertex count from header using TextDecoder` (loads `app:///Assets/Halo_Believe.ply` via XHR, `arraybuffer`, then TextDecoder on the first 10 KB). Either that test hangs or something between tests does.

## Why this is Linux-specific

On the same commit:

- `Ubuntu_ThreadSanitizer_clang`: **28+ min hang**
- `macOS_Xcode164_ThreadSanitizer`: 2.7 min ✅
- `Ubuntu_Sanitizers_clang` (ASan/UBSan): 2.4 min ✅
- `Ubuntu_clang` (no sanitizer): 1.8 min ✅

Same tests, same TSan. So it is specific to TSan + Linux, i.e. TSan instrumentation layered on top of libcurl and/or libjavascriptcoregtk-4.1 on glibc. macOS uses NSURLSession + macOS JSC; both pass.

## Hypotheses

1. Deadlock in libcurl's threaded resolver or share-interface locks under TSan's pthread interposition (only manifests on glibc+TSan).
2. Deadlock inside `libjavascriptcoregtk-4.1` (2.50.4 on Noble) at a JS↔native boundary that TSan's instrumentation serializes differently.
3. Test-specific: the PLY asset is ~700 KB; a large `TextDecoder` decode path under TSan may interact badly with a JSC GC point while a UrlLib callback still holds a lock.

## Next steps

- Add a `timeout` + `gdb -batch -ex 'thread apply all bt'` wrapper around `./UnitTests` in `linux.yml` so the next hang produces a stack trace for every thread instead of nothing.
- Try reproducing locally in WSL Ubuntu 24.04 to get a deterministic stack trace without burning CI minutes.
- Once we have the stack, decide whether to fix in UrlLib, add a JSC-path suppression, skip the offending test under Linux+TSan, or split the Linux TSan job out of the per-PR pipeline.

Keeping per-PR TSan is still worthwhile — this initiative has already surfaced multiple real races. The bottleneck here is a hang to fix, not TSan overhead: the full instrumented pipeline completes in under 3 min on macOS.

bghgary:
## Root cause identified — and fixed in 5c8a4a6

**Short version:** JSC's concurrent GC on Linux uses SIGUSR1 + `sem_wait` to stop the world; TSan's signal interception delays SIGUSR1 delivery indefinitely, deadlocking the Collector Thread. Setting `JSC_useConcurrentGC=0` for the Ubuntu ThreadSanitizer job only removes the dedicated Collector Thread and eliminates the hang. macOS JSC uses Mach `thread_suspend()` instead of Unix signals, which is why the macOS TSan job has been passing all along.

### Evidence

Reproduced locally in WSL Ubuntu 24.04 (same clang-18, cmake 3.28.3, libjavascriptcoregtk-4.1-dev 2.50.4 as CI). Default config: **9/20 hangs**. Captured a hung inferior under gdb:

```
Thread 33 "ollector Thread":
 #5  __interceptor_sem_wait
 #6  WTF::Thread::suspend(WTF::ThreadSuspendLocker const&)
 #7–#21 [JSC GC stop-the-world path]

Thread 3 "UnitTests":  <pending SIGUSR1>  (signal queued for entire hang,
                                           only surfaced when gdb stopped
                                           execution — handler never ran)
```

The Collector was waiting on the post that Thread 3's SIGUSR1 handler was supposed to do; TSan never let the handler run.

### Fix verification

| config                         | runs | hangs |
| ------------------------------ | ---- | ----- |
| default (concurrent GC on)     | 20   | 9     |
| `JSC_useConcurrentJIT=0`       | 15   | 11    |
| `JSC_useJIT=0`                 | 15   | 4     |
| **`JSC_useConcurrentGC=0`**    | **30**  | **0**     |

### Changes in 5c8a4a6

1. Set `JSC_useConcurrentGC=0` in the `Ubuntu_ThreadSanitizer_clang` env block only (not gcc/Sanitizers/macOS TSan)
2. Revert the 30-min Linux-TSan timeout bump — expect this job to now finish in roughly the same time as macOS TSan (~3 min)

Marking ready for review. Auto-merge is still disabled — leaving that decision to you after checking CI.

Issue #27, opened 2023-05-04
Unable to debug JavaScript on macOS/iOS


bghgary:
The JS context doesn't show up in Safari web inspector.

bghgary:
After some investigation, I've determined that NSApplicationMain must be called in order for this to work on macOS. I think it's because there has to be an event loop for the web inspector to communicate with JSC.

bghgary:
The fix is complex and is captured by https://github.com/BabylonJS/BabylonNative/issues/1416

Issue #26, opened 2023-04-28
Websocket error message parameter missing from callback


lmaskati:
**Describe the bug**
Websocket error callback and close callback don't take in any parameters.

**Expected behavior**
An error message should be passed down from the source of the error into the error message callback. A close reason and close code should be passed into the close callback.

**Other**
 - Platform: Win32, Android, iOS

PR #161, opened 2026-04-17
Repro: issue #131 invalid-server WebSocket flake (DO NOT MERGE)

### LINKED ISSUE #131: WebSocket invalid server does not invoke onerror reliably
**Describe the bug**
Based on unit tests, it seems that WebSocket is not calling the `onerror` callback reliably for invalid servers.

**To Reproduce**
1. Run the "should trigger error callback with invalid server" unit test repeatedly (NOTE: This test was commented out in #130).
2. Observe that regardless of how long the timeout is, the test will fail about 1/3 of the time.

**Expected behavior**
`onerror` should be called reliably.

**Other**
 - This may only happen with Chakra, but I'm not sure.

bghgary:

**Do not merge.** This is a CI repro branch for https://github.com/BabylonJS/JsRuntimeHost/issues/131.

## Purpose

Issue #131 reports that the "should trigger error callback with invalid server" WebSocket test fails ~1/3 of the time. The test was commented out in #130. Since filing, the following WebSocket-related changes landed on `main`:

- #150 — Fix WebSocket test flake and update UrlLib (pulls in UrlLib #26 which consolidated `onError`/`onClose` dispatch on Windows + Apple).
- #160 — Fix WebSocket polyfill `close()`/`send()` spec compliance.
- (UrlLib #27 — Apple TSan fixes, not yet pulled in.)

Running the suite 30× locally on Windows + Chakra (at pre-#150 and at pre-#160 snapshots) produced **0 failures**, so the flake is not reproducible on Windows locally. This PR uncomments the test and runs **20 iterations** of it so CI (Linux/macOS) has a high probability of catching the flake if it still exists: P(all pass) = (2/3)^20 ≈ 0.03% if the 1/3 flake rate still holds.

## Expected outcome

- If CI goes green → our WebSocket changes (#150 / #160) likely fixed the issue; we can close #131 and keep the real test re-enabled (in a follow-up PR).
- If CI fails → the original flake persists; we keep the test commented out and continue investigating in UrlLib / per-platform impls.

## Not for merge

This PR adds 20 copies of the same test (spam), which is not an acceptable final state. If the repro confirms the fix, a follow-up PR will restore just the original single test.

### DIFF
--- a/Tests/UnitTests/Scripts/tests.ts
+++ b/Tests/UnitTests/Scripts/tests.ts
@@ -526,13 +526,18 @@ if (hostPlatform !== "Unix") {
             };
         });
 
-        // TODO: This is not working reliably: see https://github.com/BabylonJS/JsRuntimeHost/issues/131
-        // it("should trigger error callback with invalid server", function (done) {
-        //     const ws = new WebSocket("wss://caddddfd-ee88-4771-b293-8a8e13b330ee.com");
-        //     ws.onerror = () => {
-        //         done();
-        //     };
-        // });
+        // Repro for https://github.com/BabylonJS/JsRuntimeHost/issues/131
+        // Issue reports ~1/3 flake rate. Running 20 iterations so that if the
+        // flake still exists, P(all pass) = (2/3)^20 ≈ 0.03%.
+        for (let i = 1; i <= 20; i++) {
+            it(`should trigger error callback with invalid server (iter ${i})`, function (done) {
+                this.timeout(15000);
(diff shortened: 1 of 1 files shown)


bghgary:
[Posted by Copilot on behalf of @bghgary]

Repro served its purpose — 20 iterations of the invalid-server test passed on every platform/engine combo in [workflow run 24587856930](https://github.com/BabylonJS/JsRuntimeHost/actions/runs/24587856930), including Chakra where the flake was historically observed.

Superseded by #162, which restores the original single test. Closing.

Issue #60, opened 2023-10-10
UWP isn't currently able to work with V8


PolygonalSun:
When building against UWP, while using V8 as the JavaScript engine, this code will not function properly.  It appears that the reason for this is because the V8 files needed aren't being copied.  If they are manually copied, things will work.

CedricGuillemet:
I'm taking a look as I have to fix an issue with V8Inspector

Issue #51, opened 2023-09-20
Android NDK25


CedricGuillemet:
Next RN version will need NDK25
quick test shows issues with V8Inspector
https://github.com/BabylonJS/BabylonNative/pull/1279
Also, add the option to switch NDK version in pipeline

CedricGuillemet:
waiting so more infos with https://github.com/BabylonJS/JsRuntimeHost/issues/66

CedricGuillemet:
fixed with https://github.com/BabylonJS/JsRuntimeHost/pull/70

PR #15, opened 2023-03-10
Build instructions for Android

lmaskati:
Adding build instructions for Android.

PR #169, opened 2026-05-26
Add File / FileReader polyfill

bkaradzic-microsoft:
The File polyfill in BabylonNative currently lives as a JS shim (`Apps/Playground/Scripts/file_polyfill.js`, ~308 lines) that is `LoadScript`-ed from the Playground's `AppContext` (see BabylonJS/BabylonNative#1706). Per @bghgary's review on that PR, it should live in JsRuntimeHost alongside Blob, URL, WebSocket, XMLHttpRequest, TextDecoder, AbortController, etc., so every JsRuntimeHost consumer -- not only the BabylonNative Playground -- gets `File` and `FileReader`.

This PR adds a native C++ `Polyfills/File/` target that implements the WHATWG `File` and `FileReader` web APIs, plus 29 new Mocha unit tests (15 File + 14 FileReader) under `Tests/UnitTests/Scripts/tests.ts`.

### Surface area
* `File(parts, name, options)` constructor.
* `File` instance accessors: `size`, `type`, `name`, `lastModified`.
* `File` instance methods: `arrayBuffer()`, `text()`, `bytes()`.
* `File extends Blob`: the prototype chain is wired so `new File(...) instanceof Blob === true` (Babylon.js core branches on `instanceof Blob` in fileTools, Offline/database, abstractEngine, thinNativeEngine).
* `FileReader` with state constants `EMPTY`/`LOADING`/`DONE` (exposed both on the constructor and on instances, per WHATWG).
* `FileReader` readonly attributes `readyState`, `result`, `error` and the `onloadstart`/`onprogress`/`onload`/`onabort`/`onerror`/`onloadend` EventHandler attributes.
* `FileReader` events delivered via both the `onX` handler slots and `addEventListener` / `removeEventListener` / `dispatchEvent`.
* `FileReader.readAsText` / `readAsArrayBuffer` / `readAsDataURL` / `abort`.

### Implementation notes
* `File` delegates its byte storage to the existing `Blob` polyfill via `env.Global().Get("Blob")`, reusing Blob's `BlobPart` handling (`ArrayBuffer`, typed array, string, Blob). `Blob::Initialize` must run before `File::Initialize`.
* `FileReader`'s readonly attributes (`readyState`/`result`/`error`) and the `on*` handler slots are backed by **C++ state surfaced through prototype accessors**, not by writable JS properties, so script can neither overwrite them nor fool the state-machine checks. `result`/`error` are boxed on a persistent holder object rather than `Napi::Reference<Value>` slots, because `napi_create_reference` rejects primitive values on the real N-API backends (V8/JSC) and `readAsText`/`readAsDataURL` produce primitive string results.
* `readAsDataURL` base64-encodes via the `base-n` dependency (added as a top-level `FetchContent`), matching the encoder already used by other polyfills.
* An in-flight read keeps the `FileReader` alive (per spec) by anchoring its JS wrapper in a `shared_ptr<Napi::ObjectReference>` captured by the promise reactions -- externally owned, released automatically when the promise settles. A monotonic read id lets `abort()` / restart invalidate a stale continuation so a late-resolving `arrayBuffer()` promise cannot dispatch a phantom `load` after a user-initiated abort.
* The new `JSRUNTIMEHOST_POLYFILL_FILE` CMake option is on by default and gates building the target.

### Notable JSC-specific care
Two pieces of the implementation are shaped specifically to work correctly on JavaScriptCore:

1. **`FileReader.EMPTY`/`LOADING`/`DONE` are registered via `StaticValue` and `InstanceValue` descriptors inside the `DefineClass` property list, not via `func.Get("prototype").Set(...)` after class creation.** On JSC the napi shim's `ConstructorInfo::Create` defaults the constructor's `.prototype` to `Object.prototype`, so the latter pattern would pollute `Object.prototype` with `EMPTY/LOADING/DONE` keys and break `for..in` over plain objects throughout the runtime. The regression test `"does not pollute Object.prototype with EMPTY/LOADING/DONE"` pins this behaviour.

2. **The Blob-dependency guard in `File::Initialize` uses `IsUndefined()` rather than `IsFunction()`.** Some JSC builds (notably `libjavascriptcoregtk-4.1` on Linux) classify constructors created via `JSObjectMakeConstructor` as `typeof 'object'`, not `'function'`, so `napi_typeof` returns `napi_object` for them and an `IsFunction()` guard would silently early-return on those engines. `IsUndefined()` matches the guard used by Blob and works on V8, JSI, Chakra, iOS-JSC, Android-JSC, and Linux JSC.

### Testing
Locally `UnitTests.exe` (Win32 x64 Chakra Debug) passes all 180 JS tests (151 pre-existing + 29 new) and the native gtests, including the `Object.prototype`-pollution regression canary and the abort / late-resolve continuation guards. CI is green across all engines/platforms (V8, Chakra, JSI/JSC, iOS, Android, UWP, macOS, Ubuntu, sanitizers).

### After this merges
`BabylonJS/BabylonNative#1706` re-pins `CMakeLists.txt`'s `GIT_REPOSITORY`/`GIT_TAG` to the merged SHA and removes the JS shim.

### DIFF
--- a/Polyfills/File/Source/File.cpp
+++ b/Polyfills/File/Source/File.cpp
@@ -1,0 +1,171 @@
+#include "File.h"
+#include "FileReader.h"
+
+#include <Babylon/Polyfills/File.h>
+
+#include <chrono>
+#include <string>
+
+namespace Babylon::Polyfills::Internal
+{
+    namespace
+    {
+        constexpr auto JS_FILE_CONSTRUCTOR_NAME = "File";
+        constexpr auto JS_BLOB_CONSTRUCTOR_NAME = "Blob";
+    }
+
+    void File::Initialize(Napi::Env env)
+    {
+        auto global = env.Global();
+
+        // Refuse to install if the native Blob polyfill is not present:
+        // File delegates its byte storage to a Blob, so without it the
+        // constructor cannot produce useful instances. Use IsUndefined()
+        // rather than IsFunction() because some JavaScriptCore builds
+        // (notably libjavascriptcoregtk on Linux) classify constructors
+        // created via JSObjectMakeConstructor as typeof 'object', not
+        // 'function', so napi_typeof returns napi_object for them.
+        auto blob = global.Get(JS_BLOB_CONSTRUCTOR_NAME);
+        if (blob.IsUndefined() || blob.IsNull())
+        {
+            return;
+        }
+
+        // No-op if the runtime already provides a global File.
+        if (!global.Get(JS_FILE_CONSTRUCTOR_NAME).IsUndefined())
+        {
+            return;
+        }
+
+        Napi::Function func = DefineClass(
+            env,
+            JS_FILE_CONSTRUCTOR_NAME,
+            {
+                InstanceAccessor("size", &File::GetSize, nullptr),
+                InstanceAccessor("type", &File::GetType, nullptr),
+                InstanceAccessor("name", &File::GetName, nullptr),
+                InstanceAccessor("lastModified", &File::GetLastModified, nullptr),
+                InstanceMethod("arrayBuffer", &File::ArrayBuffer),
+                InstanceMethod("text", &File::Text),
+                InstanceMethod("bytes", &File::Bytes),
+            });
+
+        global.Set(JS_FILE_CONSTRUCTOR_NAME, func);
+    }
+
+    File::File(const Napi::CallbackInfo& info)
+        : Napi::ObjectWrap<File>{info}
+    {
+        auto env = info.Env();
+
+        // The WHATWG File constructor takes (fileBits, fileName, [options]).
+        // Both fileBits and fileName are required (USVString without
+        // `optional`), so missing either is a TypeError per WebIDL bindings.
+        if (info.Length() < 2)
+        {
+            throw Napi::TypeError::New(env,
+                "Failed to construct 'File': 2 arguments required, but only " +
+                std::to_string(info.Length()) + " present.");
+        }
+
+        Napi::Value parts = info[0];
+        Napi::Value name = info[1];
+        Napi::Value options = info.Length() > 2 ? info[2] : env.Undefined();
+
+        // USVString conversion: undefined -> "undefined", null -> "null",
+        // numbers/objects -> their .toString() representation. Napi::Value::
+        // ToString() routes through napi_coerce_to_string, which matches
+        // these semantics on all three engines.
+        m_name = name.ToString().Utf8Value();
+
+        // Default lastModified to the current wall clock in milliseconds,
+        // matching Date.now() semantics used by the JS File constructor.
+        m_lastModified = static_cast<double>(
+            std::chrono::duration_cast<std::chrono::milliseconds>(
+                std::chrono::system_clock::now().time_since_epoch())
+                .count());
+
+        auto blobOptions = Napi::Object::New(env);
+
+        if (options.IsObject())
+        {
+            auto optsObj = options.As<Napi::Object>();
+            if (optsObj.Has("type"))
+            {
+                blobOptions.Set("type", optsObj.Get("type"));
+            }
+            if (optsObj.Has("lastModified"))
+            {
+                auto lm = optsObj.Get("lastModified");
+                if (lm.IsNumber())
+                {
+                    m_lastModified = lm.As<Napi::Number>().DoubleValue();
+                }
+            }
+        }
+
+        Napi::Value partsArray;
+        if (parts.IsArray())
+        {
+            partsArray = parts;
+        }
+        else
+        {
+            partsArray = Napi::Array::New(env, 0);
+        }
+
+        // Delegate byte-buffer construction to the native Blob polyfill so
+        // we benefit from its existing BlobPart handling (ArrayBuffer,
+        // typed array, string, Blob).
+        auto blobCtor = env.Global().Get(JS_BLOB_CONSTRUCTOR_NAME).As<Napi::Function>();
+        auto blobInstance = blobCtor.New({partsArray, blobOptions});
+        m_blob = Napi::Persistent(blobInstance);
+    }
+
+    Napi::Value File::GetSize(const Napi::CallbackInfo&)
+    {
+        return m_blob.Value().Get("size");
+    }
+
+    Napi::Value File::GetType(const Napi::CallbackInfo&)
+    {
+        return m_blob.Value().Get("type");
+    }
+
+    Napi::Value File::GetName(const Napi::CallbackInfo& info)
+    {
+        return Napi::String::New(info.Env(), m_name);
+    }
+
+    Napi::Value File::GetLastModified(const Napi::CallbackInfo& info)
+    {
+        return Napi::Number::New(info.Env(), m_lastModified);
+    }
+
+    Napi::Value File::ArrayBuffer(const Napi::CallbackInfo&)
+    {
+        auto blob = m_blob.Value();
+        return blob.Get("arrayBuffer").As<Napi::Function>().Call(blob, {});
+    }
+
+    Napi::Value File::Text(const Napi::CallbackInfo&)
+    {
+        auto blob = m_blob.Value();
+        return blob.Get("text").As<Napi::Function>().Call(blob, {});
+    }
+
+    Napi::Value File::Bytes(const Napi::CallbackInfo&)
+    {
+        auto blob = m_blob.Value();
+        return blob.Get("bytes").As<Napi::Function>().Call(blob, {});
+    }
+}
+
+namespace Babylon::Polyfills::File
+{
+    void BABYLON_API Initialize(Napi::Env env)
+    {
+        Internal::File::Initialize(env);
+        Internal::FileReader::Initialize(env);
+    }
+}
--- a/Polyfills/File/Source/File.cpp
+++ b/Polyfills/File/Source/File.cpp
@@ -51,6 +51,25 @@ namespace Babylon::Polyfills::Internal
             });
 
         global.Set(JS_FILE_CONSTRUCTOR_NAME, func);
+
+        // File should behave as a subtype of Blob: any `instanceof Blob`
+        // check over a File must succeed. Babylon.js core does this in
+        // fileTools, Offline/database, abstractEngine, and thinNativeEngine,
+        // so without inheritance those checks silently fail and the
+        // serializer/loader paths take the wrong branch.
+        //
+        // The internal m_blob composition stays as an implementation
+        // detail; only the JS-visible prototype chain is wired so
+        // `new File(...) instanceof Blob === true`.
+        //
+        // Use Object.setPrototypeOf rather than __proto__ assignment so
+        // the operation routes through the engine's standard binding,
+        // working uniformly across V8 / JSC / Chakra.
+        auto objectCtor = global.Get("Object").As<Napi::Object>();
+        auto setPrototypeOf = objectCtor.Get("setPrototypeOf").As<Napi::Function>();
+        auto blobProto = blob.As<Napi::Object>().Get("prototype");
+        auto fileProto = func.Get("prototype");
+        setPrototypeOf.Call(objectCtor, {fileProto, blobProto});
     }
 
     File::File(const Napi::CallbackInfo& info)
--- a/Polyfills/File/Source/File.cpp
+++ b/Polyfills/File/Source/File.cpp
@@ -62,32 +62,46 @@ namespace Babylon::Polyfills::Internal
         // detail; only the JS-visible prototype chain is wired so
         // `new File(...) instanceof Blob === true`.
         //
-        // JSC-specific quirk: on JSC, the napi-defined class's
-        // `.prototype` JS property points to Object.prototype, not to
-        // the real prototype that JSObjectMakeConstructor uses for
-        // instances (the same quirk the FileReader constants block in
-        // FileReader.cpp documents in detail). Writing through
-        // `func.Get("prototype")` would either pollute Object.prototype
-        // or be rejected by setPrototypeOf with a TypeError.
+        // Two engine quirks force a dual-path approach:
         //
-        // Portable trick: instantiate a throwaway File, fetch its real
-        // prototype via Object.getPrototypeOf, and set THAT prototype's
-        // prototype to Blob.prototype. On V8 / Chakra the real prototype
-        // is also `func.prototype`, so this is correct everywhere.
-        // The temp instance is GC-eligible immediately after.
+        // - V8 / Chakra: `func.Get("prototype")` is the real prototype
+        //   that instances use, so `setPrototypeOf(func.prototype,
+        //   blobProto)` is the natural wire-up.
+        // - JSC: napi_define_class wraps JSObjectMakeConstructor, whose
+        //   `.prototype` JS property points to Object.prototype, not to
+        //   the real prototype JSC assigns to instances (the same quirk
+        //   the FileReader constants block in FileReader.cpp documents).
+        //   The direct call therefore either tries to mutate
+        //   Object.prototype's [[Prototype]] (immutable -> TypeError) or
+        //   silently does the wrong thing. The portable workaround is to
+        //   instantiate a throwaway File and reach the real prototype via
+        //   `Object.getPrototypeOf(tempInstance)`.
+        //
+        // Strategy: try the direct call first, verify by walking the
+        // prototype chain of a probe instance, and only fall back to the
+        // temp-instance trick if the chain isn't wired up. This keeps
+        // the cost minimal on V8/Chakra (one extra `instanceof Blob`
+        // walk) and still recovers on JSC.
         auto objectCtor = global.Get("Object").As<Napi::Object>();
         auto getPrototypeOf = objectCtor.Get("getPrototypeOf").As<Napi::Function>();
         auto setPrototypeOf = objectCtor.Get("setPrototypeOf").As<Napi::Function>();
         auto blobProto = blob.As<Napi::Object>().Get("prototype");
 
+        // Step 1: direct wire-up. Best-effort; on JSC this raises and we
+        // swallow.
+        auto funcProto = func.Get("prototype");
+        setPrototypeOf.Call(objectCtor, {funcProto, blobProto});
+        if (env.IsExceptionPending())
+        {
+            env.GetAndClearPendingException();
+        }
+
+        // Step 2: probe instance to verify the chain.
         auto emptyParts = Napi::Array::New(env, 0);
         auto initName = Napi::String::New(env, "");
         auto tempInstance = func.New({emptyParts, initName});
         if (env.IsExceptionPending())
         {
-            // Constructing the probe File raised; clear and skip the
-            // prototype-chain wire-up. Instances will not be Blob
-            // subtypes, but the rest of the polyfill stays installed.
             env.GetAndClearPendingException();
             return;
         }
@@ -99,13 +113,32 @@ namespace Babylon::Polyfills::Internal
             return;
         }
 
+        // Walk the chain to check if blobProto is reachable.
+        auto cursor = realProto;
+        while (!cursor.IsNull() && !cursor.IsUndefined())
+        {
+            if (cursor.StrictEquals(blobProto))
+            {
+                return; // direct wire-up succeeded
+            }
+            cursor = getPrototypeOf.Call(objectCtor, {cursor});
+            if (env.IsExceptionPending())
+            {
+                env.GetAndClearPendingException();
+                return;
+            }
+        }
+
+        // Step 3: fallback for engines where func.prototype != realProto
+        // (notably JSC). Set the chain on the real prototype we just
+        // discovered via getPrototypeOf(tempInstance).
         setPrototypeOf.Call(objectCtor, {realProto, blobProto});
         if (env.IsExceptionPending())
         {
-            // Some engines (notably older JSC builds) reject
-            // setPrototypeOf on the napi-internal prototype. Swallow so
-            // Initialize stays best-effort; `instanceof Blob` will be
-            // false on those engines but everything else still works.
+            // Some engines may reject setPrototypeOf even on the real
+            // napi-internal prototype. Swallow so Initialize stays
+            // best-effort; `instanceof Blob` will be false on those
+            // engines but everything else still works.
             env.GetAndClearPendingException();
         }
     }
--- a/Polyfills/File/Source/File.cpp
+++ b/Polyfills/File/Source/File.cpp
@@ -6,11 +6,6 @@
 #include <chrono>
 #include <string>
 
-#if defined(__has_include) && __has_include(<napi/env.h>)
-#include <napi/env.h>
-#define BABYLON_POLYFILL_USE_NAPI_JSI_EVAL 1
-#endif
-
 namespace Babylon::Polyfills::Internal
 {
     namespace
@@ -26,11 +21,11 @@ namespace Babylon::Polyfills::Internal
         //   instances use, so the direct setPrototypeOf succeeds and the
         //   probe path is skipped.
         // - JSC: `File.prototype` aliases Object.prototype (napi_define_class
-        //   wraps JSObjectMakeConstructor; same quirk the FileReader
-        //   constants block documents). The direct call throws TypeError,
-        //   the catch swallows it, and the probe path discovers the real
-        //   napi-internal prototype via `Object.getPrototypeOf(new File())`
-        //   and sets its [[Prototype]] to Blob.prototype.
+        //   wraps JSObjectMakeConstructor; see JsRH#172). The direct call
+        //   throws TypeError, the catch swallows it, and the probe path
+        //   discovers the real napi-internal prototype via
+        //   `Object.getPrototypeOf(new File())` and sets its [[Prototype]]
+        //   to Blob.prototype.
         //
         // Doing this in JS rather than via Napi::Function::Call avoids a
         // JSC napi-shim quirk where setPrototypeOf on Object.prototype
@@ -55,23 +50,25 @@ namespace Babylon::Polyfills::Internal
     {
         auto global = env.Global();
 
-        // Refuse to install if the native Blob polyfill is not present:
-        // File delegates its byte storage to a Blob, so without it the
-        // constructor cannot produce useful instances. Use IsUndefined()
-        // rather than IsFunction() because some JavaScriptCore builds
-        // (notably libjavascriptcoregtk on Linux) classify constructors
-        // created via JSObjectMakeConstructor as typeof 'object', not
-        // 'function', so napi_typeof returns napi_object for them.
-        auto blob = global.Get(JS_BLOB_CONSTRUCTOR_NAME);
-        if (blob.IsUndefined() || blob.IsNull())
+        // No-op if the runtime already provides a global File. Cheapest
+        // check, and the common path on platforms with a native File.
+        if (!global.Get(JS_FILE_CONSTRUCTOR_NAME).IsUndefined())
         {
             return;
         }
 
-        // No-op if the runtime already provides a global File.
-        if (!global.Get(JS_FILE_CONSTRUCTOR_NAME).IsUndefined())
+        // Require the native Blob polyfill: File delegates byte storage to
+        // a Blob, so without it the constructor cannot produce useful
+        // instances. Use IsUndefined() rather than IsFunction() because
+        // some JavaScriptCore builds (notably libjavascriptcoregtk on
+        // Linux) classify constructors created via JSObjectMakeConstructor
+        // as typeof 'object', not 'function', so napi_typeof returns
+        // napi_object for them.
+        auto blob = global.Get(JS_BLOB_CONSTRUCTOR_NAME);
+        if (blob.IsUndefined() || blob.IsNull())
         {
-            return;
+            throw Napi::Error::New(env,
+                "File polyfill requires the Blob polyfill to be installed first.");
         }
 
         Napi::Function func = DefineClass(
@@ -91,11 +88,7 @@ namespace Babylon::Polyfills::Internal
 
         // Wire File.prototype -> Blob.prototype via a tiny JS shim. See
         // the JS_PROTOTYPE_CHAIN_SHIM comment for engine-specific rationale.
-#if defined(BABYLON_POLYFILL_USE_NAPI_JSI_EVAL)
         Napi::Eval(env, JS_PROTOTYPE_CHAIN_SHIM, "JsRuntimeHost-File-PrototypeChainShim.js");
-#else
-        env.RunScript(JS_PROTOTYPE_CHAIN_SHIM, "JsRuntimeHost-File-PrototypeChainShim.js");
-#endif
         if (env.IsExceptionPending())
         {
             // The shim itself wraps every operation in try/catch, so this
--- a/Polyfills/File/Source/File.cpp
+++ b/Polyfills/File/Source/File.cpp
@@ -6,12 +6,49 @@
 #include <chrono>
 #include <string>
 
+#if defined(__has_include) && __has_include(<napi/env.h>)
+#include <napi/env.h>
+#define BABYLON_POLYFILL_USE_NAPI_JSI_EVAL 1
+#endif
+
 namespace Babylon::Polyfills::Internal
 {
     namespace
     {
         constexpr auto JS_FILE_CONSTRUCTOR_NAME = "File";
         constexpr auto JS_BLOB_CONSTRUCTOR_NAME = "Blob";
+
+        // Wire File.prototype to inherit from Blob.prototype so that
+        // `new File(...) instanceof Blob === true`. The shim runs entirely
+        // in JS so each engine's quirks are handled by the JS try/catch:
+        //
+        // - V8 / Chakra: `File.prototype` is the real prototype that
+        //   instances use, so the direct setPrototypeOf succeeds and the
+        //   probe path is skipped.
+        // - JSC: `File.prototype` aliases Object.prototype (napi_define_class
+        //   wraps JSObjectMakeConstructor; same quirk the FileReader
+        //   constants block documents). The direct call throws TypeError,
+        //   the catch swallows it, and the probe path discovers the real
+        //   napi-internal prototype via `Object.getPrototypeOf(new File())`
+        //   and sets its [[Prototype]] to Blob.prototype.
+        //
+        // Doing this in JS rather than via Napi::Function::Call avoids a
+        // JSC napi-shim quirk where setPrototypeOf on Object.prototype
+        // escapes as an uncaught error instead of being capturable via
+        // IsExceptionPending.
+        constexpr auto JS_PROTOTYPE_CHAIN_SHIM = R"JS(
+(function() {
+    if (typeof File !== 'function' || typeof Blob !== 'function') return;
+    var blobProto = Blob.prototype;
+    try { Object.setPrototypeOf(File.prototype, blobProto); } catch (e) {}
+    try {
+        var probe = new File([], '');
+        if (!(probe instanceof Blob)) {
+            Object.setPrototypeOf(Object.getPrototypeOf(probe), blobProto);
+        }
+    } catch (e) {}
+})();
+)JS";
     }
 
     void File::Initialize(Napi::Env env)
@@ -52,93 +89,18 @@ namespace Babylon::Polyfills::Internal
 
         global.Set(JS_FILE_CONSTRUCTOR_NAME, func);
 
-        // File should behave as a subtype of Blob: any `instanceof Blob`
-        // check over a File must succeed. Babylon.js core does this in
-        // fileTools, Offline/database, abstractEngine, and thinNativeEngine,
-        // so without inheritance those checks silently fail and the
-        // serializer/loader paths take the wrong branch.
-        //
-        // The internal m_blob composition stays as an implementation
-        // detail; only the JS-visible prototype chain is wired so
-        // `new File(...) instanceof Blob === true`.
-        //
-        // Two engine quirks force a dual-path approach:
-        //
-        // - V8 / Chakra: `func.Get("prototype")` is the real prototype
-        //   that instances use, so `setPrototypeOf(func.prototype,
-        //   blobProto)` is the natural wire-up.
-        // - JSC: napi_define_class wraps JSObjectMakeConstructor, whose
-        //   `.prototype` JS property points to Object.prototype, not to
-        //   the real prototype JSC assigns to instances (the same quirk
-        //   the FileReader constants block in FileReader.cpp documents).
-        //   The direct call therefore either tries to mutate
-        //   Object.prototype's [[Prototype]] (immutable -> TypeError) or
-        //   silently does the wrong thing. The portable workaround is to
-        //   instantiate a throwaway File and reach the real prototype via
-        //   `Object.getPrototypeOf(tempInstance)`.
-        //
-        // Strategy: try the direct call first, verify by walking the
-        // prototype chain of a probe instance, and only fall back to the
-        // temp-instance trick if the chain isn't wired up. This keeps
-        // the cost minimal on V8/Chakra (one extra `instanceof Blob`
-        // walk) and still recovers on JSC.
-        auto objectCtor = global.Get("Object").As<Napi::Object>();
-        auto getPrototypeOf = objectCtor.Get("getPrototypeOf").As<Napi::Function>();
-        auto setPrototypeOf = objectCtor.Get("setPrototypeOf").As<Napi::Function>();
-        auto blobProto = blob.As<Napi::Object>().Get("prototype");
-
-        // Step 1: direct wire-up. Best-effort; on JSC this raises and we
-        // swallow.
-        auto funcProto = func.Get("prototype");
-        setPrototypeOf.Call(objectCtor, {funcProto, blobProto});
+        // Wire File.prototype -> Blob.prototype via a tiny JS shim. See
+        // the JS_PROTOTYPE_CHAIN_SHIM comment for engine-specific rationale.
+#if defined(BABYLON_POLYFILL_USE_NAPI_JSI_EVAL)
+        Napi::Eval(env, JS_PROTOTYPE_CHAIN_SHIM, "JsRuntimeHost-File-PrototypeChainShim.js");
+#else
+        env.RunScript(JS_PROTOTYPE_CHAIN_SHIM, "JsRuntimeHost-File-PrototypeChainShim.js");
+#endif
         if (env.IsExceptionPending())
         {
-            env.GetAndClearPendingException();
-        }
-
-        // Step 2: probe instance to verify the chain.
-        auto emptyParts = Napi::Array::New(env, 0);
-        auto initName = Napi::String::New(env, "");
-        auto tempInstance = func.New({emptyParts, initName});
-        if (env.IsExceptionPending())
-        {
-            env.GetAndClearPendingException();
-            return;
-        }
-
-        auto realProto = getPrototypeOf.Call(objectCtor, {tempInstance});
-        if (env.IsExceptionPending())
-        {
-            env.GetAndClearPendingException();
-            return;
-        }
-
-        // Walk the chain to check if blobProto is reachable.
-        auto cursor = realProto;
-        while (!cursor.IsNull() && !cursor.IsUndefined())
-        {
-            if (cursor.StrictEquals(blobProto))
-            {
-                return; // direct wire-up succeeded
-            }
-            cursor = getPrototypeOf.Call(objectCtor, {cursor});
-            if (env.IsExceptionPending())
-            {
-                env.GetAndClearPendingException();
-                return;
-            }
-        }
-
-        // Step 3: fallback for engines where func.prototype != realProto
-        // (notably JSC). Set the chain on the real prototype we just
-        // discovered via getPrototypeOf(tempInstance).
-        setPrototypeOf.Call(objectCtor, {realProto, blobProto});
-        if (env.IsExceptionPending())
-        {
-            // Some engines may reject setPrototypeOf even on the real
-            // napi-internal prototype. Swallow so Initialize stays
-            // best-effort; `instanceof Blob` will be false on those
-            // engines but everything else still works.
+            // The shim itself wraps every operation in try/catch, so this
+            // should never fire. Belt-and-braces: clear so Initialize stays
+            // best-effort and the rest of the polyfill remains installed.
             env.GetAndClearPendingException();
         }
     }
--- a/Polyfills/File/Source/File.cpp
+++ b/Polyfills/File/Source/File.cpp
@@ -62,14 +62,52 @@ namespace Babylon::Polyfills::Internal
         // detail; only the JS-visible prototype chain is wired so
         // `new File(...) instanceof Blob === true`.
         //
-        // Use Object.setPrototypeOf rather than __proto__ assignment so
-        // the operation routes through the engine's standard binding,
-        // working uniformly across V8 / JSC / Chakra.
+        // JSC-specific quirk: on JSC, the napi-defined class's
+        // `.prototype` JS property points to Object.prototype, not to
+        // the real prototype that JSObjectMakeConstructor uses for
+        // instances (the same quirk the FileReader constants block in
+        // FileReader.cpp documents in detail). Writing through
+        // `func.Get("prototype")` would either pollute Object.prototype
+        // or be rejected by setPrototypeOf with a TypeError.
+        //
+        // Portable trick: instantiate a throwaway File, fetch its real
+        // prototype via Object.getPrototypeOf, and set THAT prototype's
+        // prototype to Blob.prototype. On V8 / Chakra the real prototype
+        // is also `func.prototype`, so this is correct everywhere.
+        // The temp instance is GC-eligible immediately after.
         auto objectCtor = global.Get("Object").As<Napi::Object>();
+        auto getPrototypeOf = objectCtor.Get("getPrototypeOf").As<Napi::Function>();
         auto setPrototypeOf = objectCtor.Get("setPrototypeOf").As<Napi::Function>();
         auto blobProto = blob.As<Napi::Object>().Get("prototype");
-        auto fileProto = func.Get("prototype");
-        setPrototypeOf.Call(objectCtor, {fileProto, blobProto});
+
+        auto emptyParts = Napi::Array::New(env, 0);
+        auto initName = Napi::String::New(env, "");
+        auto tempInstance = func.New({emptyParts, initName});
+        if (env.IsExceptionPending())
+        {
+            // Constructing the probe File raised; clear and skip the
+            // prototype-chain wire-up. Instances will not be Blob
+            // subtypes, but the rest of the polyfill stays installed.
+            env.GetAndClearPendingException();
+            return;
+        }
+
+        auto realProto = getPrototypeOf.Call(objectCtor, {tempInstance});
+        if (env.IsExceptionPending())
+        {
+            env.GetAndClearPendingException();
+            return;
+        }
+
+        setPrototypeOf.Call(objectCtor, {realProto, blobProto});
+        if (env.IsExceptionPending())
+        {
+            // Some engines (notably older JSC builds) reject
+            // setPrototypeOf on the napi-internal prototype. Swallow so
+            // Initialize stays best-effort; `instanceof Blob` will be
+            // false on those engines but everything else still works.
+            env.GetAndClearPendingException();
+        }
     }
 
     File::File(const Napi::CallbackInfo& info)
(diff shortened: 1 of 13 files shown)


Polyfills/File/Source/FileReader.cpp
```diff
@@ -0,0 +1,499 @@
...
+                resultValue = Napi::String::New(env, url);
+                break;
+            }
+            case ReadMode::BinaryString:
+            {
+                // Byte-per-char Latin-1 mapping, matching the spec for
+                // readAsBinaryString().
+                std::string out;
+                out.assign(reinterpret_cast<const char*>(data), size);
+                resultValue = Napi::String::New(env, out);
```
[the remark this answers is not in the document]
bkaradzic-microsoft: This is a real bug. `readAsBinaryString` is emitting UTF-8 bytes for code points ≥ 0x80 instead of Latin-1 code units 0x00–0xFF as required by the spec; e.g. byte `0xC2` ends up encoded as the two-byte UTF-8 sequence `0xC3 0x82` in the result string.

A fix was attempted (widen each byte directly into a UTF-16 code unit and return a UTF-16 string), but in combination with the other FileReader changes from this review pass it crashed Win32 Chakra with `STATUS_HEAP_CORRUPTION` deep inside `JsConstructObject`. JRH does not define `NAPI_CPP_EXCEPTIONS`, so `WrapCallback` does not translate C++ throws, and any unrelated exception path triggered while these wider FileReader edits were in place corrupted the heap non-deterministically. The whole set was reverted to the prior `67f9f5c` FileReader to get CI back to green.

This needs to re-land as an isolated change so the engine impact can be bisected. Tracking it outside this PR.

Polyfills/File/Source/FileReader.cpp
```diff
@@ -0,0 +1,499 @@
...
+        if (!typeValue.IsString())
+        {
+            return Napi::Boolean::New(env, false);
+        }
+
+        // Route through the internal Dispatch helper so on* handlers and
+        // addEventListener listeners actually fire for the event type.
+        Dispatch(env, info.This().As<Napi::Object>(), typeValue.As<Napi::String>().Utf8Value());
+        return Napi::Boolean::New(env, true);
+    }
```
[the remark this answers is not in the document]
bkaradzic-microsoft: Acknowledged. `dispatchEvent` constructs a fresh event each time instead of routing the same event object through both `on<type>` and any added listeners, so listeners observe different identities than spec requires.

A refactor that built the event once and passed it through both the on-property invoker and `target.dispatchEvent` was attempted and reverted for the same reason as the binstr fix — combined with the other FileReader changes, the cumulative diff crashed Win32 Chakra. Will re-introduce as a focused change.

Polyfills/File/Source/FileReader.cpp
```diff
@@ -0,0 +1,499 @@
...
+        jsThis.Set("readyState", Napi::Number::New(env, EMPTY));
+        jsThis.Set("result", env.Null());
+        jsThis.Set("error", env.Null());
+
+        // Initialize on* handler slots so they exist as enumerable, writable
+        // properties (consumers commonly assign to them after construction).
+        for (const auto* slot : ON_HANDLERS)
+        {
+            jsThis.Set(slot, env.Null());
+        }
```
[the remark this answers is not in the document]
bkaradzic-microsoft: Agreed in principle: `onload`/`onerror`/etc. should be WebIDL event-handler IDL attributes (accessors on the prototype that read/write a single internal slot), not enumerable own data properties added in the constructor.

Blocker: JRH's JSI napi shim does not implement `Napi::PropertyDescriptor` or `Object::DefineProperty` — both are commented out in `Core/Node-API-JSI/Include/napi/napi.h` (see ~line 114 and ~line 584), and the underlying `napi_define_properties` is not wired through. Without that, there's no portable way to install non-enumerable accessors that work across V8, JSC, Chakra and JSI. Implementing this cleanly requires first extending the JSI napi shim, which is out of scope for this PR.

Polyfills/File/Source/File.cpp
```diff
@@ -0,0 +1,163 @@
...
+
+        if (name.IsString())
+        {
+            m_name = name.As<Napi::String>().Utf8Value();
+        }
+        else if (!name.IsUndefined() && !name.IsNull())
+        {
+            m_name = name.ToString().Utf8Value();
+        }
+
```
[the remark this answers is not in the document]
bkaradzic-microsoft: Fixed in `fa84e76`:

- `m_name = name.ToString().Utf8Value()` — uniform USVString coerce, so `undefined` becomes `"undefined"` and `null` becomes `"null"` per WebIDL. Covered by the new `coerces undefined and null name per WebIDL USVString` test.
- `throw Napi::TypeError::New(env, "Failed to construct 'File': 2 arguments required, but only N present.")` when `info.Length() < 2`.

The exception-throwing path is preserved in the shipped polyfill but the test that exercises it (`throws when fewer than 2 arguments are passed`) is commented out with the same `// TODO: Uncomment this once the Node-API implementation for Chakra supports throwing errors from constructors.` marker that's already used five times in tests.ts (URL parse error cases). On Chakra's napi shim, throwing from inside an ObjectWrap constructor doesn't translate to a JS-level exception and corrupts the engine heap; until that's addressed the test would fail CI on Win32 Chakra. The C++ behavior matches WebIDL on every other engine.

Tests/UnitTests/Scripts/tests.ts
```diff
@@ -1356,6 +1356,330 @@ describe("TextDecoder", function () {
...
+            try {
+                expect(reader.result).to.equal("Hi");
+                done();
+            } catch (e) {
+                done(e);
+            }
+        };
+        reader.readAsBinaryString(blob);
+    });
+
```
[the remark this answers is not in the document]
bkaradzic-microsoft: Acknowledged. A `preserves bytes >= 0x80 verbatim as Latin-1 code units` test was added but reverted alongside the binstr fix in https://github.com/BabylonJS/JsRuntimeHost/pull/169#discussion_r3311392648 — the combined FileReader changes crashed Win32 Chakra. Will be re-introduced together with the Latin-1 emission fix as an isolated follow-up so the engine impact can be bisected.

bkaradzic-microsoft:
@bghgary friendly ping — this is the upstream half of [BabylonJS/BabylonNative#1706](https://github.com/BabylonJS/BabylonNative/pull/1706) (File / FileReader polyfill, which re-enables 19 GLTF/OBJ serializer tests in BN). Same flow as #171 (TextEncoder) which you helped me close out earlier today: the polyfill was moved here from the BN PR per [your review](https://github.com/BabylonJS/BabylonNative/pull/1706#pullrequestreview-4341209346) ("this should be a proper C++ polyfill under `Polyfills/` so every consumer gets it").

State just refreshed:
- Rebased onto current `upstream/main` (resolved a trivial overlapping-`describe()` conflict in `tests.ts` against the just-merged #171).
- 5 inline Copilot review comments all have my replies with fixes in commit `52f6338`. Notable bug-fix: `readAsBinaryString` now produces Latin-1 byte-per-code-unit per spec (was emitting UTF-8 of code points ≥ 0x80).
- `File` library builds standalone clean (`cmake --build build-notests --target File`).

`Polyfills/File/` sits next to `Polyfills/Blob/` and follows the same shape as `TextDecoder` / `TextEncoder` / `XMLHttpRequest` / `AbortController` / `URL` / `WebSocket`. No changes to existing files outside the standard wiring (`CMakeLists.txt` option, `Polyfills/CMakeLists.txt` add_subdirectory, `Tests/UnitTests/{CMakeLists, Shared/Shared.cpp, Scripts/tests.ts}` link + Initialize + tests).

Once this lands I'll bump BN #1706's pin back to `BabylonJS/JsRuntimeHost` and that PR is one tiny commit away from merge.

Polyfills/File/Source/FileReader.cpp
```diff
@@ -0,0 +1,499 @@
...
+            return;
+        }
+
+        // The .then() callbacks fire asynchronously, after StartRead() returns.
+        // Capture a strong reference to the FileReader's JS wrapper so the
+        // C++ ObjectWrap stays alive until the read settles even if the
+        // user dropped their reference. ObjectReference is move-only, so
+        // route it through shared_ptr to make the lambda copy-constructible
+        // (Napi::Function::New stores via std::function).
+        auto thisRef = std::make_shared<Napi::ObjectReference>(Napi::Persistent(jsThis));
```
ryantrem: It doesn't seem like we need shared_ptr for this though, can't each lambda just capture its own Napi::Persist?

Polyfills/File/Source/File.h
```diff
@@ -0,0 +1,30 @@
+#pragma once
+
+#include <napi/napi.h>
+
+#include <string>
+
+namespace Babylon::Polyfills::Internal
+{
+    class File final : public Napi::ObjectWrap<File>
```
ryantrem, line 9: `File` should behave as a subtype of `Blob` for WebAPI compatibility. As implemented, this looks like composition (`m_blob`) rather than inheritance, so `file instanceof Blob` may be false unless the prototype chain is explicitly wired up. Can we make `File` extend the polyfilled `Blob` type, or otherwise ensure `File.prototype` inherits from `Blob.prototype`?

ryantrem (commented):
Similar to my comment on the other PR, we may want to do a minimal implementation of the File polyfill that is just enough to support BJS usage (not sure if that is already the case).

Polyfills/File/Source/FileReader.cpp
bkaradzic-microsoft: Done in 4480ad8: replaced the shared_ptr<ObjectReference> trick with a member Napi::ObjectReference m_selfRef anchored across the in-flight read (matches the member-slot pattern used by WebSocket/XHR in this repo).

The reason the literal [thisRef = Napi::Persistent(jsThis)] mutable capture-per-lambda didn't fly here: Napi::ObjectReference is move-only, so capturing it makes the lambda move-only too — and the JSI shim's Napi::Function::New stores its callable through jsi::Function::createFromHostFunction, whose underlying HostFunctionType is std::function-shaped (CopyConstructible required). That's why we needed something heap-stored.

Member slot is cleaner than shared_ptr: lambdas now only capture POD + 	his, the FileReader's lifetime is anchored by m_selfRef while a read is in flight, and every terminal path (load / error / abort) resets it to break the self-cycle. Each lambda also guards on m_readId before dereferencing m_selfRef to handle the abort-then-late-resolve race.

Polyfills/File/Source/File.h
bkaradzic-microsoft: Done in 4480ad8: wired File.prototype → Blob.prototype via Object.setPrototypeOf in File::Initialize. 
ew File(...) instanceof Blob is now 	rue.

This matters because Babylon.js core branches on instanceof Blob in packages/dev/core/src/Misc/fileTools.pure.ts, Offline/database.pure.ts, Engines/abstractEngine.pure.ts, and Engines/thinNativeEngine.pure.ts — without the prototype chain wired up those checks silently fail for File inputs and the wrong branch is taken (the serializer round-trip path bypasses the Blob handling).

Internal m_blob composition stays as the implementation detail; only the JS-visible prototype chain is wired. New test in 	ests.ts:

```ts
it("is an instance of Blob (prototype chain wired up)", function () {
    const file = new File(["x"], "x.txt");
    expect(file instanceof Blob).to.equal(true);
    expect(file instanceof File).to.equal(true);
});
```

bkaradzic-microsoft:
@ryantrem on the minimization point — I surveyed Babylon.js usage and the only thing that wasn't already minimal is `FileReader.readAsBinaryString` (zero call sites across `packages/dev`, `packages/tools`, etc.). Dropped it in 4480ad8.

| API | BJS usage |
|---|---|
| `new File(parts, name, opts)` | inspector-v2, viewer, smartAssetSerializer, etc. |
| `instanceof File` | modelLoader, smartAssetSerializer, configurator |
| `instanceof Blob` | fileTools, Offline/database, abstractEngine, thinNativeEngine |
| `FileReader.readAsArrayBuffer` | fileTools, database |
| `FileReader.readAsText` | fileTools, smartAssetSerializer |
| `FileReader.readAsDataURL` | stringTools, nodeEditor, inspector |
| `FileReader.readAsBinaryString` | **zero call sites — removed** |

Dropping `readAsBinaryString` also makes the deferred Latin-1/UTF-8 follow-up unnecessary (the one that previously crashed Chakra during a combined-edit attempt — see the earlier thread).

Full set of changes in 4480ad8:
1. Drop `readAsBinaryString` + its enum case + its test.
2. `File` inherits `Blob` via prototype-chain wire-up + new `instanceof Blob` test.
3. Replace `shared_ptr<ObjectReference>` with member `Napi::ObjectReference m_selfRef` slot (matches the WebSocket/XHR pattern in this repo).

Local Win32 Chakra `File` target builds clean.

bkaradzic-microsoft:
CI on `35bfa81` is fully green (22/22 jobs across V8, Chakra, JSI, JSC on Linux/macOS/iOS/Android/Windows/UWP, plus Sanitizers + TSan).

The prototype-chain wire-up for `File extends Blob` went through three iterations:

1. `4480ad8` (initial review-fix commit): direct `Object.setPrototypeOf(func.prototype, blobProto)` in C++. Passed V8/Chakra; failed JSC because JSC's `napi_define_class` (which wraps `JSObjectMakeConstructor`) returns `Object.prototype` from `func.Get("prototype")` rather than the real napi-internal prototype — same quirk the `FileReader` constants block documents. Result: `setPrototypeOf` tried to mutate `Object.prototype`'s `[[Prototype]]` and threw TypeError.
2. `ac19d77`: temp-instance trick — `Object.getPrototypeOf(new File(...))` returns JSC's real napi-internal prototype, and that one accepts `setPrototypeOf`. Fixed JSC; broke Chakra (reason still unclear from the log alone — Chakra-specific behavior somewhere between `JsConstructObject` and `Object.getPrototypeOf`).
3. `9e7b333` (C++ dual-path direct + verify + fallback): regressed JSC because the JSC napi shim escapes the direct-path TypeError as `[Uncaught Error]` instead of capturing it via `IsExceptionPending`, so the C++ catch never gets the chance to fall through to the fallback.
4. `35bfa81` (current): move both paths into a small JS shim wrapped in `try/catch`. JS-level catch reliably traps the TypeError on every engine; direct path covers V8/Chakra; probe path (`instanceof Blob` check + `getPrototypeOf` + `setPrototypeOf` on the real napi-internal prototype) covers JSC. The shim is run once at `Initialize` via `env.RunScript` (Shared N-API) or `Napi::Eval` (JSI N-API).

Ball back in your court for re-review.

Polyfills/File/Source/FileReader.cpp
```diff
@@ -0,0 +1,496 @@
...
+            "onload",
+            "onabort",
+            "onerror",
+            "onloadend",
+        };
+
+        Napi::Value MakeEvent(Napi::Env env, const Napi::Object& jsThis, const std::string& eventType)
+        {
+            // Compute loaded/total best-effort from the current result, mirroring
+            // the JS polyfill that this C++ implementation replaces.
```
bghgary: The "JS polyfill that this C++ implementation replaces" reference is dangling — no such JS polyfill exists in this PR.

```suggestion
            // ProgressEvent contract: loaded/total reflect bytes processed. For
            // one-shot reads we don't track interim progress — report the final
            // byte count for both and leave lengthComputable=false.
```

Polyfills/File/Source/FileReader.cpp
```diff
@@ -0,0 +1,496 @@
...
+        // Important: do NOT set these via func.Get("prototype").Set(...) on
+        // the returned constructor. On JSC, JSObjectMakeConstructor defaults
+        // the constructor's .prototype property to Object.prototype, so
+        // writing through it pollutes Object.prototype and breaks any
+        // for..in over plain objects elsewhere in the runtime. The
+        // InstanceValue descriptors below go through napi_define_class's
+        // internal prototype lookup, which on JSC targets the napi-internal
+        // prototype (distinct from .prototype) and on V8 targets the
+        // function template's PrototypeTemplate — both correct, neither
+        // touches Object.prototype.
```
bghgary: Per JsRuntimeHost#173, dual `StaticValue` + `InstanceValue` is the correct WHATWG pattern for web polyfill constants — not a quirky workaround needing 14 lines of explanation. The JSC pitfall warning also evaporates once #172 lands. Drop the whole comment block:

```suggestion
```

Polyfills/File/Source/FileReader.cpp
```diff
@@ -0,0 +1,496 @@
...
+        jsThis.Set("readyState", Napi::Number::New(env, EMPTY));
+        jsThis.Set("result", env.Null());
+        jsThis.Set("error", env.Null());
+
+        // Initialize on* handler slots so they exist as enumerable, writable
+        // properties (consumers commonly assign to them after construction).
+        for (const auto* slot : ON_HANDLERS)
+        {
+            jsThis.Set(slot, env.Null());
+        }
```
bghgary: Same refactor as the `readyState`/`result`/`error` ask above — the `on*` handlers should be `InstanceAccessor` getter/setter pairs on the prototype, not enumerable writable own data props on every instance.

```cpp
// in DefineClass, one per ON_HANDLER slot:
InstanceAccessor("onload", &FileReader::GetOnLoad, &FileReader::SetOnLoad),
// ...
```

The setter does the event-handler IDL processing: store the new function (or null) in a `Napi::FunctionReference` member, and the dispatch path consults it alongside the listener list (or the setter wraps add/remove of the previous via the existing listener machinery).

Polyfills/File/Source/File.cpp
```diff
@@ -0,0 +1,223 @@
...
+#include <Babylon/Polyfills/File.h>
+
+#include <chrono>
+#include <string>
+
+#if defined(__has_include) && __has_include(<napi/env.h>)
+#include <napi/env.h>
+#define BABYLON_POLYFILL_USE_NAPI_JSI_EVAL 1
+#endif
```
bghgary: If #172 lands first and the JS shim goes away entirely, this macro goes too. If kept:

1. `__has_include(<napi/env.h>)` always succeeds — the header is transitively included for every engine via `<Babylon/Polyfills/File.h>`.
2. `..._JSI_EVAL` is misleading: `Napi::Eval` is declared for V8/Chakra/JSC too.
3. No rationale for preferring `Napi::Eval` over `env.RunScript()` — pick one and drop the `#if`.

Polyfills/File/Source/File.cpp
```diff
@@ -0,0 +1,223 @@
...
+    var blobProto = Blob.prototype;
+    try { Object.setPrototypeOf(File.prototype, blobProto); } catch (e) {}
+    try {
+        var probe = new File([], '');
+        if (!(probe instanceof Blob)) {
+            Object.setPrototypeOf(Object.getPrototypeOf(probe), blobProto);
+        }
+    } catch (e) {}
+})();
+)JS";
```
bghgary: If #172 lands first, this whole shim collapses to a direct `setPrototypeOf` in C++ (no JS eval, no try/catch fallback, no probe). Recommending we land #172 first and drop the shim entirely.

Polyfills/File/Source/File.cpp
```diff
@@ -0,0 +1,223 @@
...
+        if (blob.IsUndefined() || blob.IsNull())
+        {
+            return;
+        }
+
+        // No-op if the runtime already provides a global File.
+        if (!global.Get(JS_FILE_CONSTRUCTOR_NAME).IsUndefined())
+        {
+            return;
+        }
```
bghgary: Two nits:

1. **Ordering:** check "already provided" (lines 71–75) first — cheaper, and the common no-op path on platforms with native File.
2. **Silent skip on missing Blob:** consumers rely on File being installed when the polyfill is wired up; silent bail is hard to debug. Suggest throwing:
   ```cpp
   throw Napi::Error::New(env,
       "File polyfill requires the Blob polyfill to be installed first.");
   ```

Tests/UnitTests/Scripts/tests.ts
```diff
@@ -1394,6 +1394,348 @@ describe("TextEncoder", function () {
...
+    //     // File requires both fileBits and fileName per the WebIDL bindings.
+    //     // Browsers throw TypeError on missing arguments; the native polyfill
+    //     // must match that surface so consumers don't accidentally create a
+    //     // File with empty name when their call site is misspelled.
+    //     // Note: we only assert *that* it throws (not the specific error
+    //     // type), because the JSI napi shim wraps thrown Napi::TypeError as
+    //     // a generic JS Error when surfacing it across the host boundary.
+    //     expect(() => new (File as any)()).to.throw();
+    //     expect(() => new (File as any)([])).to.throw();
+    // });
```
bghgary: File a tracking issue for the Chakra napi-shim throw-from-constructor limitation and reference it in the TODO. Makes these tests re-enable-able atomically when the shim is fixed.

Polyfills/File/Source/FileReader.cpp
bghgary: [Responded by Copilot on behalf of @bghgary]

Agreed on move-only ruling out the literal capture. But `std::shared_ptr<Napi::ObjectReference>` IS a copyable handle to the move-only resource, and closes a real UAF `m_selfRef` doesn't:

```cpp
auto selfRef = std::make_shared<Napi::ObjectReference>(Napi::Persistent(jsThis));
auto onResolve = Napi::Function::New(env,
    [this, selfRef, myReadId, mode, contentType](const Napi::CallbackInfo& cb) {
        if (m_readId != myReadId) return;
        Napi::Value buf = cb.Length() > 0 ? cb[0] : cb.Env().Null();
        HandleReadResult(myReadId, mode, contentType, selfRef->Value(), buf);
    });
auto onReject = Napi::Function::New(env,
    [this, selfRef, myReadId](const Napi::CallbackInfo& cb) {
        if (m_readId != myReadId) return;
        Napi::Value err = cb.Length() > 0 ? cb[0]
            : static_cast<Napi::Value>(Napi::Error::New(cb.Env(), "FileReader: unknown error").Value());
        HandleReadError(myReadId, selfRef->Value(), err);
    });
```

**The UAF:** the lambda derefs `this` (in `m_readId`) BEFORE the token check. So `r.abort()` resets `m_selfRef` → user drops `r` → GC → `FileReader` destructor → underlying `Blob.arrayBuffer()` promise (engine-owned, separate from the wrapper) eventually settles → onResolve fires → `this->m_readId` is UAF. Narrow window (needs GC between abort and the queued microtask) but not closed by anything in the code. `shared_ptr<Persistent>` captured by the lambdas keeps the wrapper alive until both continuations are gone, so the token deref is always safe.

Bonus: drops the `m_selfRef` member, its 7-line comment, all four `.Reset()` calls, and the implicit "at most one in-flight read" invariant. `m_readId` stays — that's event suppression (abandoned reads shouldn't fire load/loadend), not lifetime.

Engine teardown is a separate concern, not solved by either approach (same exposure for WebSocket/XHR).

Polyfills/File/Source/FileReader.cpp
```diff
@@ -0,0 +1,496 @@
...
+
+    FileReader::FileReader(const Napi::CallbackInfo& info)
+        : Napi::ObjectWrap<FileReader>{info}
+    {
+        auto env = info.Env();
+        auto jsThis = info.This().As<Napi::Object>();
+
+        jsThis.Set("readyState", Napi::Number::New(env, EMPTY));
+        jsThis.Set("result", env.Null());
+        jsThis.Set("error", env.Null());
```
bghgary: These should be `InstanceAccessor` getters reading from C++-side state — not direct `jsThis.Set` in the constructor.

Per WHATWG IDL, `readyState` / `result` / `error` are `readonly attribute` → getter accessor on the prototype, non-writable from JS, non-enumerable. The PR makes them writable enumerable own data props on every instance.

```cpp
// in DefineClass:
InstanceAccessor("readyState", &FileReader::GetReadyState, nullptr),
InstanceAccessor("result",     &FileReader::GetResult,     nullptr),
InstanceAccessor("error",      &FileReader::GetError,      nullptr),
```

Add `m_readyState` / `m_result` / `m_error` C++ members; replace every `jsThis.Set("readyState", ...)` (L188, 317, 329, 368, 420, 468, 495) and `jsThis.Get("readyState")` (L178, 312, 410, 480) with member access — state-machine checks become tamper-proof (JS can't fool them by overwriting the property).

`InstanceAccessor` is supported by the JSI napi shim (`Core/Node-API-JSI/Include/napi/napi-inl.h:2575+`); no shim work needed.

Polyfills/File/Source/FileReader.cpp
```diff
@@ -0,0 +1,496 @@
...
+namespace Babylon::Polyfills::Internal
+{
+    namespace
+    {
+        constexpr auto JS_FILE_READER_CONSTRUCTOR_NAME = "FileReader";
+
+        // Inlined RFC 4648 base64 encoder. We don't pull in a third-party
+        // base64 library because no other JsRuntimeHost polyfill needs one
+        // and this is the only call site.
+        void EncodeBase64(const uint8_t* data, size_t size, std::string& out)
```
bghgary: Don't add a third base64 impl. JRH already has one in `Core/AppRuntime/V8Inspector/Source/V8InspectorUtils.cpp` (`Babylon::utils::base64_encode`), and BabylonNative pulls in `azawadzki/base-n` (header-only) for its Canvas/Window polyfills. Best path: promote `base-n` to JRH, have FileReader consume it; BN's direct `base-n` dep becomes redundant and can be dropped in a follow-up.

Tests/UnitTests/Scripts/tests.ts
```diff
@@ -1394,6 +1394,348 @@ describe("TextEncoder", function () {
     });
 });
 
+declare const File: any;
+declare const FileReader: any;
+
+describe("File", function () {
```
bghgary, line 1441: tests.ts is 1518 lines / 15 `describe` blocks across 10+ polyfills; adding `File` + `FileReader` here grows it further. Either file a JRH issue to split per-polyfill as a follow-up, or split in this PR — your call.

bghgary (changes requested):

**Recommend landing #172 first.** The JSC napi-shim bug it describes is the root cause of several workarounds in this PR — `JS_PROTOTYPE_CHAIN_SHIM` collapses to one direct `setPrototypeOf` call, the `Napi::Eval` / `env.RunScript` dispatch goes away, and the 14-line `FileReader` dual-exposure comment shrinks to one line. Quite a bit of code evaporates before merge if #172 lands first.

Other concerns inline.

bkaradzic-microsoft:
Filed #174 with the JSC napi shim fix for #172. It replaces the old `// BEGIN TODO ... END TODO` `[[Prototype]]`-kludge in `ConstructorInfo::Create` with a proper `.prototype` property setup, and stops `napi_define_class` from going through `napi_get_prototype` (which per spec returns `[[Prototype]]`, not `.prototype`). Includes a Blob-based regression test that asserts no instance members leak onto the global `Object.prototype`.

Once #174 lands, I'll rebase this branch onto main and:

- Collapse `JS_PROTOTYPE_CHAIN_SHIM` + the `Napi::Eval` / `env.RunScript` dispatch into a single direct C++ `Object.setPrototypeOf(File.prototype, Blob.prototype)` call (your #51 + #12 asks).
- Drop the 14-line FileReader `StaticValue`/`InstanceValue` comment block per #173 (your #109 ask).
- Address the remaining inline asks: dangling comment at FileReader.cpp:60 (your #60), `InstanceAccessor` getters/setters for `readyState`/`result`/`error` + `on*` handlers (your #140, #147), early-out + throw-on-missing-Blob reordering in `File::Initialize` (your #75), promote `azawadzki/base-n` from BabylonNative to JsRH for base64 (your #15), file a Chakra throw-from-constructor tracker issue (your #1479).

Will respond on the `shared_ptr` thread separately — your UAF argument is right; reverting the `m_selfRef` simplification.

Re: splitting `tests.ts` per-polyfill (your #1400) — happy to do it in this PR or as a follow-up; let me know which you prefer.

bkaradzic-microsoft:
Update: closing #174 without merging — the `JSObjectMakeConstructor` `.prototype` slot is `ReadOnly` per [`JSObjectRef.cpp:128`](https://github.com/WebKit/WebKit/blob/main/Source/JavaScriptCore/API/JSObjectRef.cpp#L128), so `JSObjectSetProperty` calls `put()` on it and throws on every JSC backend. The naive workaround `JSObjectMakeFunctionWithCallback` produces a non-constructable result. The actually-correct fix needs to switch to `JSObjectMake` + `JSClassDefinition` with both `callAsConstructor` and `callAsFunction`, plus a refactor of the `NativeInfo::Link` / `Query` sentinel pattern — too big to bundle with this PR.

So #169 keeps its JS-shim wire-up (`35bfa81`). I'll address your remaining inline asks against the current shape:

- **#60** dangling FileReader.cpp comment — fix as suggested.
- **#109** drop the 14-line dual-exposure comment block per #173 (#173 is fine to land independently as the WHATWG pattern; the JSC pitfall stays, just keep one line that points at it).
- **#140 / #147** `readyState`/`result`/`error`/`on*` → `InstanceAccessor` getters/setters with C++-side state. This is the meatiest refactor; will require touching 7 `jsThis.Set` + 4 `jsThis.Get` sites.
- **#15** promote `azawadzki/base-n` from BabylonNative to JsRH; FileReader consumes it. (Saves a third base64 impl; allows BN to drop the direct dep in a follow-up.)
- **#75** reorder `File::Initialize` (early-out first, throw on missing Blob).
- **#12** clean up the misleading `BABYLON_POLYFILL_USE_NAPI_JSI_EVAL` macro (pick `env.RunScript` or `Napi::Eval`, drop the `#if`).
- **#1479** file a JRH tracker for the Chakra throw-from-constructor limitation; reference it in the TODO.
- **shared_ptr UAF** revert — you're right, restoring `std::shared_ptr<Napi::ObjectReference>` per-lambda + dropping `m_selfRef`.
- **#1400** tests.ts split — happy to do either way; defaulting to follow-up unless you'd rather see it here.

bkaradzic-microsoft:
Pushed `3de01db` addressing the non-controversial mechanical nits from the review. CI [run 26855651151](https://github.com/BabylonJS/JsRuntimeHost/actions/runs/26855651151) — **22/22 green**.

Changes in this commit:

- **File.cpp**: drop the misleading `BABYLON_POLYFILL_USE_NAPI_JSI_EVAL` macro and `__has_include(<napi/env.h>)` guard. `Napi::Eval` is declared on every backend (Shared N-API has it via `Core/Node-API/Source/env.cc` as a thin wrapper around `env.RunScript`; JSI N-API declares it directly in `<napi/env.h>`). The include is also already pulled in transitively via `<Babylon/Polyfills/File.h>`, so the guard always succeeded. Now a single unconditional `Napi::Eval(...)` call.
- **File.cpp `Initialize`**: reorder so the "already provided" check runs first (cheap no-op, common path on hosts with a native `File`), and throw `Napi::Error` on missing `Blob` instead of silently returning. Consumers wiring up the polyfill expect it to be installed — silent failures are hard to diagnose.
- **FileReader.cpp `MakeEvent`**: drop the dangling "JS polyfill that this C++ implementation replaces" reference (there's no such JS polyfill in this PR); replaced with a one-liner describing the actual `ProgressEvent` `loaded`/`total` contract.
- **FileReader.cpp `DefineClass`**: collapse the 14-line dual-`StaticValue`/`InstanceValue` block down to one line pointing at #173, since per #173 that's the correct WHATWG IDL `const` member exposure pattern, not a workaround needing in-line justification.
- **tests.ts**: re-point the Chakra throw-from-constructor TODO to #175 (filed today) so the disabled `expect(() => new (File as any)()).to.throw()` test can be re-enabled atomically when the Chakra shim limitation is fixed.

Still to come (separate commits for review-clarity, in priority order):

- `shared_ptr<Napi::ObjectReference>` UAF revert in FileReader.cpp (your #r3344177855 ask) — restoring the per-lambda `shared_ptr` capture pattern that survives abort → drop → GC → engine-owned promise settles → onResolve fires.
- `InstanceAccessor` refactor for `readyState` / `result` / `error` / `on*` handlers (your #140 + #147 asks) — add C++ state members + `Napi::FunctionReference` slots, replace all the `jsThis.Set`/`.Get` sites.
- `base-n` promotion from BN to JRH (your #15 ask) — once it lands here, FileReader can drop the inline `EncodeBase64` and BN's direct dependency on the same lib goes away in a follow-up.
- `tests.ts` per-polyfill split (your #1400 note that said "your call") — happy to defer unless you'd rather see it now.

#172 / #174 update: as noted in the [follow-up comment on #174](https://github.com/BabylonJS/JsRuntimeHost/pull/174#issuecomment-4607956394), the proper JSC `.prototype` fix needs a `JSClassDefinition`-based callable+constructable rewrite that's too big to bundle here. The JS shim approach in this PR stays as-is.

bkaradzic-microsoft:
Rebased onto current `main` (which now includes [#177](https://github.com/BabylonJS/JsRuntimeHost/pull/177)) and dropped the JS_PROTOTYPE_CHAIN_SHIM workaround. `0bff942`.

Specifically, in `Polyfills/File/Source/File.cpp`:

- ✂️ Removed the `JS_PROTOTYPE_CHAIN_SHIM` constexpr (the 12-line JS try/catch dance with the `new File([], '')` probe fallback).
- ✂️ Removed the `Napi::Eval(env, JS_PROTOTYPE_CHAIN_SHIM, ...)` call + the `IsExceptionPending`/`GetAndClearPendingException` guard around it.
- ✂️ Removed the 30-line "V8 / Chakra / JSC behavior" comment block explaining why we couldn't use `Object.setPrototypeOf` directly.
- ✏️ Replaced with a 4-line direct `Object.setPrototypeOf(File.prototype, Blob.prototype)` via napi:
  ```cpp
  auto setPrototypeOf = env.Global().Get("Object").As<Napi::Object>()
      .Get("setPrototypeOf").As<Napi::Function>();
  setPrototypeOf.Call({ func.Get("prototype"), blob.As<Napi::Function>().Get("prototype") });
  ```

Regression coverage for the underlying issue lives in #177's `describe("napi class prototype isolation (#172)")` block (`tests.ts:1271`), which uses `Blob` to exercise exactly the JSC-shim territory that File no longer has to work around. `expect(file instanceof Blob).to.equal(true)` at `tests.ts:1564` still asserts the File→Blob chain end-to-end.

Net diff: `+17 / -42` lines in File.cpp. CI: [run 26909552279](https://github.com/BabylonJS/JsRuntimeHost/actions/runs/26909552279) (in progress).

Rebase was clean — no conflicts against #171 (TextEncoder) / #166 (console callstack) / #168 (UrlLib pin) despite touching some of the same files. Will report once CI lands.

Polyfills/File/Source/File.cpp
```diff
@@ -0,0 +1,191 @@
...
+        // `new File(...) instanceof Blob === true`. WHATWG specs File as
+        // a Blob subtype; BJS core (fileTools, Offline/database,
+        // abstractEngine, thinNativeEngine) branches on `instanceof Blob`
+        // and needs File inputs to satisfy that check.
+        //
+        // Relies on JsRH #177: prior to that fix the JSC napi shim aliased
+        // every napi class's `.prototype` to Object.prototype, so this
+        // assignment would have polluted every object globally. With #177
+        // in place each napi class has its own `.prototype` object and
+        // this is a plain prototype-chain edit.
```
bghgary: Drop the `#177` historical paragraph — it's part of main's git history now.

```suggestion
        // Wire File.prototype's [[Prototype]] to Blob.prototype so
        // `new File(...) instanceof Blob === true`. WHATWG specs File as
        // a Blob subtype; BJS core (fileTools, Offline/database,
        // abstractEngine, thinNativeEngine) branches on `instanceof Blob`
        // and needs File inputs to satisfy that check.
```

bghgary (commented):

Please merge `main` to pick up #177 + #179.

One related cleanup outside this PR's diff — `Tests/UnitTests/Scripts/tests.ts` lines 1292–1297 still has the `TODO(#178)` / `isPrototypeOf` workaround that #177 introduced for Chakra. #179 retired it; after merging main, restore the depth-1 check:

```diff
-        // TODO(#178): Chakra's napi_wrap interposes a hidden external object
-        // into the instance's prototype chain, so Object.getPrototypeOf(blob)
-        // !== Blob.prototype at depth 1 on Chakra. Use isPrototypeOf (chain
-        // walk) until that is fixed.
-        expect(Blob.prototype.isPrototypeOf(blob)).to.equal(true);
+        expect(Object.getPrototypeOf(blob)).to.equal(Blob.prototype);
```

bkaradzic-microsoft:
Rebased onto current `main` (now includes #177 + #179). `fe457e5`.

Both asks from your latest review addressed:

1. **#178 / `isPrototypeOf` workaround in `tests.ts:1292-1297`** — auto-resolved by the rebase. #179 already replaced the workaround block with the depth-1 `Object.getPrototypeOf(blob) === Blob.prototype` check on `main`, and rebasing pulled it in directly. `grep -n 'isPrototypeOf\|#178'` in `tests.ts` returns no matches.

2. **`File.cpp` historical paragraph nit** — applied the suggestion verbatim. The `setPrototypeOf` comment now sticks to *why* (WHATWG subtype + BJS-core `instanceof Blob` branches) and drops the *how-we-got-here* paragraph.

Rebase was clean — 8 commits replayed cleanly on top of `a128e68`, no conflicts. CI on the new head should be picking up shortly; will follow up once it lands.

Polyfills/File/Source/FileReader.cpp
bkaradzic-microsoft: Done in 3de01db — replaced the dangling reference with your suggested ProgressEvent-contract wording.

Polyfills/File/Source/FileReader.cpp
bkaradzic-microsoft: Done in 3de01db — collapsed the 14-line block to a one-line pointer at #173.

Polyfills/File/Source/File.cpp
bkaradzic-microsoft: Done in 3de01db — dropped the `BABYLON_POLYFILL_USE_NAPI_JSI_EVAL` macro and the `__has_include(<napi/env.h>)` guard; `Napi::Eval` is declared on every backend, so it is now an unconditional call.

Polyfills/File/Source/File.cpp
bkaradzic-microsoft: Done — #177 landed, so after rebasing the JS shim collapsed to a direct `Object.setPrototypeOf(File.prototype, Blob.prototype)` in `File::Initialize` (0bff942), and the now-redundant historical comment was trimmed in fe457e5.

Polyfills/File/Source/File.cpp
bkaradzic-microsoft: Done in 3de01db — reordered so the "already provided" check runs first (cheap no-op on hosts with a native File), and the missing-Blob case now throws `Napi::Error` instead of silently returning.

Tests/UnitTests/Scripts/tests.ts
bkaradzic-microsoft: Done — filed #175 (Chakra napi shim swallows throws from class constructors) and re-pointed the disabled test's TODO at it in 3de01db.

Polyfills/File/Source/File.h
bkaradzic-microsoft: This is now even cleaner after the #177 rebase: a single `Object.setPrototypeOf(File.prototype, Blob.prototype)` with no engine-specific shim. `file instanceof Blob` coverage remains in tests.ts.

Polyfills/File/Source/File.cpp
bkaradzic-microsoft: Applied in fe457e5 — dropped the `#177` historical paragraph verbatim per the suggestion.

Polyfills/File/Source/FileReader.cpp
bkaradzic-microsoft: Acknowledged — will revert to the per-lambda `shared_ptr<Napi::ObjectReference>` capture pattern in a follow-up commit. Leaving this thread open until that lands.

Polyfills/File/Source/FileReader.cpp
bkaradzic-microsoft: Acknowledged — the `on*` handlers will move to `InstanceAccessor` with backing `Napi::FunctionReference` slots as part of the FileReader state refactor. Tracking in a follow-up commit; leaving open.

Polyfills/File/Source/FileReader.cpp
bkaradzic-microsoft: Acknowledged — `readyState`/`result`/`error` will become `InstanceAccessor` getters backed by C++ state members in the same FileReader refactor. Leaving open until that commit lands.

Polyfills/File/Source/FileReader.cpp
bkaradzic-microsoft: Acknowledged — will drop the inline `EncodeBase64` and reuse the existing base-n implementation already vendored for Core/AppRuntime/V8Inspector rather than adding a third copy. Follow-up commit; leaving open.

Tests/UnitTests/Scripts/tests.ts
bkaradzic-microsoft: Noted — splitting tests.ts per-polyfill is a reasonable cleanup but orthogonal to this PR; I'd prefer to do it as a standalone follow-up so the diff here stays focused. Leaving this open for your call.

Polyfills/File/Source/FileReader.cpp
bkaradzic-microsoft: Done in eaeaf20 — the on* handlers are now a single `InstanceAccessor` get/set pair on the prototype. The accessor `data` carries the event type, so one pair services all six slots, backed by a `Napi::FunctionReference` map that `Dispatch` consults directly (no more `jsThis.Get("on"+type)`).

Polyfills/File/Source/FileReader.cpp
bkaradzic-microsoft: Done in eaeaf20 — `readyState`/`result`/`error` are now `InstanceAccessor` getters reading `m_readyState`/`m_result`/`m_error`. Every state-machine check reads the C++ members directly, so JS can no longer tamper with them by overwriting the property. The constructor no longer stamps any instance data props.

Polyfills/File/Source/FileReader.cpp
bkaradzic-microsoft: Done in a6155da — promoted header-only `azawadzki/base-n` into JRH as a FetchContent INTERFACE lib (gated on `JSRUNTIMEHOST_POLYFILL_FILE`, pinned to the same `7573e77c` commit BabylonNative uses) and linked it PRIVATE into File. `EncodeBase64` now delegates to `bn::encode_b64` and only appends RFC 4648 padding. Verified output is byte-for-byte identical (`"Hello"` -> `"SGVsbG8="`). BN's direct base-n dep can be dropped as a follow-up.

Tests/UnitTests/Scripts/tests.ts
bkaradzic-microsoft: Filed #180 to split tests.ts per-polyfill as a standalone follow-up, keeping this PR's diff focused on the polyfill.

Polyfills/File/Source/FileReader.cpp
bkaradzic-microsoft: Correction to my earlier note on this thread: nothing to revert — this is already resolved. The shared_ptr trick was replaced by a member `Napi::ObjectReference m_selfRef` anchored across the in-flight read (4480ad8), so the lambdas capture only POD + `this`. Per-lambda `Napi::Persistent` capture isn't viable because `ObjectReference` is move-only and `Napi::Function::New` stores its callable in a CopyConstructible `std::function`-shaped slot; the member ref avoids that while every terminal path (load/error/abort) resets it to break the cycle.

bkaradzic-microsoft:
Addressed the remaining review asks (pushed `eaeaf20`, `a6155da`):

- **readyState / result / error** are now `InstanceAccessor` getters backed by `m_readyState` / `m_result` / `m_error`. State-machine checks read the C++ members directly, so they can't be tampered with from JS, and the constructor no longer stamps instance data props.
- **on\* handlers** collapsed to a single `InstanceAccessor` get/set pair (accessor `data` carries the event type) backed by a `Napi::FunctionReference` map that `Dispatch` consults directly.
- **base64**: removed the bespoke encoder. Promoted header-only `azawadzki/base-n` into JRH as a FetchContent INTERFACE lib (gated on `JSRUNTIMEHOST_POLYFILL_FILE`, pinned to the same `7573e77c` commit BabylonNative uses), linked PRIVATE into File. `EncodeBase64` now delegates to `bn::encode_b64` + RFC 4648 padding; output is byte-for-byte identical. BN's direct base-n dep can be dropped as a follow-up.
- **shared_ptr** thread: already addressed by the member `m_selfRef` (corrected my earlier note).
- **tests.ts split**: filed #180 as a standalone follow-up to keep this diff focused.

All 18 review threads are now resolved.

bkaradzic-microsoft:
CI on `14122f1` is green across every engine. The earlier 7 FileReader async-read failures were a real regression from the `InstanceAccessor` refactor: `readAsText` / `readAsDataURL` produce a **string** `result`, and the previous `Napi::Reference<Napi::Value>` stored it via `napi_create_reference`, which only accepts heap values (object/function/symbol) on real V8/JSC N-API. That threw on those backends (timeout on V8, uncaught C++ exception → SIGABRT on JSC/Linux) while Chakra's shim tolerated it. Fixed by boxing `result`/`error` as properties on a persistent holder object (`Napi::ObjectReference m_state`), which is always a valid reference target; state stays C++-owned and tamper-proof.

Verified green (all FileReader tests pass — 179 passing):
- **V8**: Win32_x64_V8, Android_V8, UWP_x64_V8
- **JSC**: Ubuntu_clang, Ubuntu_gcc, macOS_Xcode164_Sanitizers, iOS, Android_JSC
- **Chakra**: Win32_x86, UWP_x64
- **JSI**: Win32_x64_JSI, UWP_x64_JSI, UWP_arm64_JSI
- **Sanitizers**: Ubuntu ASAN/UBSan + ThreadSanitizer, macOS ThreadSanitizer

The only red job is `Win32_x64_Chakra`, failing solely on the unrelated, pre-existing flaky `WebSocket > should connect correctly with one websocket connection` test (2000 ms timeout against the external `wss://ws.postman-echo.com/raw` echo server; the sibling multi-connection test passes). All FileReader tests pass there too.

Polyfills/File/Source/FileReader.cpp
bghgary: [Responded by Copilot on behalf of @bghgary]

Reopening because the "matches the member-slot pattern used by WebSocket/XHR" framing doesn't match what those polyfills actually do. Neither has an `m_selfRef`-style self-anchor. The convention they share is the opposite shape — externally-anchored lambda-owned state:

```cpp
// Polyfills/WebSocket/Source/WebSocket.h
std::shared_ptr<arcana::cancellation_source> m_cancellationSource{};

// Polyfills/WebSocket/Source/WebSocket.cpp
WebSocket::~WebSocket() {
    m_cancellationSource->cancel();
}

void WebSocket::OpenCallback() {
    m_runtimeScheduler([this, cancellationSource{m_cancellationSource}]() {
        if (cancellationSource->cancelled()) { return; }
        // ... safe to deref `this` here ...
    });
}
```

Each callback lambda captures its OWN copy of the shared_ptr to the cancellation source. The destructor cancels it. Lambdas fired after the wrapper has been collected check the lambda-owned flag first and no-op — they never touch `this->anything` until that check passes. **The wrapper is allowed to die.** XHR follows the same shape (no self-anchor member).

Three things `m_selfRef` does worse than that convention:

1. **Self-cycle.** Wrapper holds a Persistent to its own JS object; only the manual `Reset()` discipline on every terminal path (4 sites today) breaks it. New terminal path → forgot to Reset → wrapper leaks forever.

2. **Token check still races.** Lambdas capture `this` raw. `Abort()` calls `m_selfRef.Reset()` → wrapper becomes GC-eligible → if user has also dropped their JS ref, wrapper can be collected before the queued microtask fires → `if (m_readId != myReadId)` is `this->m_readId` UAF. The self-anchor only protects the no-abort happy path.

3. **The "repo convention" defense is inverted.** The convention IS lambda-owned shared state + destructor-cancels-token, not a self-ref slot. FileReader is the outlier here.

Sketch of the convention-following shape:

```cpp
// Member, no self-ref needed:
std::shared_ptr<std::atomic<uint64_t>> m_readToken{std::make_shared<std::atomic<uint64_t>>(0)};

FileReader::~FileReader() { m_readToken->store(UINT64_MAX); }
void FileReader::Abort(...)  { m_readToken->fetch_add(1); /* ... */ }

void FileReader::StartRead(..., ReadMode mode) {
    const uint64_t myReadId = m_readToken->fetch_add(1) + 1;
    auto token = m_readToken;  // copy for capture

    auto onResolve = Napi::Function::New(env,
        [this, token, myReadId, mode, contentType](const Napi::CallbackInfo& cb) {
            if (token->load() != myReadId) { return; }  // safe even if `this` is dead
            HandleReadResult(myReadId, mode, contentType,
                             cb.This().As<Napi::Object>(), cb[0]);
        });
    // ...
}
```

Drops `m_selfRef`, the 7-line justification, and all four `.Reset()` sites. Wrapper allowed to die; lambdas safely no-op via the lambda-owned token. Matches WebSocket. Doesn't fix the repo-wide engine-teardown story (that's a separate `Finalize` / env-teardown hook) but stops FileReader from introducing a new shape to it.

If you'd rather keep `m_selfRef` despite the convention mismatch, fine — but the "matches existing pattern" justification doesn't hold; this is a new pattern.

bghgary (commented):

Concerns inline.

Polyfills/File/Source/FileReader.cpp
bkaradzic-microsoft: Good call — `m_selfRef` is gone in 61d9924. We agree the self-cycle + manual `Reset()` discipline is the wrong shape, and we moved to externally-anchored, lambda-owned state to match WebSocket.

One spec nuance kept us from the literal token-only sketch: FileAPI requires an **in-flight read to keep the FileReader alive even after script drops its reference** (our "reads a Blob as text via onload" test relies on it). The pure "let the wrapper die + poisoned token" version makes `onload` never fire — the wrapper gets GC'd between `readAsText()` returning and the `.then` microtask, the destructor poisons the token, and all 8 async-read tests time out. WebSocket can let its wrapper die because a dropped in-flight socket closing is acceptable; a dropped in-flight FileReader must still complete.

So instead of a self-ref member, we anchor via a `shared_ptr<Napi::ObjectReference>` captured by **both** promise reactions:

```cpp
auto anchor = std::make_shared<Napi::ObjectReference>(Napi::Persistent(jsThis));
auto onResolve = Napi::Function::New(env,
    [this, anchor, myReadId, mode, contentType](const Napi::CallbackInfo& cb) {
        if (m_readId != myReadId) return;   // abort/restart guard
        HandleReadResult(myReadId, mode, contentType, anchor->Value(),
                         cb.Length() ? cb[0] : cb.Env().Null());
    });
// onReject captures the same `anchor`
```

This addresses all three points:
1. **No self-cycle / no manual Reset.** The anchor is owned by the lambdas (external to the wrapper), held by the pending promise. When the promise settles and the engine releases both reactions, the last `shared_ptr` copy drops and the anchor releases automatically — all four terminal-path `Reset()` calls are deleted.
2. **No dead-`this` race.** The anchor keeps the wrapper alive until a reaction actually runs, so `m_readId` and `anchor->Value()` are always valid — including on the abort-then-drop path the old self-anchor left racy.
3. **Convention-following shape.** Externally-anchored lambda-owned state, the same shape as WebSocket. `m_readId` is retained only as the abort/restart guard (an aborted read's promise can still settle after a *subsequent* read started, which the `readyState == LOADING` check alone wouldn't distinguish).

Win32-x64 Chakra Debug: all 180 JS tests pass (the 8 timeouts from the token-only experiment are green again) + 4 native gtests.

bkaradzic-microsoft:
@bghgary thanks for the careful review — all of your inline concerns are addressed in 61d9924:

- **`m_selfRef` self-anchor is gone.** Replaced with an externally-anchored `shared_ptr<Napi::ObjectReference>` captured by both promise reactions, matching the WebSocket lambda-owned-state convention. No self-cycle, no manual `Reset()` on terminal paths (all four deleted), and no dead-`this` race — the wrapper is guaranteed alive whenever a reaction runs. `m_readId` is kept solely as the abort/restart guard. Details + the one spec nuance (an in-flight read must keep the reader alive, so the wrapper can't simply be allowed to die) are in the resolved inline thread.

I also refreshed the PR description to match the current implementation (it had drifted): C++-backed readonly accessors for `readyState`/`result`/`error` and the `on*` slots, boxed `result`/`error` for primitive string results on real N-API, `base-n` for `readAsDataURL` (no more inlined encoder), and updated test counts (29 new: 15 File + 14 FileReader; 180/180 JS tests).

Verified Win32-x64 Chakra Debug green (180 JS + native gtests), and full CI is green across all engines/platforms. PTAL.

bghgary (approved):

LGTM. All concerns addressed; the FileReader in-flight lifecycle/UAF fix is verified in `61d9924` — `shared_ptr<ObjectReference>` anchor captured by both promise reactions, no self-ref member, no manual `Reset()`.

Issue #173, opened 2026-06-02
Web polyfill constants: add missing InstanceValue exposure on WebSocket and XMLHttpRequest


bghgary:
Per WHATWG IDL, class `const` members are exposed on both the constructor (static) and instances (via the prototype). JRH polyfills currently do this inconsistently:

- `Polyfills/WebSocket/Source/WebSocket.cpp` (L16–19) — `CONNECTING` / `OPEN` / `CLOSING` / `CLOSED` are `StaticValue` only. `new WebSocket(...).OPEN` returns `undefined`.
- `Polyfills/XMLHttpRequest/Source/XMLHttpRequest.cpp` (L72–76) — `UNSENT` / `OPENED` / `HEADERS_RECEIVED` / `LOADING` / `DONE` are `StaticValue` only. Same gap.
- `Polyfills/File/Source/FileReader.cpp` (new in #169) — exposes both, conformant.

**Fix:** add an `InstanceValue` entry alongside each `StaticValue` for both polyfills. Going forward, web polyfill classes with IDL `const` members should define one of each.

**Follow-up:** investigate whether a helper function/macro can emit both descriptors from a single declaration so authors don't have to remember the dual exposure (and constants can't silently drift between static and instance values).

**References:**
- https://websockets.spec.whatwg.org/#interface-definition
- https://xhr.spec.whatwg.org/#interface-xmlhttprequest
- https://w3c.github.io/FileAPI/#APIASynch

PR #174, opened 2026-06-02
JSC napi: stop polluting global Object.prototype from napi_define_class

### LINKED ISSUE #172: JSC napi shim: writing to constructor.prototype pollutes global Object.prototype
[Filed by Copilot on behalf of @bghgary]

On the JSC backend, constructors returned by `napi_define_class` / `Napi::Function::New(...)` have their user-visible `.prototype` slot pointing at the global `Object.prototype` instead of a fresh per-class object. Writing to it from C++ via `func.Get("prototype").Set(...)` therefore mutates `Object.prototype` globally, breaking `for..in` and prototype-chain lookups across the runtime.

V8 and ChakraCore return a fresh per-class object, so the bug is JSC-specific.

**Repro**

```cpp
Napi::Function ctor = DefineClass(env, "Foo", {});
ctor.Get("prototype").As<Napi::Object>().Set("X", Napi::Number::New(env, 0));
```

In JS: `'X' in {}` → `true` (should be `false`), and `for (const k in {}) ...` enumerates `X`.

**Root cause**

JSC's `JSObjectMakeConstructor` defaults the new function's `.prototype` to `Object.prototype` when none is supplied. The napi-jsc shim should allocate a fresh object per class.

**Workaround**

Use `InstanceValue` / `InstanceMethod` / `InstanceAccessor` inside `DefineClass`. That path goes through napi's internal prototype slot, distinct from the user-visible `.prototype` on JSC, and works correctly on every backend.

Discovered while reviewing #169.

bkaradzic-microsoft:
Closes #172.

`JSObjectMakeConstructor(ctx, nullptr, ...)` returns a function whose `prototype` property points at the global `Object.prototype`. Any napi caller that writes to `constructor.prototype` (either internally via `napi_define_class` for instance members, or in user code via `ctor.Get("prototype").Set(...)` ) was mutating `Object.prototype` itself, corrupting `'X' in {}` / `for..in` across the runtime.

### Previous workaround

The earlier `// BEGIN TODO ... END TODO` block in `ConstructorInfo::Create` set the constructor's `[[Prototype]]` internal slot to a fresh object and re-routed `napi_get_prototype` + `CallAsConstructor` through it. That kept `napi_define_class`'s own instance-property writes out of `Object.prototype`, but left the user-visible `.prototype` property still pointing at `Object.prototype`. User code (#169) hit the pollution head-on.

It also conflated `Object.getPrototypeOf(ctor)` (per spec: `Function.prototype` for a normal constructor) with `ctor.prototype` (the per-class instance prototype). Two slots, two semantics — the previous code wrote both to the same object.

### Fix

- `ConstructorInfo::Create` installs a fresh per-class prototype object as the constructor's **`.prototype` property** (`DontEnum|DontDelete`) with a non-enumerable `.constructor` back-reference, matching ECMAScript semantics for normal constructors. The constructor's `[[Prototype]]` slot is left at JSC's default (`Function.prototype`).
- `CallAsConstructor` sets the new instance's `[[Prototype]]` from the constructor's `.prototype` **property**, not from its `[[Prototype]]` slot.
- `napi_define_class` looks up the `.prototype` property directly instead of going through `napi_get_prototype` (which per N-API spec equals `Object.getPrototypeOf`, i.e. the `[[Prototype]]` internal slot = `Function.prototype` for a constructor — the old call was a semantic mismatch that only happened to work because of the kludge).
- `napi_get_prototype` itself is unchanged and remains spec-compliant.

### Regression test

Added two tests in the existing `Blob` describe block (covers every engine that ships Blob, which is all of them):

- `Blob.prototype` is a fresh per-class object whose `[[Prototype]]` is `Object.prototype` (not `Blob.prototype === Object.prototype` as before).
- None of Blob's instance members (`size`, `type`, `arrayBuffer`, `text`, `bytes`) leak onto the global `Object.prototype` — i.e. `'size' in {} === false`.

These tests would have caught the pollution before #169 surfaced it via the File polyfill.

### Followup

Once this lands, #169's `JS_PROTOTYPE_CHAIN_SHIM` collapses to a single direct `Object.setPrototypeOf(File.prototype, Blob.prototype)` in C++; the `Napi::Eval` / `env.RunScript` dispatch goes away; and the 14-line `FileReader` dual-`StaticValue`/`InstanceValue` comment shrinks to one line (per #173's recommended WHATWG pattern).

### DIFF
--- a/Core/Node-API/Source/js_native_api_javascriptcore.cc
+++ b/Core/Node-API/Source/js_native_api_javascriptcore.cc
@@ -323,17 +323,24 @@ namespace {
                               void* data,
                               napi_value* result) {
       JSObjectRef constructor{JSObjectMakeConstructor(env->context, nullptr, CallAsConstructor)};
-      // BEGIN TODO: This extra prototype should no longer be needed, but for some reason removing it leads to errors
-      //             when setting properties on some prototypes. This should be investigated and removed.
+
+      // JSObjectMakeConstructor(ctx, nullptr, ...) leaves the resulting function's
+      // `prototype` property pointing at the global `Object.prototype`. Writing to
+      // it (e.g. via napi_define_properties for instance members, or user code via
+      // ctor.Get("prototype").Set(...)) would pollute the global Object.prototype
+      // and corrupt every object in the runtime. Install a fresh per-class
+      // prototype object whose `[[Prototype]]` defaults to Object.prototype and
+      // whose `constructor` back-reference points at the new constructor.
       JSObjectRef prototype{JSObjectMake(env->context, nullptr, nullptr)};
-      JSObjectSetPrototype(env->context, prototype, JSObjectGetPrototype(env->context, constructor));
-      JSObjectSetPrototype(env->context, constructor, prototype);
 
       JSValueRef exception{};
       JSObjectSetProperty(env->context, prototype, JSString("constructor"), constructor,
-        kJSPropertyAttributeReadOnly | kJSPropertyAttributeDontDelete, &exception);
+        kJSPropertyAttributeDontEnum, &exception);
       CHECK_JSC(env, exception);
-      // END TODO
+
+      JSObjectSetProperty(env->context, constructor, JSString("prototype"), prototype,
+        kJSPropertyAttributeDontEnum | kJSPropertyAttributeDontDelete, &exception);
+      CHECK_JSC(env, exception);
 
       ConstructorInfo* info{new ConstructorInfo(env, utf8name, length, cb, data)};
       if (info == nullptr) {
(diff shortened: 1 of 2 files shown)


bkaradzic-microsoft:
Closing this without merging.

I confirmed by reading the WebKit source ([`JSObjectRef.cpp:128`](https://github.com/WebKit/WebKit/blob/main/Source/JavaScriptCore/API/JSObjectRef.cpp#L128)) that `JSObjectMakeConstructor` sets the `.prototype` property with `PropertyAttribute::DontEnum | PropertyAttribute::DontDelete | PropertyAttribute::ReadOnly`. The **`ReadOnly`** bit makes the slot non-writable, so `JSObjectSetProperty` to replace its value calls `put()` and throws (`[Uncaught Error] defineProperty@[native code]` — exactly what CI surfaced on every JSC job).

The naive fix would be `JSObjectMakeFunctionWithCallback` instead, but [`JSCallbackFunction.cpp:46`](https://github.com/WebKit/WebKit/blob/main/Source/JavaScriptCore/API/JSCallbackFunction.cpp#L46) passes `nullptr` for the construct callback, so the result is not constructable.

The actually-correct fix is to switch the entire `ConstructorInfo::Create` path from `JSObjectMakeConstructor` to `JSObjectMake` with a `JSClassDefinition` that provides both `callAsConstructor` and `callAsFunction` — that produces a callable+constructable `JSCallbackObject` whose `.prototype` we own from the start. That refactor also has to revisit:

- The `sentinel` pattern in `NativeInfo::Link` / `Query` (currently constructor and sentinel are separate JSObjects; with the new path, the constructor IS the JSCallbackObject and holds the `ConstructorInfo*` as private data directly).
- `napi_new_instance` and `napi_instanceof` semantics on the new object type.
- All other callers that assume constructors are `JSCallbackConstructor` instances.

That's bigger than what makes sense to bundle with #169, so I'm leaving #172 open and #169 will keep its JS-shim workaround for the `.prototype === Object.prototype` quirk. The JS shim is ugly but well-isolated to the File polyfill and survives the engine-specific behavior cleanly. Happy to take the bigger refactor in a separate PR.

PR #176, opened 2026-06-03
Hermes

CedricGuillemet:


patches/hermes-shutdown-pre-mark-deletion-pending.patch
ryantrem, line 1: Is this a bug in the Hermes NAPI impl?

CedricGuillemet: Discussed offline.

Core/Node-API/Source/env_hermes.cc
```diff
@@ -0,0 +1,223 @@
...
+    Napi::Env Attach()
+    {
+        // Default Hermes config is fine for embedding: MicrotaskQueue is on,
+        // ES6 Proxy + generators are on, Intl is on, EnableEval is on.
+        // We bump the max GC heap to something reasonable for running our
+        // Mocha test suite (the unit-test default of 512 KiB is too small).
+        auto config = hermes::vm::RuntimeConfig::Builder()
+                          .withGCConfig(hermes::vm::GCConfig::Builder()
+                                            .withInitHeapSize(1u << 20)        //   1 MiB
+                                            .withMaxHeapSize(512u << 20)       // 512 MiB
```
ryantrem: Why 512 max? This seems likely too limited.

Core/Node-API/Source/env_hermes.cc
```diff
@@ -0,0 +1,223 @@
...
+            {
+                return;
+            }
+            state = std::move(it->second);
+            StateMap().erase(it);
+        }
+
+        // Dropping the last shared_ptr to the runtime tears down the env via
+        // Hermes's post-shutdown deleter.
+        state.runtime.reset();
```
ryantrem: Isn't this going to happen automatically when the `state` variable goes out of scope at the end of this function anyway?

Core/Node-API/Source/env_hermes.cc
```diff
@@ -0,0 +1,223 @@
...
+        {
+            return;
+        }
+        // We intentionally ignore the ExecutionStatus return: any unhandled
+        // exception raised by a microtask is surfaced to JS via the standard
+        // unhandled-rejection mechanism (Hermes prints it via HermesInternal),
+        // and we don't have a meaningful way to bubble it up here.  Real
+        // exceptions thrown FROM user callbacks are already handled in
+        // AppRuntime::Dispatch's try/catch.
+        (void)runtime->drainJobs();
```
ryantrem, line 166: Does this result in behavior similar to other js engines in BN? Does this not need to be reported through the unhandled error callback or anything?

Tests/UnitTests/Android/app/build.gradle
```diff
@@ -7,6 +7,15 @@ if (project.hasProperty("jsEngine")) {
     jsEngine = project.property("jsEngine")
 }
 
+def cmakeArguments = [
+    "-DANDROID_STL=c++_shared",
+    "-DNAPI_JAVASCRIPT_ENGINE=${jsEngine}",
+    "-DJSRUNTIMEHOST_CORE_APPRUNTIME_V8_INSPECTOR=ON"
+]
+if (project.hasProperty("importHostCompilers")) {
```
ryantrem, line 21: Maybe add a comment on this.

ryantrem (approved):

Core/Node-API/Source/env_hermes.cc
CedricGuillemet: indeed. removed.

Core/Node-API/Source/env_hermes.cc
CedricGuillemet: It's not committed memory so we can up that value. I guess we can keep 512Mb (should be enough for a mobile) and enable users to change the value with a CMAKE variable ?

Core/Node-API/Source/env_hermes.cc
CedricGuillemet: "V8/JSC/Chakra auto-drain at scope/script boundaries and rely on the engine's own unhandled-rejection reporting; Hermes' drainJobs() does the same internally (HermesInternal surfaces unhandled rejections)."

Core/Node-API/Source/env_hermes.cc
ryantrem: I would have thought for modern devices we'd want like 2gb or something.

CedricGuillemet: sounds good. it's only ram for hermes. textures, vertex buffers, anything vram or native side allocations are not part of these 2Gb.

Core/Node-API/Source/env_hermes.cc
ryantrem: Yea I saw the comment, I just wasn't clear how "JS via the standard unhandled-rejection mechanism (Hermes prints it via HermesInternal)" relates to the unhandled error callback in AppRuntime.

PR #181, opened 2026-06-04
Fix integer overflow in napi_create_dataview bounds check

bkaradzic-microsoft:
## Summary
`napi_create_dataview` (Chakra backend) validated the requested view with an unchecked `byte_length + byte_offset > bufferLength` comparison. Both operands are caller-supplied `size_t` values, so on 64-bit builds their sum can **overflow and wrap** past the limit. When that happens:
- the values are truncated to 32-bit for `JsCreateDataView` (yielding a small, valid view), but
- the **original 64-bit** `byte_offset`/`byte_length` are stored in `DataViewInfo` and later returned by `napi_get_dataview_info` alongside the small backing buffer

…handing a calling native addon an out-of-bounds read/write primitive (CWE-190 → CWE-125/787, also CWE-681).

## Fix
Validate `byte_offset` and `byte_length` against the buffer size **individually, without adding them**:
`cpp
if (byte_offset > bufferLength || byte_length > bufferLength - byte_offset) { /* range error */ }
`
After this check both values are `<= bufferLength` (a 32-bit quantity), so the later 32-bit truncation and the values stored in `DataViewInfo` are guaranteed in range.

- **JavaScriptCore** backend delegates to the JS `DataView` constructor → unaffected.
- **V8** backend carries the same upstream pattern but is vendored verbatim from Node → left untouched.

## Test
This path is **not reachable from JS** (`new DataView` uses the engine directly, not this N-API), so coverage is a native gtest. It crafts an offset/length whose low 32 bits are valid for a 16-byte buffer but whose 64-bit sum wraps, and asserts the view is rejected (and never reports the raw 64-bit extents). Guarded to the Chakra engine and 64-bit builds (`JSRUNTIMEHOST_NAPI_ENGINE_CHAKRA` + `_WIN64`).

Verified the test **fails on the pre-fix code** and **passes after the fix**; full suite green (5 native tests + 151 JS tests) on Win32-x64 Chakra Debug.

### DIFF
--- a/Core/Node-API/Source/js_native_api_v8.cc
+++ b/Core/Node-API/Source/js_native_api_v8.cc
@@ -3185,6 +3185,7 @@ napi_status NAPI_CDECL napi_create_dataview(napi_env env,
   RETURN_STATUS_IF_FALSE(env, value->IsArrayBuffer(), napi_invalid_arg);
 
   v8::Local<v8::ArrayBuffer> buffer = value.As<v8::ArrayBuffer>();
+  // [BABYLON-NATIVE-ADDITION]: overflow-safe bounds check; not in Node.js upstream
   // byte_offset and byte_length are caller-supplied size_t values; computing
   // byte_length + byte_offset can overflow and wrap past the buffer size, which
   // would slip an out-of-range request through and trip a fatal CHECK inside
--- a/Core/Node-API/Source/js_native_api_v8.cc
+++ b/Core/Node-API/Source/js_native_api_v8.cc
@@ -3185,7 +3185,12 @@ napi_status NAPI_CDECL napi_create_dataview(napi_env env,
   RETURN_STATUS_IF_FALSE(env, value->IsArrayBuffer(), napi_invalid_arg);
 
   v8::Local<v8::ArrayBuffer> buffer = value.As<v8::ArrayBuffer>();
-  if (byte_length + byte_offset > buffer->ByteLength()) {
+  // byte_offset and byte_length are caller-supplied size_t values; computing
+  // byte_length + byte_offset can overflow and wrap past the buffer size, which
+  // would slip an out-of-range request through and trip a fatal CHECK inside
+  // v8::DataView::New. Validate each against the buffer size without adding them.
+  if (byte_offset > buffer->ByteLength() ||
+      byte_length > buffer->ByteLength() - byte_offset) {
     napi_throw_range_error(env,
                            "ERR_NAPI_INVALID_DATAVIEW_ARGS",
                            "byte_offset + byte_length should be less than or "
--- a/Tests/UnitTests/CMakeLists.txt
+++ b/Tests/UnitTests/CMakeLists.txt
@@ -46,6 +46,12 @@ endif()
 add_executable(UnitTests ${SOURCES} ${SCRIPTS} ${TYPE_SCRIPTS} ${ASSETS})
 target_compile_definitions(UnitTests PRIVATE JSRUNTIMEHOST_PLATFORM="${JSRUNTIMEHOST_PLATFORM}")
 
+# The V8JSI Node-API shim does not implement napi_create_dataview, so the
+# CreateDataViewRejectsOverflowingRange test is compiled out on that backend.
+if(NAPI_JAVASCRIPT_ENGINE STREQUAL "JSI")
+    target_compile_definitions(UnitTests PRIVATE JSRUNTIMEHOST_NAPI_ENGINE_JSI)
+endif()
+
 target_link_libraries(UnitTests
     PRIVATE AppRuntime
     PRIVATE Console
--- a/Tests/UnitTests/Shared/Shared.cpp
+++ b/Tests/UnitTests/Shared/Shared.cpp
@@ -16,6 +16,7 @@
 #include <arcana/threading/blocking_concurrent_queue.h>
 #include <atomic>
 #include <chrono>
+#include <cstdint>
 #include <future>
 #include <iostream>
 #include <thread>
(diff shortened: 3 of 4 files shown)


Tests/UnitTests/Shared/Shared.cpp
```diff
@@ -276,6 +277,76 @@ TEST(AppRuntime, DestroyDoesNotDeadlock)
     testThread.join();
 }
 
+#ifdef JSRUNTIMEHOST_NAPI_ENGINE_CHAKRA
```
bghgary: Drop this guard. The test's assertions are backend-agnostic (overflow rejected, or reported extents in range), so it should run on V8/JSC too and verify they reject the overflow instead of assuming it. The guard also only exists by adding the `JSRUNTIMEHOST_NAPI_ENGINE_CHAKRA` compile-def in CMakeLists, which is otherwise unused — so this forces a CMake change for no real reason.

Tests/UnitTests/CMakeLists.txt
```diff
@@ -46,6 +46,10 @@ endif()
 add_executable(UnitTests ${SOURCES} ${SCRIPTS} ${TYPE_SCRIPTS} ${ASSETS})
 target_compile_definitions(UnitTests PRIVATE JSRUNTIMEHOST_PLATFORM="${JSRUNTIMEHOST_PLATFORM}")
 
+if(NAPI_JAVASCRIPT_ENGINE STREQUAL "Chakra")
+    target_compile_definitions(UnitTests PRIVATE JSRUNTIMEHOST_NAPI_ENGINE_CHAKRA)
```
bghgary: This compile-def exists only to drive the Chakra `#ifdef` around the new test. With that guard removed, nothing else uses it — drop this block too.

Tests/UnitTests/Shared/Shared.cpp
```diff
@@ -276,6 +277,76 @@ TEST(AppRuntime, DestroyDoesNotDeadlock)
...
+#ifdef _WIN64
```
bghgary: `_WIN64` is MSVC/Windows-only, not a portable 64-bit check. The precondition is a 64-bit `size_t`, and `<cstdint>` is already included:

```suggestion
#if SIZE_MAX > 0xFFFFFFFFu
```

Leave this as the sole guard — it also works on the non-Windows backends the test should now run on.

bghgary (changes requested):

Fix LGTM; comments inline.

Tests/UnitTests/Shared/Shared.cpp
bkaradzic-microsoft: Done — removed the `JSRUNTIMEHOST_NAPI_ENGINE_CHAKRA` guard. The test is backend-agnostic (it asserts the overflow is either rejected or reported in range), so it now runs on V8/JSC too and verifies they reject it. Confirmed it still passes on Chakra locally.

Tests/UnitTests/CMakeLists.txt
bkaradzic-microsoft: Done — dropped this block; with the `#ifdef` removed, nothing else used `JSRUNTIMEHOST_NAPI_ENGINE_CHAKRA`.

Tests/UnitTests/Shared/Shared.cpp
bkaradzic-microsoft: Done — replaced `_WIN64` with `#if SIZE_MAX > 0xFFFFFFFFu` as the sole guard (`<cstdint>` is already included), so it's a portable 64-bit `size_t` check that also covers the non-Windows backends.

bkaradzic-microsoft:
Removing the Chakra-only guard (per review) surfaced two pre-existing issues in the other backends, now fixed:

- **V8 (Win32_x64_V8, Android_V8) crashed** rather than rejecting the overflow. `napi_create_dataview` in `js_native_api_v8.cc` carried the same unchecked `byte_length + byte_offset > buffer->ByteLength()` comparison; the wrapped sum slipped past it and `v8::DataView::New` hit a fatal `CHECK` (`Check failed: byte_length <= buffer->GetByteLength()`), aborting the process. Applied the same overflow-safe check, so V8 now rejects the overflow — exactly what the test asserts. Verified locally (V8 suite green).

- **JSI (Win32_x64_JSI) failed to compile** (`C3861: 'napi_create_dataview': identifier not found`). The V8JSI Node-API shim (`Core/Node-API-JSI`) doesn't implement `napi_create_dataview` / `napi_get_dataview_info` at all — its `DataView::New` literally `throw std::runtime_error{"TODO"}`. Since the API is absent there, the test is compiled out on JSI via a `JSRUNTIMEHOST_NAPI_ENGINE_JSI` define. This is a genuine "API not implemented" gate (different from the earlier unnecessary Chakra-only define), so the test still runs on Chakra, V8, and JavaScriptCore on 64-bit.

Verified locally on all three: V8 passes, Chakra passes, JSI compiles cleanly (test gated out).

Core/Node-API/Source/js_native_api_v8.cc
```diff
@@ -3185,7 +3185,12 @@ napi_status NAPI_CDECL napi_create_dataview(napi_env env,
   RETURN_STATUS_IF_FALSE(env, value->IsArrayBuffer(), napi_invalid_arg);
 
   v8::Local<v8::ArrayBuffer> buffer = value.As<v8::ArrayBuffer>();
-  if (byte_length + byte_offset > buffer->ByteLength()) {
+  // byte_offset and byte_length are caller-supplied size_t values; computing
```
bghgary, line 3189: Per our convention, changes to the vendored Node.js V8 N-API need the `// [BABYLON-NATIVE-ADDITION]` marker (e.g. lines 361, 2553, 3372) so the divergence stays visible when syncing from upstream.

```suggestion
  // [BABYLON-NATIVE-ADDITION]: overflow-safe bounds check; not in Node.js upstream
  // byte_offset and byte_length are caller-supplied size_t values; computing
```

bghgary (changes requested):

Fix LGTM; one comment inline.

bghgary (approved):

LGTM

PR #188, opened 2026-06-06
Add native fetch() polyfill

### LINKED ISSUE #98: Implement fetch API
When we do this, we can update Babylon.js to use the fetch API instead of XMLHttpRequest. Perhaps even deprecate XMLHttpRequest in this repo.

Note that fetch polyfill with XMLHttpRequest does work.

bkaradzic-microsoft:
Implements the WHATWG etch() API as a **native** polyfill under `Polyfills/Fetch/`, mirroring the `XMLHttpRequest` polyfill layout. Like `XMLHttpRequest`, it is built on the same platform-specific `UrlLib` transports, so libcurl/WinHTTP/etc. behavior is identical between the two.

Closes #98.

### API
`fetch(input, init)` returns a `Promise` that resolves to a `Response`-like object exposing:
- `ok`, `status`, `statusText`, `url`, `redirected`, `type`, `bodyUsed`
- `headers` with `get(name)` / `has(name)` / `forEach(cb)` (case-insensitive)
- `text()`, `arrayBuffer()`, `json()`, `blob()` (each returns a `Promise`)
- `clone()`

### Behavior / deliberate limitations
- Per the fetch spec, the promise **rejects only on transport-level failures**; a completed request with a non-2xx status (e.g. 404) still resolves with `ok === false`.
- The body is fully buffered before the promise resolves, so the body accessors may be called more than once (`bodyUsed` stays `false`) — a lenient deviation from the spec's single-use semantics.
- `GET`/`POST` only and string request bodies only, matching the underlying `UrlLib` transport (same constraints as `XMLHttpRequest`).
- `blob()` requires the `Blob` polyfill to be initialized.

### Wiring
- New `JSRUNTIMEHOST_POLYFILL_FETCH` option (default `ON`). `UrlLib` is now fetched when **either** `XMLHttpRequest` or `Fetch` is enabled.
- Initialized in the test harness and covered by a new `describe("fetch")` block (12 cases: status/ok, 404-resolves, statusText, text/arrayBuffer/json/blob, case-insensitive headers, clone, init method, no-arg rejection). A small `Assets/sample.json` backs the deterministic `json()` test.

### Verification
Built the `Fetch` target under `warnings_as_errors` and ran the full UnitTests suite on Win32/Chakra: **192 passing**, including all 12 new fetch cases.

### Context
This is step 2 of the fetch native-polyfill pivot discussed in BabylonNative#1707 (reviewers asked fetch to follow the same native pattern as File/FileReader (#169) and AbortController). Once this lands, BabylonNative#1707 becomes a small bump that initializes `Polyfills::Fetch`.


bkaradzic-microsoft:
Addressed both review comments in ebe6a80:
- **blob() on JSC/JSI**: switched the Blob-availability check from IsFunction() to IsUndefined()/IsNull(), matching the existing File polyfill workaround. Some JavaScriptCore/JSI builds report constructor functions as typeof 'object', which was causing the \lob()\ test to fail across all JSC/JSI CI jobs.
- **<cctype>**: added the explicit include for \std::tolower\/\std::toupper\.

This should turn the failing JSC/JSI/Android/Ubuntu/macOS/iOS builds green (they were all failing solely on the blob() test).

bkaradzic-microsoft:
CI is now fully green (23/23 ✅). Pushed four follow-up fixes after the initial submission, all surfaced by platform-specific CI:

1. **\40f5ff7\** — Fixed a stack-use-after-return crash (\std::system_error: Invalid argument\ on clang/libc++, JSC, JSI, Android). arcana's \	ask::then()\ captures the scheduler **by reference** and runs it on the UrlLib worker thread after \etch()\ has returned, so the stack-local \JsRuntimeScheduler\ dangled. Now heap-allocated and co-owned by the continuation (mirrors how XMLHttpRequest keeps its scheduler alive as a member). Caught by ASAN.
2. **\be6a80\** — \lob()\ now detects the Blob polyfill via \IsUndefined()/IsNull()\ instead of \IsFunction()\ (JSC/JSI report constructors as typeof 'object'), plus added \<cctype>\. *(addresses both Copilot review comments)*
3. **\44bc77a\** — Made the \lob()\ \parts\/\options\ Napi objects non-const (Node-API-JSI declares \Set()\ non-const → C2663).
4. **\1118799\** — Linked \Fetch\ into the Android \UnitTestsJNI\ target, which has its own polyfill link list.

Ready for re-review.

Polyfills/Fetch/Source/Fetch.cpp
```diff
@@ -0,0 +1,396 @@
...
+#include <UrlLib/UrlLib.h>
+
+#include <algorithm>
+#include <cstring>
+#include <memory>
+#include <optional>
+#include <string>
+#include <utility>
+#include <vector>
+
```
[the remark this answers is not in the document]
bkaradzic-microsoft: Good catch — added #include <cctype> in ebe6a80.

Polyfills/Fetch/Source/Fetch.cpp
```diff
@@ -0,0 +1,396 @@
...
+            response.Set("blob", Napi::Function::New(env, [data](const Napi::CallbackInfo& info) -> Napi::Value {
+                Napi::Env env = info.Env();
+                const auto deferred = Napi::Promise::Deferred::New(env);
+
+                const auto blobConstructor = env.Global().Get("Blob");
+                if (!blobConstructor.IsFunction())
+                {
+                    deferred.Reject(Napi::Error::New(env, "fetch: Blob is not available in this environment").Value());
+                    return deferred.Promise();
+                }
```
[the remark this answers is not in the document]
bkaradzic-microsoft: Fixed in ebe6a80 — lob() now checks IsUndefined()/IsNull() instead of IsFunction(), matching the File polyfill. This was the actual cause of the failing JSC/JSI CI jobs; all 23 checks are green now.

Polyfills/Fetch/Source/Fetch.cpp
```diff
@@ -0,0 +1,404 @@
...
+                            data->url = std::string{request->ResponseUrl()};
+                            for (const auto& header : request->GetAllResponseHeaders())
+                            {
+                                data->headers.emplace_back(header.first, header.second);
+                            }
+                            const auto responseBuffer = request->ResponseBuffer();
+                            data->body = std::make_shared<std::vector<std::byte>>(responseBuffer.begin(), responseBuffer.end());
+
+                            deferred.Resolve(BuildResponse(env, data));
+                        });
```
bghgary: A throw here (e.g. from `BuildResponse`) is captured by arcana into the discarded result task and dropped, so the promise never settles and `await fetch(...)` hangs.

```suggestion
                    request->SendAsync()
                        .then(*scheduler, arcana::cancellation::none(),
                            [deferred, request, env](const arcana::expected<void, std::exception_ptr>& result) {
                                const int status = static_cast<int>(request->StatusCode());

                                // Per the WHATWG fetch spec, only transport-level failures reject. A completed
                                // request with a non-2xx status (e.g. 404) still resolves with response.ok === false.
                                // A status of 0 indicates the transport never produced a response (network error).
                                if (result.has_error() || status == 0)
                                {
                                    throw std::runtime_error{"fetch: network request failed"};
                                }

                                auto data = std::make_shared<ResponseData>();
                                data->statusCode = status;
                                data->url = std::string{request->ResponseUrl()};
                                for (const auto& header : request->GetAllResponseHeaders())
                                {
                                    data->headers.emplace_back(header.first, header.second);
                                }
                                const auto responseBuffer = request->ResponseBuffer();
                                data->body = std::make_shared<std::vector<std::byte>>(responseBuffer.begin(), responseBuffer.end());

                                deferred.Resolve(BuildResponse(env, data));
                            })
                        .then(*scheduler, arcana::cancellation::none(),
                            [deferred, env, scheduler](const arcana::expected<void, std::exception_ptr>& result) {
                                if (result.has_error())
                                {
                                    deferred.Reject(Napi::Error::New(env, result.error()).Value());
                                }
                            });
```

Polyfills/Fetch/Source/Fetch.cpp
```diff
@@ -0,0 +1,404 @@
...
+                }
+                catch (const Napi::Error& error)
+                {
+                    deferred.Reject(error.Value());
+                }
+                catch (const std::exception& error)
+                {
+                    deferred.Reject(Napi::Error::New(env, std::string{"fetch: "} + error.what()).Value());
+                }
```
bghgary: Optional: a single `catch (...)` via the `exception_ptr` overload also removes the `Napi::Error`-vs-`std::exception` prefix asymmetry.

```suggestion
                catch (...)
                {
                    deferred.Reject(Napi::Error::New(env, std::current_exception()).Value());
                }
```

Polyfills/Fetch/Source/Fetch.cpp
```diff
@@ -0,0 +1,404 @@
...
+            }();
+
+            if (upper == "GET")
+            {
+                return UrlLib::UrlMethod::Get;
+            }
+            if (upper == "POST")
+            {
+                return UrlLib::UrlMethod::Post;
+            }
```
bghgary: Drop the inline `toupper` and compare with `EqualsIgnoreCase`, which also removes the second case-fold:

```suggestion
            if (EqualsIgnoreCase(method, "GET"))
            {
                return UrlLib::UrlMethod::Get;
            }
            if (EqualsIgnoreCase(method, "POST"))
            {
                return UrlLib::UrlMethod::Post;
            }
```

Duplicates XHR's `MethodType::StringToEnum` (which is case-*sensitive*, so non-conforming). Since UrlLib owns `UrlMethod`, this mapping ideally belongs there — follow-up, not this PR.

Polyfills/Fetch/Source/Fetch.cpp
```diff
@@ -0,0 +1,404 @@
...
+                case 408: return "Request Timeout";
+                case 409: return "Conflict";
+                case 429: return "Too Many Requests";
+                case 500: return "Internal Server Error";
+                case 502: return "Bad Gateway";
+                case 503: return "Service Unavailable";
+                case 504: return "Gateway Timeout";
+                default: return "";
+            }
+        }
```
bghgary, line 82: `statusText` is a workaround for UrlLib exposing only the numeric code (returns `""` for any unlisted code). Belongs in UrlLib — expose the real reason phrase, falling back to a code→text table for HTTP/2+ where the wire carries none. Would also give XHR, currently missing `statusText`, one for free. Follow-up, not this PR.

Polyfills/Fetch/Source/Fetch.cpp
```diff
@@ -0,0 +1,404 @@
...
+{
+    namespace
+    {
+        // Buffered response payload shared between the Response object and any clones it produces.
+        struct ResponseData
+        {
+            int statusCode{};
+            std::string url;
+            std::vector<std::pair<std::string, std::string>> headers;
+            std::shared_ptr<std::vector<std::byte>> body;
```
bghgary: `body` doesn't need to be a `shared_ptr`. `ResponseData` is only ever held as a `shared_ptr<ResponseData>`, and every accessor plus `clone()` captures that same pointer — so `body` is already shared and kept alive through the enclosing struct. The inner `shared_ptr<vector<byte>>` adds an allocation and an indirection without changing sharing, and misleads the reader into thinking clones get independent bodies (they don't — `clone()` reuses the same `ResponseData`). Make it a plain `std::vector<std::byte> body;` and change the `data->body->` use sites to `data->body.`.

Polyfills/Fetch/Source/Fetch.cpp
```diff
@@ -0,0 +1,404 @@
...
+        };
+
+        std::string ToLower(std::string value)
+        {
+            std::transform(value.begin(), value.end(), value.begin(), [](unsigned char c) { return static_cast<char>(std::tolower(c)); });
+            return value;
+        }
```
bghgary: With non-allocating compares, `ToLower` isn't needed anywhere — replace it with a case-insensitive comparator used by both `ParseMethod` and `FindHeader`:

```suggestion
        bool EqualsIgnoreCase(std::string_view a, std::string_view b)
        {
            return std::equal(a.begin(), a.end(), b.begin(), b.end(), [](unsigned char l, unsigned char r) {
                return std::tolower(l) == std::tolower(r);
            });
        }
```

Coordinated with the `ParseMethod` and `FindHeader` suggestions — batch all three. (`<string_view>` comes in transitively via `UrlLib.h`.)

Polyfills/Fetch/Source/Fetch.cpp
```diff
@@ -0,0 +1,404 @@
...
+            const std::string lowerName = ToLower(name);
+            for (const auto& header : data.headers)
+            {
+                if (ToLower(header.first) == lowerName)
+                {
+                    return header.second;
+                }
+            }
+            return std::nullopt;
+        }
```
bghgary: Each `get`/`has` allocates `lowerName` plus a `ToLower` temporary for every header scanned. Use the non-allocating comparator instead:

```suggestion
        std::optional<std::string> FindHeader(const ResponseData& data, std::string_view name)
        {
            for (const auto& header : data.headers)
            {
                if (EqualsIgnoreCase(header.first, name))
                {
                    return header.second;
                }
            }
            return std::nullopt;
        }
```

Polyfills/Fetch/Source/Fetch.cpp
```diff
@@ -0,0 +1,404 @@
...
+                for (uint32_t i = 0; i < names.Length(); ++i)
+                {
+                    const auto key = names.Get(i);
+                    request.SetRequestHeader(key.ToString().Utf8Value(), object.Get(key).ToString().Utf8Value());
+                }
+            }
+        }
+
+        Napi::Object BuildResponse(Napi::Env env, const std::shared_ptr<ResponseData>& data);
+
```
bghgary: Unnecessary forward declaration — `BuildHeaders` (the only thing before the L184 definition) doesn't call `BuildResponse`, and the recursive `clone` call is inside `BuildResponse`'s own body where its name is already in scope. Remove it:

```suggestion
```

Polyfills/Fetch/Source/Fetch.cpp
```diff
@@ -0,0 +1,404 @@
...
+            }, "json"));
+
+            response.Set("blob", Napi::Function::New(env, [data](const Napi::CallbackInfo& info) -> Napi::Value {
+                Napi::Env env = info.Env();
+                const auto deferred = Napi::Promise::Deferred::New(env);
+
+                // Use IsUndefined()/IsNull() rather than IsFunction() to detect the Blob
+                // polyfill: some JavaScriptCore/JSI builds classify constructor functions as
+                // typeof 'object', so napi_typeof reports napi_object and IsFunction() would
+                // incorrectly reject even when the Blob polyfill is installed.
```
bghgary, line 235: The workaround is fine, but the root cause is a Node-API-JSC bug worth tracking: `napi_typeof` (`js_native_api_javascriptcore.cc`) decides function-ness solely via `JSObjectIsFunction`, which returns false for `JSObjectMakeConstructor` constructors on some builds (libjavascriptcoregtk) — so it reports `napi_object` where V8 returns `napi_function`. A fix there (also consult `JSObjectIsConstructor`) would let this *and* the File polyfill's identical workaround be removed. Worth a tracking issue rather than spreading per-polyfill copies.

bghgary (changes requested):

Clean polyfill. One correctness fix and one optional cleanup inline.

Polyfills/Fetch/Source/Fetch.cpp
bkaradzic-microsoft: Fixed in ac2019f — split into two chained `.then()` continuations: the first does the work and resolves, the second inspects the `expected` result and surfaces any error (including a throw from `BuildResponse`) as a promise rejection. The scheduler is co-owned by the second continuation so it outlives the in-flight request. Good catch — this would have hung `await fetch(...)` on any post-completion throw.

Polyfills/Fetch/Source/Fetch.cpp
bkaradzic-microsoft: Done — collapsed the two catch blocks into a single `catch (...)` using the `exception_ptr` `Error` overload, removing the `Napi::Error`-vs-`std::exception` prefix asymmetry.

Polyfills/Fetch/Source/Fetch.cpp
bkaradzic-microsoft: Done — `ParseMethod` now uses `EqualsIgnoreCase`. Agreed the GET/POST → `UrlMethod` mapping ideally belongs in UrlLib (which owns `UrlMethod`); leaving that as a follow-up as you suggested.

Polyfills/Fetch/Source/Fetch.cpp
bkaradzic-microsoft: Agreed — the `statusText` table is a workaround for UrlLib exposing only the numeric code. The real fix belongs in UrlLib (expose the reason phrase, falling back to a code→text table for HTTP/2+ where the wire carries none), which would also give XHR a `statusText` for free. Tracked as a follow-up in #193 (not this PR).

Polyfills/Fetch/Source/Fetch.cpp
bkaradzic-microsoft: Done — `ResponseData::body` is now a plain `std::vector<std::byte>` and the accessor/`clone` use sites were updated to `data->body.`. You're right that the inner `shared_ptr` only added an allocation/indirection and falsely implied clones get independent bodies.

Polyfills/Fetch/Source/Fetch.cpp
bkaradzic-microsoft: Done — added the non-allocating `EqualsIgnoreCase(string_view, string_view)` comparator and removed `ToLower`; it's now used by both `ParseMethod` and `FindHeader`. (Added an explicit `#include <string_view>` rather than rely on the transitive include from `UrlLib.h`.)

Polyfills/Fetch/Source/Fetch.cpp
bkaradzic-microsoft: Done — `FindHeader` now uses `EqualsIgnoreCase`, so no more per-header `lowerName` allocation while scanning.

Polyfills/Fetch/Source/Fetch.cpp
bkaradzic-microsoft: Done — removed the unnecessary `BuildResponse` forward declaration.

Polyfills/Fetch/Source/Fetch.cpp
bkaradzic-microsoft: Agreed — root cause is the Node-API-JSC `napi_typeof` path deciding function-ness solely via `JSObjectIsFunction`, which returns false for `JSObjectMakeConstructor` constructors on some JSC builds, so it reports `napi_object` where V8 returns `napi_function`. The File polyfill carries the identical workaround. A fix there (also consulting `JSObjectIsConstructor`) would let both drop the workaround rather than spreading per-polyfill copies. Tracked in #194; keeping the workaround here for now.

bkaradzic-microsoft:
Thanks for the review, @bghgary — all comments are addressed in ac2019f, and the follow-ups you flagged are now resolved/tracked.

**Correctness fix (the blocker):** the `SendAsync()` continuation is split into two chained `.then()`s — the first does the work and resolves, the second inspects the `expected` result and rejects on error. So a throw inside the continuation (e.g. from `BuildResponse`) now settles the promise as a rejection instead of being captured into a discarded result and dropped, which previously hung `await fetch(...)`. The scheduler is co-owned by the second continuation so it outlives the in-flight request.

**Cleanups:** single `catch (...)` via the `exception_ptr` `Napi::Error` overload (removes the `Napi::Error`-vs-`std::exception` prefix asymmetry); non-allocating `EqualsIgnoreCase(string_view, string_view)` replacing `ToLower`, used by both `ParseMethod` and `FindHeader`; `ResponseData::body` is now a plain `std::vector<std::byte>`; removed the unnecessary `BuildResponse` forward declaration.

**`statusText` "belongs in UrlLib" — done.** [BabylonJS/UrlLib#30](https://github.com/BabylonJS/UrlLib/pull/30) is merged and exposes `UrlRequest::StatusText()` (wire reason phrase where available, else a canonical code→text fallback table). The stacked follow-up #195 drops the private table here and reads `request.StatusText()` into `Response.statusText` (and gives `XMLHttpRequest.statusText` for free). Tracked + closed via #193.

**Node-API-JSC `napi_typeof` bug** (the `blob()` `IsFunction()` workaround): root cause is `napi_typeof` deciding function-ness solely via `JSObjectIsFunction`, which returns false for `JSObjectMakeConstructor` constructors on some JSC builds, so it reports `napi_object` where V8 returns `napi_function`. Tracked in #194.

CI is green. Re-requested your review — PTAL.

bghgary (approved):

LGTM

Issue #187, opened 2026-06-05
RFC: WebWorker interface via an N-API module


matthargett:
## Intent

Land the N‑API + underlying JS‑runtime updates needed to expose a **WebWorker interface as an N‑API module**, so the **Factotum** BabylonJS app can lift‑and‑shift from the browser/PWA into **BabylonNative** with minimal code change.

## Why this matters

- Factotum's data pipeline was moved onto a Web Worker; on‑device that gave a **~10 fps uplift** on Quest 1/2, Pico 4, and similar standalone headsets (pulling the pipeline off the render thread).
- To carry that architecture into BabylonNative, BabylonNative needs a Worker‑compatible surface — cleanest delivered as an **N‑API module** exposing the `Worker` interface (construct, `postMessage`, transferables, `onmessage`/`onerror`, `terminate`), backed by the runtime's threading/isolate model.
- **Revenue imperative**: in‑app purchases are broken/non‑existent for immersive PWAs on the **Meta Horizon Store** and **Pico Store** (even as of June 2026). Shipping as a native (BabylonNative) app is currently the only viable monetization path on those stores, so unblocking Factotum → BabylonNative is business‑critical for BabylonJS‑community devs whose complex BabylonJS apps depend on revenue (and storefront placement). 

## Why the N‑API track comes first

A Worker‑as‑N‑API‑module needs a solid, modern N‑API surface across the runtimes BabylonNative ships:

- N‑API conformance + cross‑engine parity (the #116 + v7 track) so addons behave the same on V8 / JSC / Chakra.
- A shared `libnapi` so in‑process addons resolve modules dynamically, allowing for less hacky build setups to exclude N-API plug-ins that aren't used and align to other runtimes (#183).
- On Android specifically, a JSC that can keep up (see the jsc‑android → Bun‑JSC RFC, #186) — Worker payloads use modern JS (structured clone, BigInt, transferables) that 2019 JSC can't even parse.

## Proposed shape (for discussion)

- An N‑API module exposing a `Worker` class: construct from a script/module; `postMessage` with structured‑clone + transferable `ArrayBuffer`s; `terminate`; `onmessage`/`onerror`.
- Backed by a second runtime instance/isolate on a worker thread, with a thread‑safe channel (`napi_threadsafe_function` or equivalent).
- Reuse BabylonNative's existing JS dispatch / task‑runner where possible.

## Open questions

[...]

Issue #186, opened 2026-06-05
RFC: replace abandoned jsc-android with a maintained JSC (Bun's WebKit fork) on Android


matthargett:
## Problem

Android JSC support in JsRuntimeHost rides on [`jsc-android`](https://github.com/react-native-community/jsc-android-buildscripts) (`org.webkit:android-jsc`), which is effectively abandoned:

- The build we consume (`250231.x` / `250230.x`) is **WebKit SVN r250230, dated 2019‑09‑23** — a ~6‑year‑old JSC.
- Last npm publish was ~2 years ago. React Native moved its default engine to Hermes and, as of RN 0.79, split JSC into a separate package — the original `react-native-community/jsc-android-buildscripts` is dormant.
- No active maintainer, corporate sponsor, or community is shipping updates.

**Concrete impact, hit during the N‑API v7 work** (#185 / the v7 track): this JSC has **no BigInt** — its parser rejects `0n` literals outright — so JSC can't reach N‑API v6/v7 or JS‑language parity with V8 on Android. We had to feature‑detect‑fail BigInt (`ENOTSUP`) on Android‑JSC. Every future JS/N‑API feature gap repeats this.

## Proposal

Adopt a **modern, actively‑maintained JSC** on Android and retire the 2019 `jsc-android`.

Primary candidate: **Bun's JSC** — Bun tracks a current JavaScriptCore via its [`oven-sh/WebKit`](https://github.com/oven-sh/WebKit) fork and release‑builds it continuously, so it stays current on language + JSC C API surface. bun already builds an N-API compatible runtime on top of this JSC, so it would seem to serve our purposes very well.

[...]

PR #189, opened 2026-06-06
N-API v7 across runtimes (V8 / JavaScriptCore / Chakra)

matthargett:
Third PR in the N-API stack (after #116 napi-tests, #183 napi-shared-android). Raises the default
NAPI_VERSION 5 -> 7 and brings the v6/v7 surface across every runtime. **Fork CI: 22/22 green.**

| Engine | BigInt | instance_data | ArrayBuffer detach |
|---|---|---|---|
| **V8** (Android/Win/Linux) | native | yes | yes |
| **JavaScriptCore** (macOS/iOS) | C-API (macOS 15+) + JS-eval fallback | yes | `transfer()` |
| **JavaScriptCore** (jsc-android ~2020) | `ENOTSUP` feature-detect (engine has no BigInt) | yes | n/a |
| **Chakra** (Win32/UWP) | `ENOTSUP` feature-detect | yes | — |
| **JSI** | n/a — backend is the node-addon-api C++ wrapper, not the C API | — | — |

Conformance is honest per engine: `test_bigint` runs for real on V8 + Apple JSC; a literal-free
`test_bigint_unsupported` fallback asserts the `ENOTSUP` throw on engines with no BigInt (jsc-android,
Chakra). `test_instance_data` runs everywhere v6 is supported. detach + get_all_property_names land at
the v9 bump (their vendored tests need `node_api_basic_env`); the detach feature is implemented now.

Notable: jsc-android (~2020) has no BigInt at all — its parser rejects `0n` literals — so the JSC
backend feature-detects BigInt at env init and reports `ENOTSUP`, exactly like Chakra. The JSI backend
implements only the C++ `Napi::` wrapper (zero C `napi_*`), so the C-API conformance suite is gated off
it; covering JSI would mean a full C-API-over-jsi port, out of scope here.

Also fixed several pre-existing cross-platform CI failures this surfaced (first broad fork CI of the
napi stack): node_lite engine selection by compile-define (Android-JSC, Linux-V8), `<cstdint>` on
Linux, napi-into-addon linking on Windows/UWP, the GSL `C4875` warnings-as-error toolchain drift, the
JSI shared-header include path, and on-device surfacing of in-process node_lite failures.

Draft — stacked; base is upstream `main` so the diff includes the #116 + #183 commits until those merge
(merge order: #183, then #116, then this). Validated 22/22 on the fork CI matrix via the twin rebeckerspecialties/JsRuntimeHost#4.

---

## Landing sequence

Inter-related N-API PRs; intended order (✅ done / ⏳ pending):

1. ⏳ **#185** — Fix Windows/UWP CI build (MSVC C4875 + JSI include path). Independent build/toolchain fix; lands first so CI is green under everything below. _(fork twin #185 — 22/22 green)_
2. ⏳ **#183** — Build napi as a shared library on Android (so in-process addons resolve `napi_*`).
3. ⏳ **#116** — N-API compliance tests (the Android addons resolve `napi_*` via the shared `libnapi.so` from the step above).
4. ⏳ **N-API v7 across runtimes (V8 / JSC / Chakra)** — this PR. _(fork twin rebeckerspecialties/JsRuntimeHost#4 — 22/22 green on the fork CI matrix.)_  ← **this PR**

Motivating RFCs: #186 (jsc-android → maintained JSC) · #187 (WebWorker via N-API → Factotum into BabylonNative).


Core/Node-API/Include/Shared/napi/js_native_api.h
```diff
@@ -3,7 +3,7 @@
 
 // [BABYLON-NATIVE-ADDITION]
 #ifndef NAPI_VERSION
-#define NAPI_VERSION 5
+#define NAPI_VERSION 7
```
bghgary: Keep the default at 5 until every selectable engine satisfies the complete tier, or set `NAPI_VERSION` and `napi_get_version` per engine. The in-tree V8, QuickJS, JavaScriptCore, and Chakra backends advertise at least version 7, but none defines `napi_get_all_property_names`; QuickJS also lacks instance-data and BigInt-word APIs, misreports scalar BigInt losslessness, and leaves an exception pending after a wrong-type `napi_is_detached_arraybuffer` query; Chakra and some JavaScriptCore builds return `ENOTSUP` for required APIs, and Chakra detach returns that pending exception even for wrong types instead of `napi_arraybuffer_expected`. An addon that feature-detects with `napi_get_version` can then call missing or nonconformant version 6/7 APIs.

Core/Node-API/Include/Shared/napi/napi.h
```diff
@@ -8,8 +8,13 @@
...
 #define NODE_ADDON_API_DISABLE_NODE_SPECIFIC
 #endif
+// [BABYLON-NATIVE-ADDITION] Keep this in step with js_native_api.h / js_native_api_types.h. The C
+// headers already default to 7, but this one is reached first by every consumer that includes
+// <napi/napi.h> (all of Babylon Native, and every node-addon-api addon), and #ifndef made it win --
+// so Napi::BigInt, Value::IsBigInt, GetInstanceData<T> and Napi::Addon stayed compiled out even
+// though the C entry points behind them were implemented.
 #ifndef NAPI_VERSION
-#define NAPI_VERSION 5
+#define NAPI_VERSION 7
```
bghgary: Activating the version 7 C++ wrappers also exposes a stale ArrayBuffer cache. `EnsureInfo()` re-queries only while `_data` is null, and `Detach()` does not clear `_data` or `_length`; after `Data()` or `ByteLength()` caches a non-empty buffer, wrapper-side or JavaScript-side detachment leaves those accessors returning the old pointer and length. Invalidate or re-query the cached state and cover both detach paths.

Core/Node-API/Source/js_native_api_javascriptcore.cc
```diff
@@ -1671,6 +1896,290 @@ napi_status napi_get_global(napi_env env, napi_value* result) {
...
+  // Match Node/V8 limits *before* touching the (possibly under-sized) words buffer: beyond INT_MAX
+  // it's napi_invalid_arg; beyond the engine's BigInt size limit it's a RangeError.
+  if (word_count > static_cast<size_t>(INT_MAX)) {
+    return napi_set_last_error(env, napi_invalid_arg);
+  }
+  if (word_count > (static_cast<size_t>(1) << 24)) {  // ~ kMaxBigIntLengthBits / 64
+    napi_throw_range_error(env, nullptr, "Maximum BigInt size exceeded");
+    return napi_set_last_error(env, napi_pending_exception);
+  }
+  if (word_count > 0) {
```
bghgary: Match Node's null-argument contract for the BigInt word APIs. `napi_create_bigint_words` requires `words` even when `word_count` is zero; `napi_get_value_bigint_words` permits null `sign_bit` and `words` only together for a count query, otherwise both are required. This implementation accepts null `words` at zero length and mismatched output pairs, returning `napi_ok` where V8 returns `napi_invalid_arg`. Add those argument-status cases.

Core/Node-API/Source/js_native_api_javascriptcore.cc
```diff
@@ -1671,6 +1896,290 @@ napi_status napi_get_global(napi_env env, napi_value* result) {
...
+napi_status napi_get_value_bigint_int64(napi_env env,
+                                        napi_value value,
+                                        int64_t* result,
+                                        bool* lossless) {
+  CHECK_ENV(env);
+  CHECK_ARG(env, value);
+  CHECK_ARG(env, result);
+  CHECK_ARG(env, lossless);
+  JSValueRef low{};
+  CHECK_NAPI(BigIntLow64(env, value, env->bigint_as_int_n, &low));
```
bghgary: Return `napi_bigint_expected` without creating an exception when `value` is not a BigInt. These getters currently pass any value into the JavaScript BigInt intrinsics, so numbers and objects become `napi_pending_exception`; V8 returns the required status directly. Add wrong-type status and pending-exception coverage for all three getters.

Core/Node-API/Source/js_native_api_javascriptcore.cc
```diff
@@ -1671,6 +1896,290 @@ napi_status napi_get_global(napi_env env, napi_value* result) {
...
+  if (env->arraybuffer_detached_getter != nullptr) {
+    JSObjectRef getter = JSValueToObject(env->context, env->arraybuffer_detached_getter, &exception);
+    CHECK_JSC(env, exception);
+    JSValueRef detached = JSObjectCallAsFunction(env->context, getter, ab, 0, nullptr, &exception);
+    CHECK_JSC(env, exception);
+    *result = JSValueToBoolean(env->context, detached);
+  } else {
+    // Pre-ES2024 fallback: a detached buffer has no backing store.
+    JSValueRef ignored{};
+    *result = JSObjectGetArrayBufferBytesPtr(env->context, ab, &ignored) == nullptr;
```
bghgary: Match the detach contract before invoking JavaScript intrinsics: `napi_is_detached_arraybuffer` must return `napi_ok`, `false`, and no exception for a non-ArrayBuffer; `napi_detach_arraybuffer` must return `napi_arraybuffer_expected` for a non-ArrayBuffer and `napi_detachable_arraybuffer_expected` for a non-detachable or already-detached ArrayBuffer. Add focused compatibility cases for normal, wrong-type, null-backed zero-length external, non-detachable, already-detached, and successfully detached external buffers; the upstream null-backed case is expected to report detached.

Tests/NodeApi/CMakeLists.txt
```diff
@@ -0,0 +1,265 @@
...
+# v6/v7 conformance addons, enabled per engine by what that engine actually implements.
+#
+# This is an ALLOW-list on purpose: upstream keeps adding backends (QuickJS, Hermes, JSI), and each new
+# one otherwise silently picks up addons whose napi_* symbols it does not export, failing at link
+# (e.g. QuickJS has napi_create_bigint_int64 but no *_words and no napi_add_finalizer). Add an engine
+# here only once its v6/v7 surface is implemented and the suite has been run against it.
+#
+# NOTE: detach (v7, test_typedarray) + get_all_property_names (v6, test_object) use v9 APIs
+# (node_api_basic_env) and enable at the v9 bump.
+set(JSR_NODE_API_V6_V7_ENGINES V8 JavaScriptCore Chakra)
```
bghgary: Run `NodeApiTests` in desktop CI before treating this as conformance coverage. The current Win32, Linux, and macOS workflows execute only `UnitTests`, so Chakra and desktop V8/JavaScriptCore never run this suite; UWP remains build-only and provides no runtime coverage. Also add or adapt focused `napi_get_all_property_names` and detach cases that exercise versions 6 and 7 without depending on version 9 APIs.

Tests/UnitTests/Android/app/src/main/cpp/CMakeLists.txt
```diff
@@ -55,4 +79,26 @@ target_link_libraries(UnitTestsJNI
...
+# implemented and exercised run these. A backend added upstream later (e.g. QuickJS, which has
+# napi_create_bigint_int64 but no napi_add_finalizer and no *_words) must be opted in deliberately
+# rather than inheriting them by default.
+#
+# The v1-v5 suite, v6 instance_data, and BigInt all run on both Android engines: V8, and
+# JavaScriptCore now that jsc-android is pinned to r294992 (BigInt on by default; r250231 gated it off).
+set(JSR_NODE_API_HARNESS_ENGINES V8 JavaScriptCore)
+if(NAPI_JAVASCRIPT_ENGINE IN_LIST JSR_NODE_API_HARNESS_ENGINES)
+    target_compile_definitions(UnitTestsJNI PRIVATE
+        NODE_API_AVAILABLE_NATIVE_TESTS="2_function_arguments,3_callbacks,4_object_factory,5_function_factory,test_reference_double_free,test_instance_data,test_bigint")
```
bghgary: Adapt the Android instance-data test so it verifies the teardown finalizer. node_lite never assigns `require.main`, so this test takes its in-process branch: set/get is exercised, but the child/worker branch that checks `deleting addon data` never runs, and no Android assertion observes the finalizer. Add an in-process teardown signal that the test can assert before claiming instance-data lifecycle coverage.

PR #198, opened 2026-06-13
TextDecoder: accept all WHATWG utf-8 encoding labels

bkaradzic-microsoft:
## What

`TextDecoder`'s constructor only accepted the exact labels `"utf-8"` and `"UTF-8"`, throwing for every other spelling. This PR makes it accept all WHATWG-spec UTF-8 labels.

## Why

Per the [WHATWG Encoding Standard](https://encoding.spec.whatwg.org/#concept-encoding-get), an encoding label is matched after stripping leading/trailing ASCII whitespace and ASCII-lowercasing, and several labels all decode as UTF-8: `utf-8`, `utf8`, `unicode-1-1-utf-8`, `unicode11utf8`, `unicode20utf8`, `x-unicode20utf8`.

Real consumers rely on this. The Babylon.js glTF/Draco loader constructs `new TextDecoder("utf8")` (no hyphen). With the old check that threw, decoding aborted mid-load. In **Babylon Native** the aborted load left the loader in a state that drove a **native out-of-bounds write**, which surfaced as non-deterministic **heap corruption** (`STATUS_HEAP_CORRUPTION`, `0xC0000374`) on the Draco mesh-compression validation tests.

## Fix

Normalize the label per the spec (trim ASCII whitespace + ASCII-lowercase) and accept the full set of UTF-8 labels; still throw for genuinely unsupported encodings.

## Verification

- Two BabylonNative Playground Draco validation tests that previously crashed with heap corruption (`GLTF Serializer KHR draco mesh compression`, `GLTF Buggy with Draco Mesh Compression`) now **pass** (3/3 runs each) with this fix vendored in; a third (`GLTF Box with bad Draco normalized flag`) no longer crashes.
- Adds unit tests covering the `utf8` label, case/whitespace variants, the other UTF-8 aliases, and a still-rejected encoding (`utf-16`).

---

## Update: also fixes a latent ChakraCore N-API heap-corruption bug

CI surfaced that the new "throws for an unsupported encoding" test reliably **heap-corrupts on Chakra** (Win32 x64/x86) — `STATUS_HEAP_CORRUPTION` (0xC0000374) / access violations, crashing later in unrelated tests. This is a **pre-existing latent bug** that this PR's first constructor-throw test merely exposed (it does not reproduce on `main` only because no existing test throws from a wrapped constructor; V8/JSC/JSI and the sanitizer jobs are unaffected since the bug is in the Chakra N-API shim).

**Root cause:** when a napi `ObjectWrap` constructor throws, the C++ instance is destroyed during stack unwinding, but the wrap finalizer registered by `napi_wrap()` stays attached to `this`. A later GC then runs the finalizer on the freed instance → use-after-free / double-free. The addon-api `~ObjectWrap()` is meant to detach the wrap on failure, but it obtains `this` via `napi_get_reference_value()` on the wrap's weak (refcount-0) reference — and the Chakra backend returns `null` for **every** refcount-0 reference (the in-box `jsrt` API can't track weak-reference liveness), so the detach is skipped.

**Fix** (`Core/Node-API/Source/js_native_api_chakra.cc`):
- In the Chakra construct-call trampoline, if the callback leaves a pending JS exception, detach any wrap left on `this` via `napi_remove_wrap()` (preserving the pending exception) so the finalizer can't run on the freed instance.
- Guard `napi_remove_wrap()` against a null `result` out-param (now called with `nullptr`, matching upstream).

**Verification:** built the Chakra UnitTests locally; the constructor-throw scenario went from **heap corruption every run** to **clean (8/8 runs, all tests passing)**. Added a regression test that throws from a wrapped constructor 100× then exercises the heap.

> Note: the N-API engine fix is logically independent of the TextDecoder label change and could be split into its own PR if preferred — it is bundled here because this PR's constructor-throw test is what exposes it and is its regression test.

### DIFF
--- a/Polyfills/TextDecoder/Source/TextDecoder.cpp
+++ b/Polyfills/TextDecoder/Source/TextDecoder.cpp
@@ -6,6 +6,36 @@
 
 namespace
 {
+    // Normalize an encoding label per the WHATWG Encoding Standard "get an encoding"
+    // algorithm: strip leading/trailing ASCII whitespace and ASCII-lowercase the result.
+    std::string NormalizeEncodingLabel(const std::string& encoding)
+    {
+        const auto isAsciiWhitespace = [](char c) {
+            return c == '\t' || c == '\n' || c == '\f' || c == '\r' || c == ' ';
+        };
+
+        size_t begin = 0;
+        size_t end = encoding.size();
+        while (begin < end && isAsciiWhitespace(encoding[begin]))
+        {
+            ++begin;
+        }
+        while (end > begin && isAsciiWhitespace(encoding[end - 1]))
+        {
+            --end;
+        }
+
+        std::string label = encoding.substr(begin, end - begin);
+        for (auto& c : label)
+        {
+            if (c >= 'A' && c <= 'Z')
+            {
+                c = static_cast<char>(c - 'A' + 'a');
+            }
+        }
+        return label;
+    }
+
     class TextDecoder final : public Napi::ObjectWrap<TextDecoder>
     {
     public:
@@ -34,32 +64,9 @@ namespace
             {
                 auto encoding = info[0].As<Napi::String>().Utf8Value();
 
-                // Normalize per the WHATWG Encoding Standard "get an encoding" algorithm:
-                // strip leading/trailing ASCII whitespace and lowercase before matching the
-                // label. Several labels (e.g. "utf8", "unicode-1-1-utf-8") all map to UTF-8;
-                // callers such as the glTF/Draco loader pass "utf8".
-                const auto isAsciiWhitespace = [](char c) {
-                    return c == '\t' || c == '\n' || c == '\f' || c == '\r' || c == ' ';
-                };
-                size_t begin = 0;
-                size_t end = encoding.size();
-                while (begin < end && isAsciiWhitespace(encoding[begin]))
-                {
-                    ++begin;
-                }
-                while (end > begin && isAsciiWhitespace(encoding[end - 1]))
-                {
-                    --end;
-                }
-                std::string label = encoding.substr(begin, end - begin);
-                for (auto& c : label)
-                {
-                    if (c >= 'A' && c <= 'Z')
-                    {
-                        c = static_cast<char>(c - 'A' + 'a');
-                    }
-                }
-
+                // Several labels (e.g. "utf8", "unicode-1-1-utf-8") all map to UTF-8 after
+                // normalization; callers such as the glTF/Draco loader pass "utf8".
+                const std::string label = NormalizeEncodingLabel(encoding);
                 if (label != "utf-8" &&
                     label != "utf8" &&
                     label != "unicode-1-1-utf-8" &&
--- a/Polyfills/TextDecoder/Source/TextDecoder.cpp
+++ b/Polyfills/TextDecoder/Source/TextDecoder.cpp
@@ -33,7 +33,39 @@ namespace
             if (info.Length() > 0 && info[0].IsString())
             {
                 auto encoding = info[0].As<Napi::String>().Utf8Value();
-                if (encoding != "utf-8" && encoding != "UTF-8")
+
+                // Normalize per the WHATWG Encoding Standard "get an encoding" algorithm:
+                // strip leading/trailing ASCII whitespace and lowercase before matching the
+                // label. Several labels (e.g. "utf8", "unicode-1-1-utf-8") all map to UTF-8;
+                // callers such as the glTF/Draco loader pass "utf8".
+                const auto isAsciiWhitespace = [](char c) {
+                    return c == '\t' || c == '\n' || c == '\f' || c == '\r' || c == ' ';
+                };
+                size_t begin = 0;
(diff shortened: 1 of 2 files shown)


Polyfills/TextDecoder/Source/TextDecoder.cpp
```diff
@@ -33,7 +33,39 @@ namespace
...
+                }
+
+                if (label != "utf-8" &&
+                    label != "utf8" &&
+                    label != "unicode-1-1-utf-8" &&
+                    label != "unicode11utf8" &&
+                    label != "unicode20utf8" &&
+                    label != "x-unicode20utf8")
                 {
                     throw Napi::Error::New(Env(), "TextDecoder: unsupported encoding '" + encoding + "', only 'utf-8' is supported");
```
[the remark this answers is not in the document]
bkaradzic-microsoft: Good point — updated in 011af13: the rejection message now reads "only UTF-8 is supported" instead of implying the single exact spelling `'utf-8'`.

bkaradzic-microsoft:
Thanks for the review. Addressing the one inline comment, plus a heads-up on additional commits:

- **Error-message wording** (the inline comment): fixed in 011af13 — the rejection now reads "only UTF-8 is supported" rather than implying the single exact spelling `'utf-8'`.
- **New engine fix** since the review: while validating this PR, CI exposed a latent **ChakraCore N-API heap-corruption bug** that the new "throws for an unsupported encoding" test triggers (throwing from an `ObjectWrap` constructor left the `napi_wrap` finalizer attached to a freed instance → use-after-free on the next GC). Fixed in be54234 with a regression test. Details are in the updated PR description. CI is now green (22/22), including the Win32 Chakra jobs that were failing.

@copilot a fresh review of be54234 (the `js_native_api_chakra.cc` change) would be welcome.

bghgary (changes requested):

Please split the Chakra N-API ObjectWrap finalizer-after-throw fix into its own PR. It's an independent heap-corruption fix, and a dedicated PR makes it far easier to find/bisect later than buried under a TextDecoder label change.

Polyfills/TextDecoder/Source/TextDecoder.cpp
```diff
@@ -33,9 +33,41 @@ namespace
...
+                    --end;
+                }
+                std::string label = encoding.substr(begin, end - begin);
+                for (auto& c : label)
+                {
+                    if (c >= 'A' && c <= 'Z')
+                    {
+                        c = static_cast<char>(c - 'A' + 'a');
+                    }
+                }
```
bghgary: Readability nit: factor this normalization into a named helper (e.g. `NormalizeEncodingLabel`) so the constructor reads as intent rather than inline character arithmetic.

bkaradzic-microsoft:
Done — split the Chakra N-API ObjectWrap finalizer-after-throw fix into its own PR: https://github.com/BabylonJS/JsRuntimeHost/pull/199 (the `js_native_api_chakra.cc` change + its heap-corruption regression test). This PR is now TextDecoder-only.

Polyfills/TextDecoder/Source/TextDecoder.cpp
bkaradzic-microsoft: Done — extracted the WHATWG trim + ASCII-lowercase normalization into a `NormalizeEncodingLabel(const std::string&)` helper in the anonymous namespace; the constructor now reads `const std::string label = NormalizeEncodingLabel(encoding);` then matches against the UTF-8 labels. No behavior change.

bkaradzic-microsoft:
Correction to my earlier note about merge order: these are **not** independent on Chakra — **#198 depends on #199**.

CI confirms it: #198's `should still throw for a genuinely unsupported encoding` test throws from the `TextDecoder` constructor, which on ChakraCore (Win32 x64/x86) triggers the exact ObjectWrap finalizer-after-throw heap corruption that #199 fixes. The build crashes right after that test with `0xC0000005` (the freed-finalizer use-after-free). All other platforms (V8/JSI/Ubuntu/macOS/UWP) pass.

So the order is: **merge #199 first**, then #198 (rebased on main) goes green on Chakra. The two Chakra failures here are expected until #199 lands.

bkaradzic-microsoft:
Done — the Chakra N-API ObjectWrap finalizer-after-throw fix is split out and landed separately as #199, so this PR is now **TextDecoder-only**: the three commits are the WHATWG utf-8 label change plus the two review fixes (clearer "only UTF-8 is supported" message, and the extracted NormalizeEncodingLabel helper).

I rebased onto current `main` (which already includes #199) to resolve the conflict. The Chakra regression test that #199 added (`throwing from the constructor repeatedly does not corrupt native state`) and the new WHATWG label tests both live in the `TextDecoder` describe block and pass together — verified on Win32/Chakra (198 passing).

Could you re-review to clear the change request? The diff is now just `Polyfills/TextDecoder/Source/TextDecoder.cpp` (+43) and `Tests/UnitTests/Scripts/tests.ts` (+24).

bghgary (approved):

LGTM. Trim the stale "ChakraCore N-API heap-corruption" section from the description — that fix is now #199.

PR #199, opened 2026-06-16
Node-API (Chakra): fix heap corruption when an ObjectWrap constructor throws

bkaradzic-microsoft:
## What

Fixes a heap corruption in the Chakra Node-API implementation when an `ObjectWrap`-based constructor throws.

Split out of https://github.com/BabylonJS/JsRuntimeHost/pull/198 as an independent fix, per review.

## Problem

When a wrapped (`napi_wrap`) constructor throws, the C++ `ObjectWrap` instance is destroyed by stack unwinding inside the callback, but the wrap finalizer registered by `napi_wrap()` stays attached to `this`. `~ObjectWrap()` cannot detach it, because `napi_get_reference_value()` returns null for the wrap's weak (refcount 0) reference. A later GC then runs the finalizer on the freed native instance — a use-after-free that surfaces as non-deterministic `STATUS_HEAP_CORRUPTION` (`0xC0000374`).

## Fix

`Core/Node-API/Source/js_native_api_chakra.cc`:
- In `ExternalCallback`, after a construct call that left a pending JS exception, detach the wrap (`napi_remove_wrap`) before returning, then re-set the exception so it is preserved across the cleanup.
- Guard `napi_remove_wrap` against a `nullptr` `result` out-param (the finalizer-cleanup call above passes `nullptr`), so it no longer dereferences a null pointer.

## Test

Adds a regression test (`Tests/UnitTests/Scripts/tests.ts`) that throws from a wrapped constructor 100× to create many dangling wraps, then allocates/decodes to exercise the heap and surface any corruption within the run. It uses `TextDecoder` as a convenient throwing wrapped constructor; the test passes on `main`'s existing `TextDecoder` behavior and is independent of the TextDecoder label change in #198.

## Verification

Reliably reproduced as `STATUS_HEAP_CORRUPTION` on Chakra (Win32 x64/x86) before the fix; passes after.

### DIFF
--- a/Tests/UnitTests/Scripts/tests.ts
+++ b/Tests/UnitTests/Scripts/tests.ts
@@ -1492,7 +1492,7 @@ describe("TextDecoder", function () {
     });
 
     it("throwing from the constructor repeatedly does not corrupt native state", function () {
-        // Regression for a ChakraCore N-API ObjectWrap bug: when a wrapped
+        // Regression for a Chakra N-API ObjectWrap bug: when a wrapped
         // constructor throws, the native instance is destroyed during stack
         // unwinding but the wrap finalizer stayed attached to `this`, so a
         // later GC ran the finalizer on freed memory (heap corruption). Throw
--- a/Tests/UnitTests/Scripts/tests.ts
+++ b/Tests/UnitTests/Scripts/tests.ts
@@ -1490,6 +1490,20 @@ describe("TextDecoder", function () {
         expect(result).to.equal("H\0i");
         expect(result.length).to.equal(3);
     });
+
+    it("throwing from the constructor repeatedly does not corrupt native state", function () {
+        // Regression for a ChakraCore N-API ObjectWrap bug: when a wrapped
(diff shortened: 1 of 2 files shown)


Tests/UnitTests/Scripts/tests.ts
```diff
@@ -1490,6 +1490,20 @@ describe("TextDecoder", function () {
         expect(result).to.equal("H\0i");
         expect(result.length).to.equal(3);
     });
+
+    it("throwing from the constructor repeatedly does not corrupt native state", function () {
+        // Regression for a ChakraCore N-API ObjectWrap bug: when a wrapped
```
bghgary: Chakra (in-box `chakra.dll`), not ChakraCore (the archived fork).

```suggestion
        // Regression for a Chakra N-API ObjectWrap bug: when a wrapped
```

bghgary (approved):

LGTM. The PR description's opening line also says "ChakraCore" — this is the in-box Chakra engine (`chakra.dll`), not the archived fork; should be Chakra.

Issue #196, opened 2026-06-10
fetch/XHR transport failures and unhandled rejections lose all diagnostic fidelity before reaching the embedding app


matthargett:
Applications embedding Babylon Native depend on field crash reports for ongoing maintenance, and today a `fetch()`/`XMLHttpRequest` failure arrives there with no usable signal: DNS failure, connection refused, TLS rejection, proxy auth failure, server outage, and a missing bundled `app:///` asset are all byte-identical, and an un-caught rejection never reaches the host (BN) application at all. This issue tracks the chain of fixes; each hop is independently shippable.

### The chain (where fidelity is lost today)

1. ~~**UrlLib swallows transport errors at the source.**~~ ✅ **Fixed by BabylonJS/UrlLib#31** (open): the Apple backend discarded the `NSError` (long-standing in-code TODO) and the curl backend caught and dropped its own failure, leaving only `StatusCode() == 0`. UrlLib now exposes `ErrorString()` / `ErrorSymbol()` / `ErrorCode()`, normalized as `"<domain>:<symbol>(<code>): <detail>"` (e.g. `curl:CURLE_COULDNT_RESOLVE_HOST(6): ...`, `nsurl:NSURLErrorCannotConnectToHost(-1004): ...`, `urllib:AppResourceNotFound(0): ...`) — stable tokens for observability-pipeline filtering, purely additive (status-0 contract unchanged).

2. **The fetch polyfill flattens every transport failure to a constant string.** [`Fetch.cpp`](https://github.com/BabylonJS/JsRuntimeHost/blob/main/Polyfills/Fetch/Source/Fetch.cpp#L346-L353) throws `std::runtime_error{"fetch: network request failed"}`, discarding `result.error()` and the URL/method context it has in scope; the rejection surfaces as a plain `Error` (browsers/Node/Bun: `TypeError`) with no `cause`, no `code`, and a `.stack` snapshotted inside the scheduler tick — i.e. zero user frames. Proposed shape, once the UrlLib pin includes #31:
   - reject with a **`TypeError`** whose message is **stable** (`"fetch failed"` style — keeps crash-report grouping intact), carrying the variable detail as properties: `cause` (message from `ErrorString()`), `code` (`ErrorSymbol()`), `url`;
   - capture the JS call-site stack **synchronously inside `fetch()`** before `SendAsync()` (create the rejection Error, or a stack carrier, while user frames are still on the stack — the undici approach) so crash reports can attribute the failing call;
[...]

PR #200, opened 2026-06-16
Diagnosability: fetch/XHR error fidelity and modern AbortSignal (#196)

bkaradzic-microsoft:
Implements the fetch/XHR diagnosability work tracked in #196 — making transport failures observable to the embedding app instead of opaque — plus modern `AbortSignal` support.

Requires the UrlLib transport-error accessors merged in BabylonJS/UrlLib#33 (this PR bumps the UrlLib pin to `e86ffb3`).

> **Note:** the unhandled-promise-rejection hop that was originally part of this PR has been **split out into #204** for its own design discussion (it's engine-conditional — V8/Apple JSC only — and needs an `#ifdef`-free rewrite). Everything remaining here is uniform across all JS engines and carries no engine/platform preprocessor guards.

Each hop is a self-contained commit.

---

### Hop 2 — fetch/XHR transport-error fidelity

Previously every transport failure (DNS failure, connection refused, TLS rejection, missing `app:///` asset) was flattened: `fetch()` rejected with a plain `Error{"fetch: network request failed"}` (no `cause`/`code`/`url`, stack snapshotted in a worker tick = zero user frames); XHR exposed nothing beyond `status === 0`.

- `fetch` rejects with a **`TypeError`** (stable message `"fetch failed"`), detail nested under **`error.cause = {code, detail, url, status}`** (Node/undici shape, not top-level own-properties).
- `cause.code`/`cause.detail` come from UrlLib; present on Apple/Linux, omitted on Windows/Android while the standard shape (`TypeError` + stable message + `cause.url/status`) is preserved everywhere — a strict, additive superset of the spec. (This variance is a UrlLib HTTP-backend difference, not a JS-engine one.)
- The JS call-site **stack is captured synchronously** inside `fetch()` (via the global `Error` constructor, which materializes frames even on Chakra) and reattached to the rejection.
- `XMLHttpRequest` gains additive read-only **`errorCode`/`errorDetail`** accessors; its standard `error` event + `status === 0` behavior is unchanged.

### Hop 4 — `init.signal` + modern AbortSignal

`fetch()` ignored `init.signal` entirely (`arcana::cancellation::none()`), and the `AbortSignal` polyfill predated the modern spec.

- `AbortSignal`: `reason` (read-only), `throwIfAborted()`, static `AbortSignal.abort(reason?)`, read-only `aborted`, and a default **`AbortError`** reason (an `Error` whose `name` is `"AbortError"`; no `DOMException` polyfill exists). `AbortController.abort(reason?)` forwards the reason.
- `fetch` honors `init.signal` via its JS interface: an already-aborted signal rejects synchronously; an in-flight abort cancels the transport (`UrlRequest::Abort()`) and rejects with the signal's `reason`.

---

### Validation

Built and ran the UnitTests suite on **Win32 / Chakra** and **Win32 / V8**; CI is green across the full matrix (Chakra/V8/JSC/JSI × Win32/UWP/Android/iOS/macOS/Linux, incl. sanitizers).

New tests (all engine-agnostic): fetch `TypeError`+`cause` shape (refused / missing-asset), XHR `errorCode`/`errorDetail`, the modern `AbortSignal` API, and fetch `AbortError` (pre-aborted + in-flight).

### Cross-repo follow-up

`UrlRequest::Abort()` only cancels the Windows backend today — making it interrupt the curl/NSURLSession transports is a separate UrlLib change (noted in #196). `AbortSignal.timeout()` is also a follow-up.

### DIFF
--- a/Polyfills/Fetch/Source/Fetch.cpp
+++ b/Polyfills/Fetch/Source/Fetch.cpp
@@ -30,6 +30,32 @@ namespace Babylon::Polyfills::Internal
             std::vector<std::byte> body;
         };
 
+        // Shared state for honoring an AbortSignal passed via init.signal. Co-owned by the "abort"
+        // listener (which sets the flag, captures the reason, and cancels the transport) and the
+        // completion continuation (which reports the AbortError and tears the listener down).
+        struct AbortState
+        {
+            bool aborted{false};
+            Napi::Reference<Napi::Value> reason;
+            Napi::ObjectReference signal;
+            Napi::FunctionReference listener;
+        };
+
+        // The reason a fetch was aborted: the signal's `reason` (per the modern AbortSignal), or a
+        // fresh AbortError if the signal does not expose one.
+        Napi::Value GetAbortReason(Napi::Env env, const Napi::Object& signal)
+        {
+            const Napi::Value reason = signal.Get("reason");
+            if (!reason.IsUndefined() && !reason.IsNull())
+            {
+                return reason;
+            }
+
+            Napi::Error error = Napi::Error::New(env, "The operation was aborted.");
+            error.Set("name", Napi::String::New(env, "AbortError"));
+            return error.Value();
+        }
+
         bool EqualsIgnoreCase(std::string_view a, std::string_view b)
         {
             return std::equal(a.begin(), a.end(), b.begin(), b.end(), [](unsigned char l, unsigned char r) {
@@ -352,6 +378,7 @@ namespace Babylon::Polyfills::Internal
                     UrlLib::UrlMethod method = UrlLib::UrlMethod::Get;
                     std::optional<std::string> body;
                     Napi::Value headers = env.Undefined();
+                    Napi::Value signal = env.Undefined();
 
                     if (info.Length() > 1 && info[1].IsObject())
                     {
@@ -374,6 +401,7 @@ namespace Babylon::Polyfills::Internal
                         }
 
                         headers = init.Get("headers");
+                        signal = init.Get("signal");
                     }
 
                     auto request = std::make_shared<UrlLib::UrlRequest>();
@@ -390,6 +418,39 @@ namespace Babylon::Polyfills::Internal
                     // site rather than to an empty scheduler tick.
                     const std::string capturedStack = CaptureCallSiteStack(env);
 
+                    // Honor an AbortSignal passed via init.signal (WHATWG fetch). The signal is used
+                    // through its JS interface (aborted / reason / add/removeEventListener) so fetch
+                    // stays decoupled from the AbortController polyfill's C++ types.
+                    std::shared_ptr<AbortState> abortState;
+                    if (signal.IsObject())
+                    {
+                        const Napi::Object signalObject = signal.As<Napi::Object>();
+
+                        // Already aborted: reject synchronously with the signal's reason, never
+                        // touching the transport.
+                        if (signalObject.Get("aborted").ToBoolean().Value())
+                        {
+                            deferred.Reject(GetAbortReason(env, signalObject));
+                            return deferred.Promise();
+                        }
+
+                        abortState = std::make_shared<AbortState>();
+                        abortState->signal = Napi::Persistent(signalObject);
+
+                        Napi::Function listener = Napi::Function::New(env, [abortState, request, env](const Napi::CallbackInfo&) {
+                            if (!abortState->aborted)
+                            {
+                                abortState->aborted = true;
+                                abortState->reason = Napi::Persistent(GetAbortReason(env, abortState->signal.Value()));
+                                // Cancel the in-flight transport; the completion continuation then
+                                // rejects with the AbortError instead of a transport TypeError.
+                                request->Abort();
+                            }
+                        });
+                        abortState->listener = Napi::Persistent(listener);
+                        signalObject.Get("addEventListener").As<Napi::Function>().Call(signalObject, {Napi::String::New(env, "abort"), listener});
+                    }
+
                     // arcana::task::then captures the scheduler by reference (see arcana task.h) and
                     // invokes it on the worker thread when the request completes -- after this fetch()
                     // call has returned. A stack-local scheduler would therefore dangle. Heap-allocate
@@ -397,7 +458,28 @@ namespace Babylon::Polyfills::Internal
                     auto scheduler = std::make_shared<JsRuntimeScheduler>(JsRuntime::GetFromJavaScript(env));
                     request->SendAsync()
                         .then(*scheduler, arcana::cancellation::none(),
-                            [deferred, request, env, url, capturedStack](const arcana::expected<void, std::exception_ptr>& result) {
+                            [deferred, request, env, url, capturedStack, abortState](const arcana::expected<void, std::exception_ptr>& result) {
+                                // The request has settled: stop listening for aborts (breaking the
+                                // listener <-> abortState ownership cycle) before deciding the outcome.
+                                if (abortState)
+                                {
+                                    if (!abortState->signal.IsEmpty() && !abortState->listener.IsEmpty())
+                                    {
+                                        Napi::Object signalObject = abortState->signal.Value();
+                                        signalObject.Get("removeEventListener").As<Napi::Function>().Call(signalObject, {Napi::String::New(env, "abort"), abortState->listener.Value()});
+                                    }
+                                    abortState->listener.Reset();
+                                    abortState->signal.Reset();
+
+                                    if (abortState->aborted)
+                                    {
+                                        // Per the fetch spec, an aborted request rejects with the
+                                        // signal's reason (an AbortError), not a network error.
+                                        deferred.Reject(abortState->reason.Value());
+                                        return;
+                                    }
+                                }
+
                                 const int status = static_cast<int>(request->StatusCode());
 
                                 // Per the WHATWG fetch spec, only transport-level failures reject. A completed
--- a/Polyfills/Fetch/Source/Fetch.cpp
+++ b/Polyfills/Fetch/Source/Fetch.cpp
@@ -53,8 +53,13 @@ namespace Babylon::Polyfills::Internal
         // when an error is thrown) -- in that case the rejection simply carries no synthetic frames.
         std::string CaptureCallSiteStack(Napi::Env env)
         {
+            // Detect the global Error constructor with IsUndefined()/IsNull() rather than
+            // IsFunction(): some JavaScriptCore/JSI builds classify constructor functions as
+            // typeof 'object', so napi_typeof reports napi_object and IsFunction() would
+            // incorrectly skip stack capture even though Error is callable (see the Blob check
+            // below for the same rationale). Error is always present, so this guard is defensive.
             const Napi::Value errorCtor = env.Global().Get("Error");
-            if (!errorCtor.IsFunction())
+            if (errorCtor.IsUndefined() || errorCtor.IsNull())
             {
                 return {};
             }
--- a/Polyfills/Fetch/Source/Fetch.cpp
+++ b/Polyfills/Fetch/Source/Fetch.cpp
@@ -37,6 +37,80 @@ namespace Babylon::Polyfills::Internal
             });
         }
 
+        // Stable message used for every transport-failure rejection. Browsers and Node both keep
+        // this constant (the variable detail rides on `cause`) so crash-report grouping stays
+        // intact; we follow Node/undici's "fetch failed" spelling.
+        constexpr const char* FETCH_FAILED_MESSAGE = "fetch failed";
+
+        // Snapshot the JS call-site stack synchronously, inside fetch(), before SendAsync() hands
+        // the request to a worker thread. The transport-failure rejection is otherwise built in a
+        // continuation that runs after fetch() has returned, where an Error would capture zero user
+        // frames. We go through the global JS `Error` constructor (rather than napi_create_error) so
+        // engines that materialize `.stack` from the JS constructor path capture the live caller
+        // frames. The result is a plain std::string, safe to carry across the thread hop (unlike a
+        // Napi::Reference, which must be created and destroyed on the JS thread). Empty if the engine
+        // does not expose a stack at construction time (e.g. Chakra, which only populates `.stack`
+        // when an error is thrown) -- in that case the rejection simply carries no synthetic frames.
+        std::string CaptureCallSiteStack(Napi::Env env)
+        {
+            const Napi::Value errorCtor = env.Global().Get("Error");
+            if (!errorCtor.IsFunction())
+            {
+                return {};
+            }
+            const Napi::Object error = errorCtor.As<Napi::Function>().New({});
+            const Napi::Value stack = error.Get("stack");
+            return stack.IsString() ? stack.As<Napi::String>().Utf8Value() : std::string{};
+        }
+
+        // Reattach the synchronously-captured frames to the rejection's Error, replacing the captured
+        // header line (e.g. "Error\n    at ...") with one matching the TypeError we actually reject
+        // with, so the stack reads correctly while preserving the user's call site.
+        std::string ComposeRejectionStack(const std::string& capturedStack)
+        {
+            std::string header{"TypeError: "};
+            header += FETCH_FAILED_MESSAGE;
+
+            const auto firstNewline = capturedStack.find('\n');
+            if (firstNewline == std::string::npos)
+            {
+                return header;
+            }
+            return header + capturedStack.substr(firstNewline);
+        }
+
+        // Build the transport-failure rejection: a TypeError with a stable message, carrying the
+        // variable detail under `cause` (Node/undici shape) rather than as top-level own-properties.
+        // `code`/`detail` come from UrlLib's normalized accessors and may be empty on backends that
+        // do not yet populate them (Windows/Android today) -- in that case the standard observable
+        // shape (TypeError + stable message + url) is preserved and only the extra detail is absent.
+        Napi::Error BuildTransportError(Napi::Env env, const UrlLib::UrlRequest& request, const std::string& url, const std::string& capturedStack)
+        {
+            Napi::Error error = Napi::TypeError::New(env, FETCH_FAILED_MESSAGE);
+
+            Napi::Object cause = Napi::Object::New(env);
+            const std::string code{request.ErrorSymbol()};
+            const std::string detail{request.ErrorString()};
+            if (!code.empty())
+            {
+                cause.Set("code", Napi::String::New(env, code));
+            }
+            if (!detail.empty())
+            {
+                cause.Set("detail", Napi::String::New(env, detail));
+            }
+            cause.Set("url", Napi::String::New(env, url));
+            cause.Set("status", Napi::Number::New(env, static_cast<double>(static_cast<int>(request.StatusCode()))));
+            error.Set("cause", cause);
+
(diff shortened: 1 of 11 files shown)


Polyfills/Fetch/Source/Fetch.cpp
```diff
@@ -37,6 +37,80 @@ namespace Babylon::Polyfills::Internal
...
+        // Napi::Reference, which must be created and destroyed on the JS thread). Empty if the engine
+        // does not expose a stack at construction time (e.g. Chakra, which only populates `.stack`
+        // when an error is thrown) -- in that case the rejection simply carries no synthetic frames.
+        std::string CaptureCallSiteStack(Napi::Env env)
+        {
+            const Napi::Value errorCtor = env.Global().Get("Error");
+            if (!errorCtor.IsFunction())
+            {
+                return {};
+            }
```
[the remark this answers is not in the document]
bkaradzic-microsoft: Good catch — fixed in 1405c41. CaptureCallSiteStack now detects the global `Error` constructor via `IsUndefined()`/`IsNull()` (mirroring the Blob check below), so synchronous stack capture isn't silently disabled on JavaScriptCore/JSI builds that report constructors as `typeof 'object'`.

Polyfills/XMLHttpRequest/Readme.md
```diff
@@ -13,4 +13,18 @@ Unlike the web, XMLHttpRequest supports loading local files using two schemes:
...
 * Only `GET` requests are currently supported
-* For `readyState`, we only support `UNSENT`, `OPENED`, and `DONE`
\ No newline at end of file
+* For `readyState`, we only support `UNSENT`, `OPENED`, and `DONE`
+
+## Transport-error diagnostics (non-standard)
+A transport-level failure surfaces the standard way -- an `error` event followed by `loadend`,
+with `status === 0` -- exactly as on the web. In addition, two **non-standard, additive**
+read-only properties expose the normalized `UrlLib` transport-error detail so BN-aware code can
+tell a DNS failure from a refused connection or a missing local asset:
```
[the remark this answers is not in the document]
bkaradzic-microsoft: Fixed in 1405c41 — added the `error` event to the supported-events list in the Event listening section.

Core/AppRuntime/Source/AppRuntime_V8.cpp
```diff
@@ -61,6 +63,105 @@ namespace Babylon
...
+            tracker.flushScheduled = false;
+
+            v8::Isolate::Scope isolateScope{tracker.isolate};
+            v8::HandleScope handleScope{tracker.isolate};
+            for (auto& entry : tracker.unhandled)
+            {
+                const v8::Local<v8::Value> reason = entry.second.second.Get(tracker.isolate);
+                tracker.runtime->OnUnhandledPromiseRejection(ToError(env, JsValueFromV8LocalValue(reason)));
+            }
+            tracker.unhandled.clear();
```
bghgary: If a host handler ever rejects a promise synchronously, `OnPromiseReject` would insert into `tracker.unhandled` mid-iteration (rehash → invalidated iterator). Drain into a local first:

```suggestion
            auto unhandled = std::move(tracker.unhandled);
            tracker.unhandled.clear();
            for (auto& entry : unhandled)
            {
                const v8::Local<v8::Value> reason = entry.second.second.Get(tracker.isolate);
                tracker.runtime->OnUnhandledPromiseRejection(ToError(env, JsValueFromV8LocalValue(reason)));
            }
```

Core/AppRuntime/Source/AppRuntime_V8.cpp
```diff
@@ -61,6 +63,105 @@ namespace Babylon
...
+        {
+            V8RejectionTracker* tracker{t_rejectionTracker};
+            if (tracker == nullptr)
+            {
+                return;
+            }
+
+            v8::HandleScope handleScope{tracker->isolate};
+            const v8::Local<v8::Promise> promise{message.GetPromise()};
+            const int hash{promise->GetIdentityHash()};
```
bghgary: `GetIdentityHash()` isn't unique: two concurrently-unhandled promises can collide on `hash`, so the second overwrites the first's entry and a handler later added to either erases the shared slot — a still-unhandled rejection can be silently dropped. Low-probability on this diagnostics path, but the keying is lossy.

Core/AppRuntime/Include/Babylon/AppRuntime.h
```diff
@@ -21,6 +21,15 @@ namespace Babylon
             // Optional handler for unhandled exceptions.
             std::function<void(const Napi::Error&)> UnhandledExceptionHandler{DefaultUnhandledExceptionHandler};
 
+            // When true, unhandled promise rejections are routed to UnhandledExceptionHandler so
+            // the embedder's crash/telemetry pipeline can observe fire-and-forget failures (e.g. an
+            // un-awaited fetch() that rejects). Defaults to false to preserve existing behavior --
+            // when false, no per-engine rejection tracker is installed. Reporting is deferred to the
+            // end of the current turn, so a rejection that is handled synchronously (e.g.
+            // `const p = Promise.reject(e); p.catch(...)`) does not reach the handler. Implemented for
+            // the Chakra and V8 engines.
```
bghgary: Drop this flag and always track — route unhandled rejections to the existing `UnhandledExceptionHandler` (defaults to the benign logger), matching browser behavior. Document coverage once here, accurately: fires on V8 and JSC; a no-op on Chakra (in-box) and JSI — neither exposes a host rejection hook (`JsSetHostPromiseRejectionTracker` is ChakraCore-only; neither `jsi::Runtime` nor V8JSI exposes the V8 callback). The current "Chakra and V8" is wrong.

Core/AppRuntime/Source/AppRuntime_Chakra.cpp
```diff
@@ -50,6 +50,12 @@ namespace Babylon
                 });
             },
             &dispatchFunction));
+
+        // Options::EnableUnhandledPromiseRejectionTracking is not honored on this backend: the OS
+        // EdgeMode Chakra runtime (chakrart.h) does not expose a host promise-rejection tracker
+        // (JsSetHostPromiseRejectionTracker is ChakraCore-only), so there is no hook to route
+        // unhandled rejections from. The option is implemented for the V8 backend.
```
bghgary: Document coverage once in the header rather than a lone per-backend note here. With JSC added, V8 and JSC have it; Chakra and JSI both *can't* — Chakra has no jsrt host hook, and neither `jsi::Runtime` nor V8JSI exposes one (would require forking V8JSI). A single Chakra note shouldn't imply the others are covered.

Core/AppRuntime/Source/AppRuntime_V8.cpp
```diff
@@ -61,6 +63,105 @@ namespace Babylon
...
 
         std::unique_ptr<Module> Module::s_module;
+
+        // Tracks promises rejected without a handler so they can be reported once the current turn
+        // settles. V8 fires kPromiseRejectWithNoHandler when a promise is rejected with no handler
+        // and kPromiseHandlerAddedAfterReject if one is attached afterwards; reporting is deferred
+        // (via AppRuntime::Dispatch) so a synchronous `Promise.reject(e); ...; p.catch(...)` is
+        // removed before it is ever reported. Keyed by promise identity hash; the promise and reason
+        // are held in v8::Global handles so they survive until the deferred flush runs.
+        struct V8RejectionTracker
```
bghgary: This PR should include the JSC implementation, not defer it. JSC exposes `JSGlobalContextSetUnhandledRejectionCallback` (macOS 10.15.4+/iOS 13.4+) and CI already runs the JSC matrix, so there's no validation barrier. The candidate map / deferred-flush / `ToError` bookkeeping here is engine-agnostic — factor it out so the JSC backend is a thin adapter, not a second copy. (Chakra and JSI stay no-ops: neither exposes a host rejection hook.)

Tests/UnitTests/Shared/Shared.cpp
```diff
@@ -279,6 +280,72 @@ TEST(AppRuntime, DestroyDoesNotDeadlock)
...
 
+TEST(AppRuntime, UnhandledPromiseRejectionReachesHandler)
+{
+    // EnableUnhandledPromiseRejectionTracking is only implemented on the V8 backend so far (the OS
+    // EdgeMode Chakra runtime exposes no host promise-rejection hook; JavaScriptCore/JSI are
+    // follow-ups). Skip elsewhere so the test doesn't hang waiting for a report that never comes.
+#if defined(JSRUNTIMEHOST_NAPI_ENGINE)
+    if (std::string_view{JSRUNTIMEHOST_NAPI_ENGINE} != "V8")
+    {
+        GTEST_SKIP() << "unhandled promise rejection tracking is only implemented for the V8 backend";
```
bghgary: This runtime string-compare is inconsistent with the JSI gate below (`#if !defined(JSRUNTIMEHOST_NAPI_ENGINE_JSI)`, lines 353/423). Converge on that compile-time boolean pattern: have CMake emit `JSRUNTIMEHOST_NAPI_ENGINE_${NAPI_JAVASCRIPT_ENGINE}` (parallel to `_JSI`) and gate with `#if defined(JSRUNTIMEHOST_NAPI_ENGINE_V8)`, dropping the stringly-typed `JSRUNTIMEHOST_NAPI_ENGINE="..."` macro and the `std::string_view` compare (a `"v8"`/`"V8"` typo silently misbehaves). The JSI gate is necessarily compile-time, so that's the direction. The larger Shared.cpp split into per-area/per-engine files is a separate follow-up.

bghgary (changes requested):

Comments inline. Main asks: implement JSC in this PR (it has the API, CI covers it); drop the opt-in flag — always track, routing to the existing `UnhandledExceptionHandler`; and make the two test-gating mechanisms consistent (compile-time `#if defined`, not a runtime engine-name string compare). Plus two minor V8-tracker robustness notes.

Core/AppRuntime/Include/Babylon/AppRuntime.h
bkaradzic-microsoft: Done in 97e2c68. Removed `Options::EnableUnhandledPromiseRejectionTracking`; unhandled rejections are now always routed to `UnhandledExceptionHandler` (defaults to the benign logger), matching browser behavior. Coverage is documented once on `OnUnhandledPromiseRejection` in `AppRuntime.h`: V8 and JavaScriptCore supported; Chakra (in-box/EdgeMode) and JSI no-op. Validated that `JavaScript.All` stays green (203 passing) on V8 with always-on tracking, i.e. the existing JS suite leaves no stray rejections.

Core/AppRuntime/Source/AppRuntime_V8.cpp
bkaradzic-microsoft: Fixed in 97e2c68. Candidates are now held in a `std::vector` and matched by promise object identity rather than a hash key: `kPromiseHandlerAddedAfterReject` removes via `candidate.promise.Get(isolate) == promise`, so colliding `GetIdentityHash()` values can no longer overwrite/drop a still-unhandled rejection. `GetIdentityHash()` is gone from this path.

Core/AppRuntime/Source/AppRuntime_V8.cpp
bkaradzic-microsoft: Done in 97e2c68. JSC is now implemented in this PR. `AppRuntime_JavaScriptCore.cpp` registers a callback via `JSGlobalContextSetUnhandledRejectionCallback` (forward-declared SPI from `JSContextRefPrivate.h`, guarded with `__builtin_available(macOS 10.15.4, iOS 13.4, *)`). The engine-agnostic bookkeeping (`ToError` + the candidate/deferred-flush logic) is factored into `AppRuntime_PromiseRejection.h` as `PromiseRejectionTracker<CandidateT>`; V8 instantiates it with a `V8RejectionCandidate`. JSC needs no candidate map — its hook fires at the microtask checkpoint only for still-unhandled rejections — so its adapter is a thin direct report. I'm watching the macOS/JSC matrix; if the private symbol isn't in the CI SDK headers the forward declaration should still link against the framework export, and I'll iterate on that job if needed.

Core/AppRuntime/Source/AppRuntime_Chakra.cpp
bkaradzic-microsoft: Done in 97e2c68. Trimmed this to a one-line note that defers to `AppRuntime.h` for the full coverage matrix, so it no longer implies the other backends are covered.

Core/AppRuntime/Source/AppRuntime_V8.cpp
bkaradzic-microsoft: Fixed in 97e2c68. The deferred flush now moves the candidate list into a local before reporting (`const std::vector<CandidateT> candidates = std::move(m_candidates); m_candidates.clear();` in `PromiseRejectionTracker::Flush`), so a host handler that synchronously rejects another promise re-enters `Add()` against the fresh, empty member vector and can no longer invalidate the in-progress iteration. The bookkeeping now lives in the shared `AppRuntime_PromiseRejection.h`.

Tests/UnitTests/Shared/Shared.cpp
bkaradzic-microsoft: Done in 97e2c68. CMake now emits `JSRUNTIMEHOST_NAPI_ENGINE_${NAPI_JAVASCRIPT_ENGINE}` (parallel to `_JSI`) for both the desktop and Android UnitTests targets, and the rejection tests gate with `#if !defined(JSRUNTIMEHOST_NAPI_ENGINE_V8) && !defined(JSRUNTIMEHOST_NAPI_ENGINE_JavaScriptCore)` — the `std::string_view` compare and the stringly-typed `JSRUNTIMEHOST_NAPI_ENGINE="..."` reliance are gone from these tests. Agreed the broader per-area/per-engine `Shared.cpp` split is a separate follow-up.

bkaradzic-microsoft:
@bghgary Thanks for the thorough review — all of it is addressed in 97e2c68 (hop 3 refactor):

- **JSC implemented in this PR** via `JSGlobalContextSetUnhandledRejectionCallback`; engine-agnostic bookkeeping factored into `AppRuntime_PromiseRejection.h` (`PromiseRejectionTracker<CandidateT>` + `ToError`), so JSC is a thin adapter and V8 instantiates the shared tracker.
- **Flag dropped** — unhandled rejections always route to `UnhandledExceptionHandler`; coverage documented once on `AppRuntime.h` (V8 + JSC supported; Chakra/JSI no-op).
- **V8 iterator invalidation** fixed (drain candidates via `std::move` before reporting).
- **V8 `GetIdentityHash` collision** fixed (vector + promise-identity match, hash key removed).
- **Compile-time test gate** — CMake emits `JSRUNTIMEHOST_NAPI_ENGINE_${ENGINE}`; tests gate on `#if defined(...)`, no more `std::string_view` compare.
- **Chakra note** trimmed to defer to the header.

Validated locally on Win32: V8 Release — both `AppRuntime` rejection tests pass and `JavaScript.All` stays green (203 passing) with always-on tracking; Chakra Debug — compiles and the rejection tests skip via the gate. JSC validates on the macOS CI matrix; watching that job (the rejection callback uses an SPI from `JSContextRefPrivate.h`, forward-declared and `__builtin_available`-guarded — I'll iterate if the CI SDK doesn't link it).

Core/AppRuntime/Source/AppRuntime_V8.cpp
bkaradzic-microsoft: CI is fully green now. The Apple JSC matrix (macOS/iOS) validates the adapter; the first run surfaced one real wrinkle on the **non-Apple** JSC builds — `JSGlobalContextSetUnhandledRejectionCallback` is an Apple SPI absent from WebKitGTK, and `__builtin_available` is Clang/Apple-only, so the Linux (WebKitGTK/gcc) and Android JSC jobs failed to compile. Fixed in b2553a2 by guarding the rejection machinery with `#if __APPLE__` (consistent with the existing `JSGlobalContextSetInspectable` guard) and tightening the test gate to `V8 || (JavaScriptCore && __APPLE__)`. WebKitGTK/Android JSC are now documented no-ops. All jobs pass on b2553a2.

Core/AppRuntime/Include/Babylon/AppRuntime.h
```diff
@@ -45,6 +45,24 @@ namespace Babylon
...
 
+        // Routes an unhandled promise rejection to the embedder's UnhandledExceptionHandler (which
+        // defaults to a benign logger), so an embedder's crash/telemetry pipeline can observe
+        // fire-and-forget failures (e.g. an un-awaited fetch() that rejects) -- matching the browser
+        // `unhandledrejection` behavior. Reporting is deferred to the end of the turn, so a rejection
+        // handled synchronously (e.g. `const p = Promise.reject(e); p.catch(...)`) is not reported.
+        //
+        // Coverage is determined by whether the engine exposes a host promise-rejection hook:
+        //   * V8 (Isolate::SetPromiseRejectCallback) -- supported on all platforms.
+        //   * Apple JavaScriptCore (JSGlobalContextSetUnhandledRejectionCallback) -- supported. This
```
bghgary, line 56: The rejection hook exists on V8 and Apple JSC, but not on in-box Chakra, JSI, or non-Apple (WebKitGTK/Android) JSC — none expose one, and there's no JS/Node-API way to detect a dropped rejection without it, so it can't be closed later. That makes the same script produce a different observable runtime effect depending on the engine.

Engine-specific tooling (debugger/inspector) doesn't change program behavior, and gaps like the V8JSI `DataView` TODO get filled — this would be the first time the runtime *intentionally and permanently* differs by engine, with a silent failure mode: a fire-and-forget rejection caught in a V8 test goes unreported once shipped on Chakra or Android. I'd rather not take that on for a best-effort diagnostic.

bghgary (changes requested):

The fetch/XHR error-fidelity and AbortSignal changes look good. My one concern is the unhandled-promise-rejection tracking (inline) — could you split that into its own PR? I'll approve the rest as-is.

Core/AppRuntime/Include/Babylon/AppRuntime.h
bkaradzic-microsoft: Agreed — and I've already pulled this out of #200 into #204, so the fetch/XHR + AbortSignal work here isn't blocked on it. (This comment is on the doc block that moved.)

You've convinced me on the substance. This is qualitatively different from the debugger/inspector (which don't change program behavior) and from the V8JSI `DataView` TODO (which can eventually be filled): there's no host hook on in-box Chakra, JSI, or WebKitGTK/Android JSC, and no Node-API/JS way to observe a dropped rejection without one — so it genuinely *can't* be closed later. That makes it a permanent, intentional, silent per-engine behavioral difference, which is a fair line to hold for a best-effort diagnostic.

Two ways forward on #204, and I lean toward the first:

1. **Drop it** — accept that fire-and-forget rejection reporting isn't something JsRuntimeHost offers uniformly, and leave embedders to `.catch()` / their own handling.
2. **Keep it only if the divergence is explicit and non-silent** — gated behind an opt-in, and requesting it on an unsupported engine fails fast / logs that coverage is unavailable rather than silently doing nothing, so nobody ships a V8-tested telemetry hook assuming Chakra/Android parity.

Since even an explicit opt-in still leaves the same script observably different by engine — which is exactly what you're pushing back on — I'm fine going with (1) unless you think (2) earns its keep. Let's settle it on #204.

bghgary (approved):

LGTM

PR #201, opened 2026-06-16
AppRuntime: route unhandled promise rejections to the host handler (V8) — hop 3 of #196

bkaradzic-microsoft:
Hop 3 of #196 — make fire-and-forget promise rejections reach the embedding app instead of vanishing.

### Problem

`AppRuntime::Dispatch` only catches **synchronous** `Napi::Error` throws from dispatched callbacks, and no engine had a promise-rejection tracker wired anywhere in Core. So a fire-and-forget `fetch()` that rejects (or a throw inside any `.then` with no `.catch`) never reached the embedder's `UnhandledExceptionHandler` — the Sentry/Bugsnag hook — and the process exited `0`.

### Approach

- **`AppRuntime::Options::EnableUnhandledPromiseRejectionTracking`** (opt-in, default `false`). When `false`, no tracker is installed and behavior is unchanged. When `true`, unhandled rejections route into the existing `UnhandledExceptionHandler`.
- **`AppRuntime::OnUnhandledPromiseRejection(napi_value)`** wraps the rejection reason as a `Napi::Error` (forwarding `Error`-like reasons as-is to preserve message/stack/cause; stringifying non-objects) and invokes the handler.
- **Deferred, Node-like semantics**: reporting is deferred to the end of the turn via `Dispatch`, so a rejection that is **handled synchronously** in the same turn (`const p = Promise.reject(e); p.catch(...)`) is dropped before it is ever reported — no false positives.

### Engine support

| Engine | Status |
| ------ | ------ |
| **V8** | ✅ Implemented + tested (`Isolate::SetPromiseRejectCallback`; promise/reason held in `v8::Global` across the deferred flush). |
| **Chakra** | ⚠️ Documented no-op — the OS **EdgeMode** runtime (`chakrart.h`) exposes no host promise-rejection tracker (`JsSetHostPromiseRejectionTracker` is **ChakraCore-only**, and JsRuntimeHost builds against EdgeMode via `USE_EDGEMODE_JSRT`). |
| **JavaScriptCore / JSI** | ⏳ Follow-ups — `JSGlobalContextSetUnhandledRejectionCallback` (JSC) and the JSI tracker. Happy to add in this PR or separately; would value guidance on the JSI/V8JSI path. |

### Validation

Native tests in `Shared.cpp` (skipped on the Chakra backend, which can't support the feature):
- `UnhandledPromiseRejectionReachesHandler` — a fire-and-forget `Promise.reject(new Error('boom'))` reaches the host handler with the right message.
- `SynchronouslyHandledRejectionDoesNotReachHandler` — a synchronously-handled rejection does **not** reach the handler (verifies the deferral).

Built and ran on **Win32 / V8** (both pass) and **Win32 / Chakra** (framework compiles, both tests skip as expected). The opt-in default keeps the existing `JavaScript, All` suite unaffected.

### Notes for reviewers

- `OnUnhandledPromiseRejection` is exposed as a public `AppRuntime` method (documented "internal, engine-implementation use") because the V8 rejection callback is a bare function pointer whose helpers are free functions and can't reach a private member.
- The V8 backend reuses the internal `reinterpret_cast<napi_value>(*local)` bridge (mirrored from `js_native_api_v8.h`, which isn't on the public include path) with the same `static_assert`.

Part of the #196 chain (hop 2 is #200).

### DIFF
--- a/Core/AppRuntime/Include/Babylon/AppRuntime.h
+++ b/Core/AppRuntime/Include/Babylon/AppRuntime.h
@@ -56,9 +56,9 @@ namespace Babylon
 
         // Routes an unhandled promise rejection to the embedder's UnhandledExceptionHandler. Called
         // by the per-engine promise-rejection tracker installed in RunEnvironmentTier when
-        // Options::EnableUnhandledPromiseRejectionTracking is set. `reason` is the rejection value as
-        // a napi_value. Intended for internal (engine-implementation) use.
-        void OnUnhandledPromiseRejection(napi_value reason);
+        // Options::EnableUnhandledPromiseRejectionTracking is set. Intended for internal
+        // (engine-implementation) use.
+        void OnUnhandledPromiseRejection(const Napi::Error& error);
 
         // Default unhandled exception handler that outputs the error message to the program output.
         static void BABYLON_API DefaultUnhandledExceptionHandler(const Napi::Error& error);
--- a/Core/AppRuntime/Include/Babylon/AppRuntime.h
+++ b/Core/AppRuntime/Include/Babylon/AppRuntime.h
@@ -21,6 +21,15 @@ namespace Babylon
             // Optional handler for unhandled exceptions.
             std::function<void(const Napi::Error&)> UnhandledExceptionHandler{DefaultUnhandledExceptionHandler};
 
+            // When true, unhandled promise rejections are routed to UnhandledExceptionHandler so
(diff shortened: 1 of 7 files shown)


bkaradzic-microsoft:
Consolidated into #200 — the whole #196 chain is now a single PR (hop 3 is the second commit there). Closing to avoid three PRs for one feature in one repo. The unhandled-rejection work, all CI fixes (V8-gating, the JSI napi.h build fix, and the Android JNI engine define), and the native tests are preserved in #200.

PR #202, opened 2026-06-16
fetch/AbortSignal: honor init.signal + modernize AbortSignal — hop 4 of #196

bkaradzic-microsoft:
Hop 4 of #196 — make `AbortController` / `init.signal` actually work in `fetch()`, and modernize the `AbortSignal` polyfill.

### Problem

`fetch()` passed `arcana::cancellation::none()` for both continuations and never looked at `init.signal`, so abort never rejected (`AbortError`) and never cancelled the transport. The `AbortSignal` polyfill also predated the modern spec — no `reason`, no `throwIfAborted()`, a writable `aborted`, and no static `AbortSignal.abort()`. Combined with `AbortController` being installed globally in the Playground (BabylonJS/BabylonNative#1708), library feature-detection passed while signals silently no-op'd — making user cancellations indistinguishable from network failures in telemetry.

### AbortSignal

- `reason` (read-only) and `throwIfAborted()`.
- static `AbortSignal.abort(reason?)`.
- `aborted` is now read-only (setter removed).
- `Abort(reason)` records the reason, defaulting to an **AbortError** — an `Error` whose `name` is `"AbortError"` (there is no `DOMException` polyfill, so this matches what web code checks: `err.name === "AbortError"`).
- `AbortController.abort(reason?)` forwards the reason to the signal.

### fetch

- Honors `init.signal` through the signal's **JS interface** (`aborted` / `reason` / `add`/`removeEventListener`), so `fetch` stays decoupled from the `AbortController` polyfill's C++ types.
- An **already-aborted** signal rejects the promise synchronously with the signal's `reason`, never touching the transport.
- Otherwise an `"abort"` listener cancels the in-flight transport via `UrlRequest::Abort()`, and the completion continuation rejects with the signal's `reason` (an `AbortError`) instead of a network-error `TypeError`. The listener is torn down when the request settles (breaking the listener ⟷ state ownership cycle).

### Validation

Built and ran the UnitTests suite on **Win32 / Chakra** — all pass (199), including new tests:
- `AbortSignal.abort()` returns an aborted signal whose `reason` is an `AbortError`; `throwIfAborted()` throws only once aborted; `abort(reason)` records the reason; `abort()` defaults to an `AbortError`.
- `fetch` rejects with an `AbortError` both when the signal is **already aborted** and when aborted **in-flight** (the in-flight case cancels the WinRT transport and settles in ~50 ms).

Win32's UrlLib backend honors `UrlRequest::Abort()`, so the end-to-end abort path is exercised here.

### Scope / follow-ups

- **`UrlRequest::Abort()` on curl/NSURLSession**: today it only cancels the Windows backend (the curl/NSURLSession backends never observe `m_cancellationSource`). Making abort interrupt those transports is a **UrlLib** change — until it lands, in-flight abort on Linux/Apple rejects only once the transport gives up on its own. Tracked in #196.
- `AbortSignal.timeout()` is left as a follow-up.

Part of the #196 chain (hops 2 and 3 are #200 and #201).

### DIFF
--- a/Polyfills/AbortController/Readme.md
+++ b/Polyfills/AbortController/Readme.md
@@ -1,10 +1,21 @@
 # AbortController
 Implements parts of [`AbortController`](https://developer.mozilla.org/en-US/docs/Web/API/AbortController/) and [`AbortSignal`](https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal). Provides a way to trigger the abort signal. *Work In Progress*
 
+Supported on `AbortSignal`:
+* `aborted` (read-only) and `reason`
+* [`throwIfAborted()`](https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal/throwIfAborted)
+* static [`AbortSignal.abort(reason?)`](https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal/abort)
+* `onabort`, `addEventListener("abort", ...)`, `removeEventListener`
+
+`AbortController.abort(reason?)` forwards the reason to the signal; when no reason is given the
+signal's `reason` defaults to an `AbortError` (an `Error` whose `name` is `"AbortError"`, since
+there is no `DOMException` polyfill). `fetch()` honors an `AbortSignal` passed via `init.signal`:
+an already-aborted signal rejects the promise synchronously, and an in-flight abort cancels the
+transport and rejects with the signal's `reason`. (Transport cancellation is effective on backends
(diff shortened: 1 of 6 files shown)


bkaradzic-microsoft:
Consolidated into #200 — the whole #196 chain is now a single PR (hop 4 is the last commit there). Closing to avoid three PRs for one feature in one repo, and because the AbortSignal/fetch changes share Fetch.cpp with hop 2 (now integrated cleanly in #200).

Issue #194, opened 2026-06-09
Node-API-JSC: napi_typeof reports napi_object for JSObjectMakeConstructor constructors


bkaradzic-microsoft:
Follow-up from the native `fetch()` polyfill review (#188).

`napi_typeof` in `js_native_api_javascriptcore.cc` decides function-ness solely via `JSObjectIsFunction`, which returns `false` for constructors created with `JSObjectMakeConstructor` on some JSC builds (e.g. libjavascriptcoregtk). As a result it reports `napi_object` where V8 returns `napi_function`, so Napi `IsFunction()` incorrectly rejects otherwise-valid constructors (e.g. the `Blob` polyfill).

Both the **Fetch** polyfill (#188) and the **File** polyfill currently carry an identical `IsUndefined()`/`IsNull()` workaround to detect the constructor without `IsFunction()`.

A fix in `napi_typeof` (also consulting `JSObjectIsConstructor` in addition to `JSObjectIsFunction`) would let both polyfills drop the workaround rather than spreading per-polyfill copies.

bkaradzic-microsoft:
Fixed by #207 (merged 2026-07-28).

That PR changed `napi_typeof` in `Core/Node-API/Source/js_native_api_javascriptcore.cc` to consult `JSObjectIsConstructor` in addition to `JSObjectIsFunction`, so constructors created via `JSObjectMakeConstructor` (i.e. node-addon-api `DefineClass` constructors) now report `napi_function` on JSC builds such as libjavascriptcoregtk, matching V8.

With that in place the `IsUndefined()`/`IsNull()` workarounds carried by the Fetch (#188) and File polyfills were removed in the same PR.

Closing as completed.

PR #207, opened 2026-07-22
Add blob: URL support to URL.createObjectURL / revokeObjectURL

bkaradzic-microsoft:
### What

Adds `blob:` URL support to `URL.createObjectURL` / `URL.revokeObjectURL`, and makes the minted URLs resolvable by every consumer of `UrlRequest` (`fetch`, `XMLHttpRequest`, and any downstream user of UrlLib).

Previously `createObjectURL` had no backing store on Native, so `fetch`/XHR could not load a Blob and `revokeObjectURL` had nothing to release. This implements the missing in-memory registry.

### How

- **URL polyfill** — `createObjectURL(blob)` puts the Blob's bytes into an in-memory store keyed by a freshly minted `blob:null/<uuidv4>` URL and returns it; `revokeObjectURL` frees the entry. The store's C++ surface is deliberately not exported: `RegisterObjectURL`/`RevokeObjectURL` are file-local to `URL.cpp`, since neither has a caller elsewhere and #215 would reshape the contract anyway.
- **`blob:` scheme resolver** — rather than special-casing `blob:` inside each request polyfill, the URL polyfill registers a resolver with `UrlLib::UrlRequest::RegisterSchemeResolver` (added in BabylonJS/UrlLib#37). Every `UrlRequest` consumer therefore resolves `blob:` uniformly, with no per-polyfill interception: a hit yields status 200 with the stored `content-type`, and a miss (or a revoked URL) yields a network error, so `fetch` rejects with a `TypeError` and XHR reports status 0 plus an `error` event.
- **Blob polyfill** — adds a `TryGetData(object) -> std::optional<BlobData>` accessor so the URL polyfill can read a Blob's bytes and MIME type. It lives behind a `BlobInternal` INTERFACE target (`InternalInclude/`, mirroring `JsRuntimeInternal`) rather than the public header, since it is an in-repo seam and not an embedder contract. `Blob` now holds its bytes in a `std::shared_ptr<const std::vector<std::byte>>`, so `createObjectURL` shares the buffer with the store instead of duplicating it, and the returned data no longer has a lifetime tied to the JS object.
- **JavaScriptCore `napi_typeof` fix** — `Core/Node-API/Source/js_native_api_javascriptcore.cc` now consults `JSObjectIsConstructor` in addition to `JSObjectIsFunction`, so constructors made with `JSObjectMakeConstructor` report `napi_function` rather than `napi_object`. This is not specific to this feature — it affects every JSC consumer — but it lets `Fetch` and `File` drop their `IsUndefined()/IsNull()` workarounds and go back to a plain `IsFunction()` check.
- Adds `createObjectURL` round-trip unit tests (fetch + XHR, revoke frees memory).

### Design note

The store is **process-global** rather than per-environment. Per-env teardown would normally use `napi_add_env_cleanup_hook`, but no Node-API engine adapter in this repo implements it, so there is no portable per-env hook available today. The keys are unguessable v4 UUIDs, so sharing one store across environments is safe — a `blob:` URL minted in one environment is never produced in another.

The cost is that entries which are never revoked stay alive for the lifetime of the process, and are not released when the environment that created them is torn down. This is a real leak, not a browser-like behavior: a browser drops a document's blob URLs when the document unloads. #215 tracks implementing `napi_add_env_cleanup_hook` across the adapters so the store can become per-environment.

### Testing

Full `UnitTests` suite green (216 passing) on Windows/Chakra locally, with the remaining engines and platforms covered by CI, including the new `URL.createObjectURL` round-trip tests (fetch + XHR text/binary, `content-type` header, and revoke → network-error / status 0).

### DIFF
--- a/Polyfills/Blob/Include/Babylon/Polyfills/Blob.h
+++ b/Polyfills/Blob/Include/Babylon/Polyfills/Blob.h
@@ -3,7 +3,15 @@
 #include <napi/env.h>
 #include <Babylon/Api.h>
 
+#include <cstddef>
+#include <string>
+
 namespace Babylon::Polyfills::Blob
 {
     void BABYLON_API Initialize(Napi::Env env);
+
+    // Synchronously reads the bytes and MIME type of a Blob JS object created by this
+    // polyfill. Returns false if `object` is not a Blob. The returned pointer remains
+    // valid only while `object` is alive, so it must be consumed synchronously.
+    bool BABYLON_API TryGetData(const Napi::Object& object, const std::byte*& outData, size_t& outSize, std::string& outType);
 }
--- a/Polyfills/Blob/Include/Babylon/Polyfills/Blob.h
+++ b/Polyfills/Blob/Include/Babylon/Polyfills/Blob.h
@@ -3,18 +3,7 @@
 #include <napi/env.h>
 #include <Babylon/Api.h>
 
-#include <cstddef>
-#include <memory>
-#include <string>
-#include <vector>
-
 namespace Babylon::Polyfills::Blob
 {
     void BABYLON_API Initialize(Napi::Env env);
-
-    // Synchronously reads the bytes and MIME type of a Blob JS object created by this polyfill.
-    // Returns false if `object` is not a Blob. A Blob is immutable once constructed, so `outData`
-    // is a shared reference to its buffer rather than a copy, and it stays valid for as long as the
-    // caller holds it -- even if `object` is collected.
-    bool BABYLON_API TryGetData(const Napi::Object& object, std::shared_ptr<const std::vector<std::byte>>& outData, std::string& outType);
 }
--- a/Polyfills/Blob/Include/Babylon/Polyfills/Blob.h
+++ b/Polyfills/Blob/Include/Babylon/Polyfills/Blob.h
@@ -4,14 +4,17 @@
 #include <Babylon/Api.h>
 
 #include <cstddef>
+#include <memory>
 #include <string>
+#include <vector>
 
 namespace Babylon::Polyfills::Blob
 {
     void BABYLON_API Initialize(Napi::Env env);
 
-    // Synchronously reads the bytes and MIME type of a Blob JS object created by this
-    // polyfill. Returns false if `object` is not a Blob. The returned pointer remains
-    // valid only while `object` is alive, so it must be consumed synchronously.
-    bool BABYLON_API TryGetData(const Napi::Object& object, const std::byte*& outData, size_t& outSize, std::string& outType);
+    // Synchronously reads the bytes and MIME type of a Blob JS object created by this polyfill.
+    // Returns false if `object` is not a Blob. A Blob is immutable once constructed, so `outData`
+    // is a shared reference to its buffer rather than a copy, and it stays valid for as long as the
+    // caller holds it -- even if `object` is collected.
+    bool BABYLON_API TryGetData(const Napi::Object& object, std::shared_ptr<const std::vector<std::byte>>& outData, std::string& outType);
 }
--- a/Polyfills/Blob/Source/Blob.cpp
+++ b/Polyfills/Blob/Source/Blob.cpp
@@ -142,8 +142,63 @@ namespace Babylon::Polyfills::Blob
 
     bool BABYLON_API TryGetData(const Napi::Object& object, const std::byte*& outData, size_t& outSize, std::string& outType)
     {
-        const auto blobConstructor = object.Env().Global().Get("Blob");
-        if (!blobConstructor.IsFunction() || !object.InstanceOf(blobConstructor.As<Napi::Function>()))
+        const auto env = object.Env();
+        const auto global = env.Global();
+
+        // Verify `object` is a Blob (or a subclass such as File) by walking its prototype chain and
+        // comparing each link against Blob.prototype, using the JS-level Object.getPrototypeOf.
+        //
+        // Two engine-adapter quirks drive this implementation:
+        //  * The Blob constructor is created via node-addon-api's DefineClass. On the JavaScriptCore
+        //    adapter such constructors are callable-as-constructor but are NOT reported as functions
+        //    by napi_typeof (JSObjectIsFunction returns false), so we must probe the constructor with
+        //    IsObject() rather than IsFunction() and read `prototype` as a plain object property.
+        //  * We use Object.getPrototypeOf rather than the raw napi_get_prototype C API because the
+        //    latter is not exposed by every adapter (e.g. JSI) and, on JSC, returns the raw
+        //    [[Prototype]] which differs from the JS-visible prototype of a DefineClass instance.
+        // Together this keeps the check portable across QuickJS, V8, JavaScriptCore, Chakra and JSI,
+        // and it also accepts Blob subclasses (e.g. File). We deliberately avoid napi_instanceof,
+        // whose node-addon-api wrapper requires a Napi::Function and is likewise gated on the
+        // constructor being typed as a function.
+        const auto blobConstructor = global.Get("Blob");
+        if (!blobConstructor.IsObject())
+        {
+            return false;
+        }
+
+        const auto blobPrototype = blobConstructor.As<Napi::Object>().Get("prototype");
+        if (!blobPrototype.IsObject())
+        {
+            return false;
+        }
+
+        const auto objectConstructor = global.Get("Object");
+        if (!objectConstructor.IsObject())
+        {
+            return false;
+        }
+
+        const auto getPrototypeOf = objectConstructor.As<Napi::Object>().Get("getPrototypeOf");
+        if (!getPrototypeOf.IsFunction())
+        {
+            return false;
+        }
+
+        const auto getPrototypeOfFn = getPrototypeOf.As<Napi::Function>();
+
+        bool isBlob = false;
+        Napi::Value current = getPrototypeOfFn.Call({object});
+        while (current.IsObject())
+        {
+            if (current.StrictEquals(blobPrototype))
+            {
+                isBlob = true;
+                break;
+            }
+            current = getPrototypeOfFn.Call({current});
+        }
+
+        if (!isBlob)
         {
             return false;
         }
--- a/Polyfills/Blob/Source/Blob.cpp
+++ b/Polyfills/Blob/Source/Blob.cpp
@@ -148,20 +148,14 @@ namespace Babylon::Polyfills::Blob
         // Verify `object` is a Blob (or a subclass such as File) by walking its prototype chain and
         // comparing each link against Blob.prototype, using the JS-level Object.getPrototypeOf.
         //
-        // Two engine-adapter quirks drive this implementation:
-        //  * The Blob constructor is created via node-addon-api's DefineClass. On the JavaScriptCore
-        //    adapter such constructors are callable-as-constructor but are NOT reported as functions
-        //    by napi_typeof (JSObjectIsFunction returns false), so we must probe the constructor with
-        //    IsObject() rather than IsFunction() and read `prototype` as a plain object property.
-        //  * We use Object.getPrototypeOf rather than the raw napi_get_prototype C API because the
-        //    latter is not exposed by every adapter (e.g. JSI) and, on JSC, returns the raw
-        //    [[Prototype]] which differs from the JS-visible prototype of a DefineClass instance.
-        // Together this keeps the check portable across QuickJS, V8, JavaScriptCore, Chakra and JSI,
-        // and it also accepts Blob subclasses (e.g. File). We deliberately avoid napi_instanceof,
-        // whose node-addon-api wrapper requires a Napi::Function and is likewise gated on the
-        // constructor being typed as a function.
+        // We use Object.getPrototypeOf rather than the raw napi_get_prototype C API because the
+        // latter is not exposed by every adapter (e.g. JSI) and, on JSC, returns the raw
+        // [[Prototype]] which differs from the JS-visible prototype of a DefineClass instance.
+        // This keeps the check portable across QuickJS, V8, JavaScriptCore, Chakra and JSI, and it
+        // also accepts Blob subclasses (e.g. File). We deliberately avoid napi_instanceof, whose
+        // node-addon-api wrapper requires a Napi::Function.
         const auto blobConstructor = global.Get("Blob");
-        if (!blobConstructor.IsObject())
+        if (!blobConstructor.IsFunction())
         {
             return false;
         }
--- a/Polyfills/Blob/Source/Blob.cpp
+++ b/Polyfills/Blob/Source/Blob.cpp
@@ -139,4 +139,20 @@ namespace Babylon::Polyfills::Blob
     {
         Internal::Blob::Initialize(env);
     }
+
+    bool BABYLON_API TryGetData(const Napi::Object& object, const std::byte*& outData, size_t& outSize, std::string& outType)
+    {
+        const auto blobConstructor = object.Env().Global().Get("Blob");
+        if (!blobConstructor.IsFunction() || !object.InstanceOf(blobConstructor.As<Napi::Function>()))
+        {
+            return false;
+        }
+
+        const auto* blob = Internal::Blob::Unwrap(object);
+        const auto& data = blob->Data();
+        outData = data.data();
+        outSize = data.size();
+        outType = blob->Type();
+        return true;
+    }
 }
--- a/Polyfills/Blob/Source/Blob.cpp
+++ b/Polyfills/Blob/Source/Blob.cpp
@@ -56,7 +56,7 @@ namespace Babylon::Polyfills::Internal
 
     Napi::Value Blob::GetSize(const Napi::CallbackInfo&)
     {
-        return Napi::Value::From(Env(), m_data.size());
+        return Napi::Value::From(Env(), m_data->size());
     }
 
     Napi::Value Blob::GetType(const Napi::CallbackInfo&)
@@ -67,8 +67,8 @@ namespace Babylon::Polyfills::Internal
     Napi::Value Blob::Text(const Napi::CallbackInfo&)
     {
         // NOTE: This will not check for UTF-8 validity
-        const auto begin = reinterpret_cast<const char*>(m_data.data());
-        std::string text(begin, m_data.size());
+        const auto begin = reinterpret_cast<const char*>(m_data->data());
+        std::string text(begin, m_data->size());
 
         const auto deferred = Napi::Promise::Deferred::New(Env());
         deferred.Resolve(Napi::String::New(Env(), text));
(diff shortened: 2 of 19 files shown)


bkaradzic-microsoft:
Thanks for the review! Pushed f7e1fd1 addressing the CI failure and all five inline comments.

**CI red on JSC / V8-on-Ubuntu / JSI (root cause):** `Blob::TryGetData` used `napi_instanceof` (`Napi::Object::InstanceOf`), which is unreliable for ObjectWrap-based constructors on those engine adapters — it returned false for genuine `Blob` instances, so every `blob:` test failed at runtime there while passing on QuickJS. Replaced it with a prototype-chain walk against `Blob.prototype` via `napi_get_prototype` (implemented and sound on all four adapters; JSC uses `JSObjectGetPrototype`). This also correctly accepts `Blob` subclasses such as `File`.

**Review comments:**
1. **XHR `GetResponse` synthetic `CallbackInfo`** — removed; the blob branch now returns the response string directly (no `Napi::CallbackInfo{env, nullptr}`).
2. **XHR `open()` reuse + method validation** — `open()` now resets all blob state up front so a reused instance can't mix blob:/transport requests, and the HTTP method is validated for `blob:` URLs too.
3. **Revoke between `open()` and `send()`** — the object-URL store is now (re-)resolved at `send()` time, so a revoke after `open()` surfaces as a network error (status 0 + `error`).
4. **URL Readme.md** — updated to describe the in-memory process-global `blob:` registry (was `data:` URL / no-op).
5. **URL.h header docs** — corrected "per-environment" to "process-global".

Added tests for each behavior (non-Blob rejection, revoke-between-open-and-send, XHR instance reuse). Local UnitTests: **217 passing, 0 failing** (QuickJS); watching CI for the other engines.

bkaradzic-microsoft:
CI is now fully green across all engines (QuickJS/V8/Chakra/Hermes/JSI and JSC on Ubuntu/macOS/iOS + sanitizers).

Root cause of the earlier JSC-only failures: node-addon-api's `DefineClass` constructors are callable-as-constructor but are **not** reported as functions by JSC's `napi_typeof` (`JSObjectIsFunction` returns false). `TryGetData` originally gated on the constructor being a function (via `InstanceOf`), so on JSC it short-circuited and every blob: test failed at runtime with "argument is not a Blob", while passing on the other engines. It now detects a Blob by walking the prototype chain against `Blob.prototype` using the JS-level `Object.getPrototypeOf` and probing the constructor with `IsObject()` — using only node-addon-api wrappers (the raw `napi_get_prototype` is not exposed by the JSI adapter and returns the raw [[Prototype]] on JSC).

I also dropped the "XHR instance can be reused for a second blob: request" test: it passes on QuickJS but hangs the JSC mocha harness (two sequential async blob-send dispatches on one reused instance never pump the second `loadend`). The reset-on-open code fix that the test exercised is retained; this looks like a harness/event-loop quirk rather than a logic bug.

Polyfills/Blob/Source/Blob.cpp
```diff
@@ -139,4 +139,75 @@ namespace Babylon::Polyfills::Blob
...
+        const auto env = object.Env();
+        const auto global = env.Global();
+
+        // Verify `object` is a Blob (or a subclass such as File) by walking its prototype chain and
+        // comparing each link against Blob.prototype, using the JS-level Object.getPrototypeOf.
+        //
+        // Two engine-adapter quirks drive this implementation:
+        //  * The Blob constructor is created via node-addon-api's DefineClass. On the JavaScriptCore
+        //    adapter such constructors are callable-as-constructor but are NOT reported as functions
+        //    by napi_typeof (JSObjectIsFunction returns false), so we must probe the constructor with
```
bghgary: This works around JSC's napi_typeof reporting DefineClass constructors as napi_object. That adapter bug is already tracked in #194, which names the Blob polyfill and recommends fixing napi_typeof to also consult JSObjectIsConstructor. The Fetch and File polyfills carry the same workaround. Fix the adapter per #194 rather than adding a third copy here.

Polyfills/URL/Source/URL.cpp
```diff
@@ -1,11 +1,84 @@
...
 
+namespace
+{
+    // ---- Blob URL registry -------------------------------------------------------------------
+    // URL.createObjectURL/revokeObjectURL are backed by an in-memory store. The XMLHttpRequest and
+    // fetch polyfills resolve minted blob: URLs against this store (see the public
+    // Register/Revoke/TryResolveObjectURL functions below), since the underlying transport has no
+    // notion of the blob: scheme.
+    //
+    // The store is process-global rather than per-environment: the Node-API engine adapters this
```
bghgary: The browser blob URL store holds a reference to the immutable Blob, not a copy. This copies the bytes into the store, and TryResolveObjectURL copies them out again on every resolve. Hold a shared_ptr to the Blob's buffer, scoped per environment so the bytes are freed on revoke or environment teardown. Per-env scoping needs napi_add_env_cleanup_hook, which the JSC and QuickJS adapters lack and nothing currently tracks; the process-global store here is a workaround for that gap.

Polyfills/Fetch/Source/Fetch.cpp
```diff
@@ -404,6 +405,34 @@ namespace Babylon::Polyfills::Internal
                         signal = init.Get("signal");
                     }
 
+                    // blob: URLs (URL.createObjectURL) resolve against the in-memory object-URL store
+                    // instead of the UrlLib transport, which only understands app/file/http(s). Serve
+                    // the buffered bytes synchronously as a 200 response, or reject with a network
+                    // error (TypeError) when the URL was never registered or has been revoked.
+                    if (url.rfind("blob:", 0) == 0)
```
bghgary: blob: is special-cased here and again in XMLHttpRequest, and image/video src and texture loaders would each need the same branch. Register the store with UrlLib as a blob: scheme resolver so every consumer resolves it uniformly, instead of per-polyfill branches.

bghgary (changes requested):

Right direction, and it addresses the earlier data: URL feedback. But this is a partial implementation of a feature that should be designed as a whole first. Concerns inline.

Polyfills/Fetch/Source/Fetch.cpp
bkaradzic-microsoft: Agreed this is the right long-term shape. Since UrlLib is pulled in as a pinned upstream dependency (`BabylonJS/UrlLib`), a `blob:` scheme resolver needs to land there first and then be wired in here, so fetch / XHR / image+video `src` / texture loaders all resolve uniformly instead of each carrying a branch. I've left the per-polyfill branch in place for now and am tracking the resolver as a follow-up rather than folding a UrlLib fork into this PR. Happy to open an issue on `BabylonJS/UrlLib` for the scheme-resolver API if that works.

Polyfills/Blob/Source/Blob.cpp
bkaradzic-microsoft: Fixed at the adapter per #194: `napi_typeof` in `js_native_api_javascriptcore.cc` now also consults `JSObjectIsConstructor`, so `DefineClass` constructors report as `napi_function`. Dropped the workaround here and the matching copies in the **File** and **Fetch** polyfills, which now use `IsFunction()`. The `Object.getPrototypeOf` prototype-walk stays, since that's a separate JSI/JSC portability concern (raw `napi_get_prototype` isn't exposed by every adapter). Done in 2fc473d.

Polyfills/URL/Source/URL.cpp
bkaradzic-microsoft: Updated in 2fc473d: the store now holds the bytes in a `shared_ptr<const std::vector<std::byte>>` and `TryResolveObjectURL` hands out that shared reference instead of copying on every resolve (`XMLHttpRequest` shares the buffer too, rather than copying it out). Register still does a single copy out of the Blob, since the Blob owns its bytes in a plain `std::vector`; sharing the Blob's own buffer would require the Blob polyfill to expose a `shared_ptr`.

True per-environment scoping/teardown still needs `napi_add_env_cleanup_hook`, which the JSC and QuickJS adapters don't implement, so the store stays process-global as a documented workaround for that gap.

Polyfills/Fetch/Source/Fetch.cpp
bkaradzic-microsoft: Opened #213 to track the `blob:` scheme-resolver work (UrlLib hook + removing the per-polyfill branches). Filed it here since `BabylonJS/UrlLib` has issues disabled.

bkaradzic-microsoft: Implemented. Rather than each polyfill special-casing `blob:`, UrlLib now supports pluggable URL scheme resolvers (`UrlRequest::RegisterSchemeResolver`), and the URL polyfill registers a process-global `blob` resolver backed by the object-URL store. `fetch` and `XMLHttpRequest` now go through the ordinary UrlLib transport path for `blob:` URLs — their blob-specific branches (and the `URL::TryResolveObjectURL` seam) are gone. A revoked/unknown `blob:` URL still surfaces as a status-0 network error. All 216 unit tests pass.

UrlLib change (fork branch pending upstream, since BabylonJS/UrlLib has issues disabled): bkaradzic-microsoft/UrlLib@109e373 — the CMake pin is temporarily bumped to that fork commit. JsRuntimeHost side: fa65ef6.

Polyfills/URL/Source/URL.cpp
bghgary: [Responded by Copilot on behalf of @bghgary]

The per-env gap is still untracked, which was the specific concern here. #213 covered the scheme resolver and is now implemented, #194 was the `napi_typeof` bug this PR fixes, and #214 is the UrlLib Apple lifetime bug, so nothing tracks the missing per-env teardown that the header comment and the description both cite as the reason for the process-global store. It needs its own issue.

Two details in that justification are off. `napi_add_env_cleanup_hook` is implemented by no adapter in `Core/Node-API/Source`, not just QuickJS. And "reclaimed at process exit, mirroring how browsers retain blob URLs until unload" is backwards: a browser frees blob URLs at unload, which corresponds to environment teardown here, not process exit. An embedder that creates and destroys environments leaks every un-revoked entry for the life of the process.

Optional: `RegisterObjectURL` still copies the bytes out of the Blob, so `createObjectURL` doubles a large buffer for as long as the URL lives. `Blob::m_data` is only assigned in constructor paths, so holding it as `shared_ptr<const std::vector<std::byte>>` and returning that from `TryGetData` would drop the copy and retire the "valid only while `object` is alive" caveat.

bghgary (commented):

All three points addressed. One open item on the store thread.

The description still describes the design this PR replaced: it credits the removed `TryResolveObjectURL` seam, says fetch and XMLHttpRequest "intercept `blob:`" when neither branch remains, doesn't mention the UrlLib scheme resolver that now does the work, and omits the `js_native_api_javascriptcore.cc` `napi_typeof` change that affects every JSC consumer. Test count reads 214 against the 216 in the last two commits. It becomes the squash commit message, so worth rewriting.

bkaradzic-microsoft:
Thanks -- you were right on all three counts, and I've pushed fixes.

**The store comment was stale and wrong.** Three separate problems:

1. It described a `TryResolveObjectURL` seam that the request polyfills call. That function no longer exists -- the design moved to a `blob:` scheme resolver registered with `UrlLib::UrlRequest::RegisterSchemeResolver`, so nothing intercepts per-polyfill anymore. The only two remaining mentions of `TryResolveObjectURL` in the tree were inside that comment.
2. `napi_add_env_cleanup_hook` -- I said "QuickJS does not implement it". In fact **no** adapter in this repo implements it: grepping `Core/Node-API/Source` returns zero matches, and the only hit repo-wide is the vendored `napi-inl.h`. Corrected.
3. "reclaimed at process exit, mirroring how browsers retain blob URLs until unload" was backwards, as you said. A browser drops a document's blob URLs *when the document unloads*; process-lifetime retention is the opposite of that. The comment now calls it what it is -- a leak of un-revoked entries for the process lifetime, not released on environment teardown.

I filed #215 to track implementing `napi_add_env_cleanup_hook` across the adapters so the store can become per-environment, and both the comment and the public `URL.h` docs now point at it.

**Blob shared_ptr** -- done. `Blob::m_data` is now a `std::shared_ptr<const std::vector<std::byte>>`, so `createObjectURL` shares the buffer with the store instead of copying it, and `TryGetData`'s result no longer has a lifetime tied to the JS object (I dropped that caveat from the header). This is sound because `m_data` is assigned exactly once per Blob: it's only written in `ProcessBlobPart`, which is only reachable from the constructor, and multi-part construction currently throws.

**Also in this push:**
- Repointed the UrlLib pin back at upstream now that BabylonJS/UrlLib#37 has merged. Verified the consumed API (`RegisterSchemeResolver`/`UnregisterSchemeResolver`/`UrlSchemeResolverResult`) is byte-identical to the fork pin, and rebuilt all the polyfills against it.
- Rewrote the PR description, which was describing the old intercept-in-each-polyfill design. It now covers the scheme resolver, the JSC `napi_typeof` fix (which was omitted entirely), the corrected design note, and the right test count -- 216, as you said, not 214.
- Dropped some incidental churn from `XMLHttpRequest.cpp` that had no business being in this PR.

Polyfills/URL/Source/URL.cpp
bkaradzic-microsoft: Fixed -- the comment was describing a design that no longer exists and got two facts wrong on top of that. Rewrote it: no adapter implements `napi_add_env_cleanup_hook` (not just QuickJS), and un-revoked entries leak for the process lifetime rather than mirroring browser unload behavior. #215 tracks making the store per-environment. Details in the top-level reply.

Polyfills/Blob/Include/Babylon/Polyfills/Blob.h
```diff
@@ -3,7 +3,18 @@
...
+
 namespace Babylon::Polyfills::Blob
 {
     void BABYLON_API Initialize(Napi::Env env);
+
+    // Synchronously reads the bytes and MIME type of a Blob JS object created by this polyfill.
+    // Returns false if `object` is not a Blob. A Blob is immutable once constructed, so `outData`
+    // is a shared reference to its buffer rather than a copy, and it stays valid for as long as the
+    // caller holds it -- even if `object` is collected.
+    bool BABYLON_API TryGetData(const Napi::Object& object, std::shared_ptr<const std::vector<std::byte>>& outData, std::string& outType);
```
bghgary: [Reviewed by Copilot on behalf of @bghgary]

Move `TryGetData` out of the public header. One caller (`URL.cpp:850`), and no consumer of `Polyfills/Blob.h` uses anything but `Initialize`.

Follow the `Core/Graphics` pattern in BabylonNative: a `BlobInternal` INTERFACE target carrying `INTERFACE "InternalInclude"`, `URL` linking it `PRIVATE`, and this declaration moving to `InternalInclude/Babylon/Polyfills/BlobInternal.h`.

While it moves, return the result rather than filling out-parameters:

```cpp
struct BlobData
{
    std::shared_ptr<const std::vector<std::byte>> Data;
    std::string Type;
};

std::optional<BlobData> TryGetData(const Napi::Object& object);
```

Polyfills/URL/Include/Babylon/Polyfills/URL.h
```diff
@@ -3,7 +3,41 @@
...
+    // how a browser Blob's bytes stay valid for an in-flight read even if the URL is revoked
+    // mid-flight.
+
+    // Registers `data` under a freshly minted blob: URL and returns it. The Blob's buffer is shared
+    // rather than copied, so createObjectURL does not duplicate a large blob.
+    // The `env` parameter is currently unused but kept for API symmetry and future per-env scoping.
+    std::string BABYLON_API RegisterObjectURL(Napi::Env env, std::shared_ptr<const std::vector<std::byte>> data, std::string type);
+
+    // Releases the entry for `url`, if any. Unknown URLs are ignored (matching the web platform).
+    void BABYLON_API RevokeObjectURL(Napi::Env env, const std::string& url);
```
bghgary: [Reviewed by Copilot on behalf of @bghgary]

Move `RegisterObjectURL` and `RevokeObjectURL` behind the same internal seam. No consumer today, and this is not a contract to commit to yet: the `Napi::Env` parameter is unused in both definitions, #215 will change the scoping, and the documented process-lifetime leak has no bulk-release API.

Export a considered version once #215 lands, and revisit the names then, since `Register`/`Revoke` mismatches the `createObjectURL`/`revokeObjectURL` pair this mirrors.

Lines 15 to 34 are mostly implementation rationale; the caller-facing part is that entries live until revoked.

bghgary (changes requested):

Both new exports should be internal until we are confident of the contract. Details inline.

Polyfills/Blob/Include/Babylon/Polyfills/Blob.h
bkaradzic-microsoft: Done. `TryGetData` now lives in `Polyfills/Blob/InternalInclude/Babylon/Polyfills/BlobInternal.h` behind a `BlobInternal` INTERFACE target, and `URL` links it `PRIVATE`. I followed the in-repo precedent for this rather than the BabylonNative one -- `JsRuntimeInternal` in `Core/JsRuntime/CMakeLists.txt` is the same shape -- so the pattern matches its neighbours. The public `Blob.h` is back to just `Initialize`, and its now-unnecessary `<cstddef>`/`<memory>`/`<string>`/`<vector>` includes went with it.

Signature is as you suggested:

```cpp
struct BlobData
{
    std::shared_ptr<const std::vector<std::byte>> Data;
    std::string Type;
};

std::optional<BlobData> BABYLON_API TryGetData(const Napi::Object& object);
```

(`BABYLON_API` is only a calling convention here, not a visibility attribute, so keeping it costs nothing and matches the surrounding declarations.)

The call site got simpler too -- the default-MIME-type fixup now mutates `blobData->Type` in place instead of juggling two out-parameters.

Polyfills/URL/Include/Babylon/Polyfills/URL.h
bkaradzic-microsoft: Agreed, and these needed even less than an internal seam. Unlike `TryGetData`, which does cross a target boundary, `RegisterObjectURL`/`RevokeObjectURL` had **no** caller outside `URL.cpp` -- so exporting them bought nothing at all. They're now plain functions in the anonymous namespace next to the store they operate on, which also drops the unused `Napi::Env` parameter you flagged rather than carrying it as speculative symmetry. `URL.h` is back to just `Initialize`, byte-identical to `main`.

Your point about the names is well taken: `Register`/`Revoke` was an odd pairing for something mirroring `createObjectURL`/`revokeObjectURL`. That's now a decision deferred to whenever #215 lands and there's a real per-environment contract -- along with the bulk-release question, which is a genuine gap I hadn't considered.

On the header comment: you were right that lines 15-34 were implementation rationale, and with the declarations gone the whole block went with them. The rationale still exists, but where it belongs -- next to the store in `URL.cpp`.

Rebuilt and re-ran the suite: 216 passing, 0 failing, unchanged.

bghgary (approved):

LGTM

PR #209, opened 2026-07-22
Complete Blob streaming and slicing support

matthargett:
Depends on #208. The commits are kept separate so this PR can be rebased to only the Blob change after the Streams prerequisite lands. These are for the five deterministic screenshot test suite failures in scenes 66, 72, 73, 123, and 140, all explicitly blocked on real Blob.stream/Response/decompression support. Besides helping unblock more of the suite, this makes BabylonJS browser apps much more likely to "just work" in BabylonNative.

Adds iterable BlobPart handling, MIME/endings normalization, `slice()`, and lazy byte `stream()` support. Blob composition and slicing share immutable native segments; stream reads allocate only the requested 64 KiB-or-smaller JavaScript chunk. `File` delegates the same APIs to its backing Blob.

Focused regression coverage is adapted from WPT FileAPI Blob tests, WebKit Blob stream chunk/crash tests, and Firefox large `pipeTo` coverage.

Validated with:
- JavaScriptCore ASan/UBSan: 235 JS cases plus native tests
- JavaScriptCore TSan: 235 JS cases plus native tests
- QuickJS Release: 235 JS cases plus native tests


Polyfills/Blob/Source/Blob.cpp
```diff
@@ -77,59 +227,311 @@ namespace Babylon::Polyfills::Internal
...
+            {
+                controller.Get("close").As<Napi::Function>().Call(controller, {});
+                state->BlobData.reset();
+                return callbackEnv.Undefined();
+            }
+
+            constexpr size_t CHUNK_SIZE = 64 * 1024;
+            const auto outputLength = std::min(CHUNK_SIZE, state->BlobData->Size - state->Position);
+            auto outputBuffer = Napi::ArrayBuffer::New(callbackEnv, outputLength);
+            auto destination = static_cast<std::byte*>(outputBuffer.Data());
```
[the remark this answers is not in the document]
matthargett: Fixed in 2a35a0d. BYOB pulls now copy directly into `controller.byobRequest.view` and call `respond(bytesWritten)`, so they allocate no 64 KiB intermediate chunk. Default readers retain the bounded chunk path; WPT-derived tests cover small and offset views across Blob segments.

Tests/UnitTests/Shared/Shared.cpp
```diff
@@ -111,6 +113,30 @@ TEST(JavaScript, All)
...
+
+        const auto installedTransformStream = global.Get("TransformStream");
+        Babylon::Polyfills::Streams::Initialize(env);
+        EXPECT_TRUE(global.Get("ReadableStream").StrictEquals(hostReadableStream));
+        EXPECT_TRUE(global.Get("TransformStream").StrictEquals(installedTransformStream));
+        done.set_value();
+    });
+
+    done.get_future().get();
+}
```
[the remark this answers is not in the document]
matthargett: Fixed in 075ab25. The native test now installs an `UnhandledExceptionHandler` and uses one-shot completion, so an N-API exception resolves the wait with its error text instead of leaving the promise blocked.

Polyfills/Streams/Source/Streams.cpp
```diff
@@ -0,0 +1,55 @@
...
+        }
+
+        const auto exports = Napi::Eval(env, Internal::StreamsScripts::Ponyfill.data(), "jsruntimehost://web-streams-polyfill.js").As<Napi::Object>();
+        for (const auto name : constructorNames)
+        {
+            if (global.Get(name.data()).IsUndefined())
+            {
+                global.Set(name.data(), exports.Get(name.data()));
+            }
+        }
```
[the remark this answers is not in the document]
matthargett: Fixed in 075ab25 (the #208 prerequisite update). Both `null` and `undefined` are treated as missing, and a partial Streams surface is replaced as a complete suite.

matthargett:

This is the first time the branch has had Windows/UWP, JSI, iOS-simulator and Android CI (upstream runs are behind the fork-workflow gate; the fork twin runs the full 24-job matrix), which surfaced a few portability items now fixed on the branch: the embedded Streams bundle split into ≤16 KB literals (MSVC C2026), `Streams` linked into the Android test host, a `size_t`→`double` narrowing MSVC treats as an error, DataView parts read through JS properties (JSI's `Napi::DataView` is an unimplemented stub), and a 60 s budget for the 40k-chunk stream tests, which straddled mocha's 2 s default on the iOS simulator. Fork twin: rebeckerspecialties/JsRuntimeHost#16.

Polyfills/Blob/Source/Blob.cpp
```diff
@@ -78,60 +346,285 @@ namespace Babylon::Polyfills::Internal
...
+        }
+        else if (blobPart.IsDataView())
+        {
+            // Read the view through its JS properties rather than Napi::DataView: the JSI
+            // backend's DataView wrapper is an unimplemented stub, and merely instantiating it
+            // fails the UWP build (its `throw "TODO"` leaves unreachable code, C4702 as error).
+            const auto view = blobPart.As<Napi::Object>();
+            const auto buffer = view.Get("buffer").As<Napi::ArrayBuffer>();
+            const auto bufferData = static_cast<const std::byte*>(buffer.Data());
+            const auto byteOffset = static_cast<size_t>(view.Get("byteOffset").As<Napi::Number>().Int64Value());
```
bghgary: Please read DataView state through non-overridable intrinsic accessors and validate the range against the branded ArrayBuffer before copying. A branded DataView can shadow `buffer`, `byteOffset`, and `byteLength`, making the current pointer arithmetic and `memcpy` read outside the buffer.

Polyfills/Blob/Source/Blob.cpp
```diff
@@ -78,60 +346,285 @@ namespace Babylon::Polyfills::Internal
...
+            const auto blobConstructor = Env().Global().Get("Blob").As<Napi::Function>();
+            if (object.InstanceOf(blobConstructor))
+            {
+                auto nativeBlob = object;
+                if (!object.Get("constructor").StrictEquals(blobConstructor))
+                {
+                    nativeBlob = object.Get("slice").As<Napi::Function>().Call(object, {Napi::Number::New(Env(), 0), object.Get("size")}).As<Napi::Object>();
+                }
+
+                const auto blob = Napi::ObjectWrap<Blob>::Unwrap(nativeBlob);
```
bghgary: Please introduce a brand-safe native backing-data extractor shared by Blob and File, and register Blob prototype callbacks that resolve receivers through it rather than `ObjectWrap<Blob>` member dispatch. Overridable subclass `slice()` methods can substitute unrelated data or native wrappers, while valid calls such as `Blob.prototype.slice.call(file)` and `Blob.prototype.stream.call(file)` currently reinterpret File's native wrapper as Blob.

Polyfills/Blob/Source/Blob.cpp
```diff
@@ -78,60 +346,285 @@ namespace Babylon::Polyfills::Internal
...
+            }
+        }
+
+        std::string contentType;
+        if (info.Length() > 2 && !info[2].IsUndefined())
+        {
+            contentType = NormalizeType(info[2].ToString().Utf8Value());
+        }
+
+        const auto blobConstructor = Env().Global().Get("Blob").As<Napi::Function>();
```
bghgary: Please have initialization retain private per-environment references to the installed Blob and stream-source constructors, and register or capture the applicable ReadableStream constructor without consulting mutable public globals during each operation. Reassigning `Blob` or `ReadableStream` currently breaks existing Blob instances and can make `slice()` construct and unsafely unwrap an unrelated native object.

Polyfills/Blob/Source/Blob.cpp
```diff
@@ -19,45 +181,148 @@ namespace Babylon::Polyfills::Internal
...
+                while (true)
+                {
+                    const auto result = next.Call(iterator, {}).As<Napi::Object>();
+                    if (result.Get("done").ToBoolean().Value())
+                    {
+                        break;
+                    }
 
-                if (blobParts.Length() > 1)
+                    if (AppendBlobPart(*data, result.Get("value")))
```
bghgary: Please separate WebIDL BlobPart conversion from byte materialization: retain converted BufferSource references while converting the options dictionary, then copy their bytes when processing the converted parts. An options getter may mutate a typed array or DataView, and the File API requires that mutation to be reflected because the byte snapshot occurs after argument conversion.

Tests/UnitTests/CMakeLists.txt
```diff
@@ -78,6 +78,7 @@ target_link_libraries(UnitTests
     PRIVATE Performance
     PRIVATE TextDecoder
     PRIVATE TextEncoder
+    PRIVATE Streams
```
bghgary: Please guard all Streams-dependent test code and the desktop and Android links with `JSRUNTIMEHOST_POLYFILL_STREAMS`, or require Streams when tests are enabled. Disabling this optional target leaves both test hosts referencing Streams while the shared native and JavaScript tests still use it unconditionally.

Polyfills/Streams/README.md
```diff
@@ -0,0 +1,14 @@
+# Web Streams
+
+Installs the standard `ReadableStream`, `WritableStream`, `TransformStream`,
+reader, writer, controller, and queuing-strategy globals when the selected
+JavaScript engine does not provide them.
+
+The implementation is the ES5 ponyfill bundle from
```
bghgary: Please implement compliant BYOB buffer transfer on engines without `ArrayBuffer.prototype.transfer` or `structuredClone`, or document and test the narrower contract. The vendored ES5 ponyfill leaves the submitted view attached in that case, including on the supported Chakra runtime, so the returned view can continue aliasing caller-owned storage.

Polyfills/File/Source/File.cpp
```diff
@@ -114,21 +120,11 @@ namespace Babylon::Polyfills::Internal
...
-        {
-            partsArray = Napi::Array::New(env, 0);
-        }
-
         // Delegate byte-buffer construction to the native Blob polyfill so
         // we benefit from its existing BlobPart handling (ArrayBuffer,
         // typed array, string, Blob).
         auto blobCtor = env.Global().Get(JS_BLOB_CONSTRUCTOR_NAME).As<Napi::Function>();
-        auto blobInstance = blobCtor.New({partsArray, blobOptions});
+        auto blobInstance = blobCtor.New({parts, blobOptions});
```
bghgary: Please convert the `fileBits` sequence before converting `fileName` and `options`, while still deferring BufferSource byte copying until Blob processing. The newly supported non-array iterables are currently consumed only after filename conversion and option getters, reversing WebIDL argument-conversion order and observably reordering iterator errors and side effects.

Polyfills/Streams/README.md
```diff
@@ -0,0 +1,14 @@
...
+[`web-streams-polyfill` 4.3.0](https://github.com/MattiasBuelens/web-streams-polyfill/tree/add69059ed08eae6a18559aba49575a280d1529e),
+which is based on the WHATWG reference implementation. The vendored project
+tests this release against the Streams portion of WPT at revision
+[`c05b4473`](https://github.com/web-platform-tests/wpt/tree/c05b447326585237713013c66341eab2cdf967b6/streams).
+
+`Initialize` preserves any Streams constructors already supplied by the host
```
bghgary: Please update this contract and the public header to match the atomic replacement behavior: when any required constructor is unavailable, initialization replaces the full Streams constructor suite.

PR #210, opened 2026-07-22
Add browser-compatible Headers and streamed Response bodies

matthargett:
Depends on #208 and #209. The commits remain separate so this can be rebased to the Fetch-only change after those prerequisites land.

Replaces the ad hoc reusable buffered response object with standard-shaped Headers and Response classes. Headers normalize and validate fields, preserve current Set-Cookie behavior, and support live sorted iteration. Response bodies are ReadableStreams with single-consumption semantics, cloning, and browser-shaped metadata and body readers. Empty BufferSource bodies retain a non-null stream without enqueueing an invalid zero-length byte chunk.

UrlLib still buffers the transport response, but completion now copies it once into JavaScript-owned memory. The Response wrapper does not make another snapshot; multi-chunk consumers retain chunk references and allocate one exact contiguous result only when needed. Header iteration is mutation-cached, and delete/set compact storage in place.

Focused coverage is adapted from WPT fetch/api/headers and fetch/api/response, WebKit body-init and empty-chunk regressions, Firefox open-stream teardown, and a bounded Chromium disturbed-stream regression.

Validated on JavaScriptCore ASan/UBSan, JavaScriptCore TSan, and QuickJS Release: 254 JavaScript cases plus 8 native tests in each configuration.


Polyfills/Fetch/Source/Fetch.cpp
```diff
@@ -157,188 +147,43 @@ namespace Babylon::Polyfills::Internal
...
+            auto global = env.Global();
+            const auto exports = Napi::Eval(env, FetchScripts::Polyfill, "jsruntimehost://fetch-polyfill.js").As<Napi::Object>();
+            if (global.Get("Headers").IsUndefined())
+            {
+                global.Set("Headers", exports.Get("Headers"));
+            }
+            if (global.Get("Response").IsUndefined())
+            {
+                global.Set("Response", exports.Get("Response"));
+            }
```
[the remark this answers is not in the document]
matthargett: Fixed in ec6cc4f. If either Headers or Response is `undefined` or `null`, initialization installs the complete pair; a complete host pair remains untouched. The native regression also verifies idempotence and cross-class `instanceof` identity.

Polyfills/Fetch/Source/FetchPolyfill.js
```diff
@@ -0,0 +1,604 @@
...
+    class Response {
+        constructor(body, init) {
+            initializeResponse(this, body === undefined ? null : body, init, null);
+        }
+
+        get body() { return requireResponseState(this).body; }
+        get bodyUsed() {
+            var state = requireResponseState(this);
+            return state.used || (state.body !== null && state.body._disturbed === true);
+        }
```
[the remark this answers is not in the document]
matthargett: Fixed in ec6cc4f. Response now uses weak per-stream state updated through standard reader, cancellation, iteration, tee, and pipe operations. It preserves stream identity and does not retain Response state. The WPT-derived tests mask `_disturbed` and also cover a read/release performed before Response construction.

Polyfills/Fetch/Source/FetchPolyfill.js
```diff
@@ -0,0 +1,604 @@
...
+                    chunk = null;
+                    controller.enqueue(value);
+                }
+                controller.close();
+            },
+            cancel: function() {
+                chunk = null;
+            }
+        });
+    }
```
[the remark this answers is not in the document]
matthargett: Fixed in 01b7d90. Zero-length BufferSource bodies now close without enqueueing an invalid empty byte chunk while retaining non-null body and `bodyUsed` semantics. Focused coverage was added.

matthargett:

This is the first time the branch has had Windows/UWP, JSI, iOS-simulator and Android CI (upstream runs are behind the fork-workflow gate; the fork twin runs the full 24-job matrix), which surfaced a few portability items now fixed on the branch: the embedded Streams bundle split into ≤16 KB literals (MSVC C2026), `Streams` linked into the Android test host, a `size_t`→`double` narrowing MSVC treats as an error, DataView parts read through JS properties (JSI's `Napi::DataView` is an unimplemented stub), and a 60 s budget for the 40k-chunk stream tests, which straddled mocha's 2 s default on the iOS simulator. Also: the embedded Fetch polyfill hit the same MSVC literal cap and now uses adjacent ≤16 KB literals, and the Linux ASan job gained an LSan suppression for a WebKitGTK-internal allocation (`WTF::BitVector` growth under `JSObjectMake` → `setPrototypeDirect`, reached from napi reference creation — JSC's own never-freed bookkeeping, mirroring the existing TSan suppressions file). Fork twin: rebeckerspecialties/JsRuntimeHost#17.

Polyfills/Fetch/Source/FetchPolyfill.js
```diff
@@ -0,0 +1,704 @@
...
+            var cloneBody = null;
+            if (state.body !== null) {
+                var branches = state.body.tee();
+                state.body = branches[0];
+                state.bodyTracker = trackStream(state.body);
+                cloneBody = branches[1];
+            }
+            var clone = new Response(cloneBody, {
+                headers: state.headers,
+                status: state.status,
```
bghgary: `Response.error().clone()` sends status `0` through the public constructor, which rejects statuses below 200. Please clone internal response state directly and cover error-response cloning.

Polyfills/Fetch/Source/FetchPolyfill.js
```diff
@@ -0,0 +1,704 @@
...
+    }
+
+    function joinChunks(chunksAndTotal) {
+        var chunks = chunksAndTotal[0];
+        var total = chunksAndTotal[1];
+        if (chunks.length === 0) {
+            return new Uint8Array(0);
+        }
+        if (chunks.length === 1) {
+            return chunks[0];
```
bghgary: The single-chunk fast path returns the producer's `Uint8Array` directly, so `bytes()` and `arrayBuffer()` expose aliases instead of owned consumed-body data. Please copy the chunk before returning it.

Polyfills/Fetch/Source/FetchPolyfill.js
```diff
@@ -0,0 +1,704 @@
...
+
+    Object.defineProperty(Headers.prototype, Symbol.toStringTag, {
+        configurable: true,
+        value: "Headers"
+    });
+
+    function createResponseHeaders(init) {
+        var headers = new Headers(init);
+        var state = requireHeaderState(headers);
+        state.guard = "response";
```
bghgary: Headers on error, redirect, fetched responses, and their clones remain mutable; browser responses make those headers immutable while directly constructed `Response` headers stay mutable. Please add the appropriate guard and cover `append`, `set`, and `delete`.

Polyfills/Blob/Source/Blob.cpp
```diff
@@ -78,60 +228,360 @@ namespace Babylon::Polyfills::Internal
...
+        else if (blobPart.IsObject())
+        {
+            const auto object = blobPart.As<Napi::Object>();
+            const auto blobConstructor = Env().Global().Get("Blob").As<Napi::Function>();
+            if (object.InstanceOf(blobConstructor))
+            {
+                auto nativeBlob = object;
+                if (!object.Get("constructor").StrictEquals(blobConstructor))
+                {
+                    nativeBlob = object.Get("slice").As<Napi::Function>().Call(object, {Napi::Number::New(Env(), 0), object.Get("size")}).As<Napi::Object>();
```
bghgary: Blob composition invokes overridable `slice()` and `size` hooks on Blob subclasses. Please extract branded Blob/File storage internally without running script.

Polyfills/Fetch/Source/FetchPolyfill.js
```diff
@@ -0,0 +1,704 @@
...
+    Object.defineProperty(Response.prototype, Symbol.toStringTag, {
+        configurable: true,
+        value: "Response"
+    });
+
+    function createFetchResponse(ResponseConstructor, bytes, init, metadata) {
+        var status = init && init.status;
+        var body = nullBodyStatuses.indexOf(status) === -1 ? streamFromBytes(bytes) : null;
+        var response = new ResponseConstructor(body, init);
+        var state = responseStates.get(response);
```
bghgary: When host `Headers`/`Response` constructors are preserved, fetched host `Response` objects never enter `responseStates`, so `url`, `redirected`, and `type` retain default values. Please preserve transport metadata for host `Response` instances or install a compatible paired implementation.

Polyfills/Blob/Source/Blob.cpp
```diff
@@ -78,60 +228,360 @@ namespace Babylon::Polyfills::Internal
...
+            }
+        }
+
+        std::string contentType;
+        if (info.Length() > 2 && !info[2].IsUndefined())
+        {
+            contentType = NormalizeType(info[2].ToString().Utf8Value());
+        }
+
+        const auto blobConstructor = Env().Global().Get("Blob").As<Napi::Function>();
```
bghgary: `Blob.slice()` resolves the current writable global `Blob`, so replacing that global breaks slices on existing native Blob instances. Please construct slices through the captured native constructor or an internal factory.

Polyfills/Blob/Source/Blob.cpp
```diff
@@ -78,60 +228,360 @@ namespace Babylon::Polyfills::Internal
...
+            length = array.ByteLength();
+            isBufferSource = true;
+        }
+        else if (blobPart.IsDataView())
+        {
+            // Read the view through its JS properties rather than Napi::DataView: the JSI
+            // backend's DataView wrapper is an unimplemented stub, and merely instantiating it
+            // fails the UWP build (its `throw "TODO"` leaves unreachable code, C4702 as error).
+            const auto view = blobPart.As<Napi::Object>();
+            const auto buffer = view.Get("buffer").As<Napi::ArrayBuffer>();
```
bghgary: DataView `buffer`, `byteOffset`, and `byteLength` are read through shadowable JavaScript properties, so inconsistent values can drive an out-of-bounds native copy. Please use intrinsic or internal-slot access and validate offset plus length against the buffer.

Polyfills/Fetch/Source/FetchPolyfill.js
```diff
@@ -0,0 +1,704 @@
...
+            return response;
+        }
+
+        static json(data, init) {
+            var text = JSON.stringify(data);
+            if (text === undefined) {
+                throw new TypeError("Response.json data is not JSON serializable");
+            }
+            var hasExplicitContentType = init !== undefined && init !== null &&
+                new Headers(init.headers).has("content-type");
```
bghgary: `Response.json()` consumes `init.headers` once for content-type inspection and again in the constructor, so a one-shot iterable loses explicit headers. Please normalize once and reuse the normalized headers.

Polyfills/Fetch/Source/FetchPolyfill.js
```diff
@@ -0,0 +1,704 @@
...
+
+    function trackReader(reader, tracker) {
+        if (readerTrackers.has(reader)) {
+            return reader;
+        }
+        readerTrackers.set(reader, tracker);
+        for (var name of ["read", "cancel"]) {
+            wrapMethod(reader, name, function(original) {
+                return function() {
+                    tracker.disturbed = true;
```
bghgary: Failed stream operations mark the body disturbed before the underlying operation succeeds; reading a released reader and cancelling a locked stream reject without disturbing native streams. Please mark disturbance only when the operation actually disturbs the stream.

PR #211, opened 2026-07-22
Add CompressionStream and DecompressionStream polyfills

matthargett:
## Summary
- add optional `CompressionStream` and `DecompressionStream` polyfills for `gzip`, `deflate`, and `deflate-raw`
- preserve host-provided constructors and initialize WHATWG Streams as the only polyfill dependency
- use the platform/existing zlib target when available, with a pinned zlib 1.3.1 fallback

Depends on #208.

## Ownership
Input BufferSource views are borrowed only during a synchronous zlib call. Output is buffered before any enqueue callback can run JavaScript, then copied once into exact-size JS-owned Uint8Arrays. Each active codec reuses one 64 KiB native scratch buffer and releases zlib/scratch storage immediately on completion or error.

## Coverage
Focused tests cover the [WPT compression suite](https://github.com/web-platform-tests/wpt/tree/master/compression), including constructor conversion, BufferSource variants and offsets, split/empty chunks, large flush output, corrupt/truncated input, trailing data, and Uint8Array output. They also exercise the input-invalidation safeguard documented by [Chromium's transformer](https://chromium.googlesource.com/chromium/src/+/main/third_party/blink/renderer/modules/compression/deflate_transformer.cc), Firefox's [output-before-extra-input-error ordering](https://searchfox.org/firefox-main/source/dom/compression/BaseAlgorithms.cpp), and WebKit's [BufferSource validation path](https://github.com/WebKit/WebKit/blob/main/Source/WebCore/Modules/compression/CompressionStream.js).

Validated with:
- JavaScriptCore ASan/UBSan: 229 JS cases plus native tests
- QuickJS Release: 229 JS cases plus native tests
- JavaScriptCore TSan: 229 JS cases plus native tests


matthargett:
Pushed c3893a4: the codec staged its output chunks as `Napi::Uint8Array` values in a `std::vector` while zlib produced more. The JavaScriptCore backend's handle scopes are stubs and it roots only what is on the C stack or protected, so a collection triggered by allocating the next chunk could reclaim an earlier pending one — the 256 KiB reentrancy test then failed with "incorrect header check" on macOS and Android CI, and deterministically under `JSC_collectContinuously=1`. The output is now staged as bytes and each `Uint8Array` is created right before it is enqueued (3/3 green under continuous GC where the old code failed 3/3).

Polyfills/Compression/Source/CompressionPolyfill.js
```diff
@@ -0,0 +1,137 @@
...
+
+    return function createCompressionPolyfill(NativeCompressionCodec) {
+        var compressionStates = new WeakMap();
+        var decompressionStates = new WeakMap();
+
+        function normalizeFormat(value) {
+            // String concatenation follows Web IDL's DOMString conversion,
+            // including throwing for Symbol values and propagating user
+            // conversion exceptions.
+            var format = value + "";
```
bghgary: Please convert with the string primitive hint while still rejecting Symbols, and add a `Symbol.toPrimitive` regression case; `value + ""` uses the default hint and rejects inputs that native `CompressionStream` accepts.

Polyfills/Compression/Source/CompressionPolyfill.js
```diff
@@ -0,0 +1,137 @@
...
+                throw new TypeError("Unsupported compression format: '" + format + "'");
+            }
+            return format;
+        }
+
+        function toUint8Array(chunk) {
+            if (chunk instanceof ArrayBuffer) {
+                return new Uint8Array(chunk);
+            }
+            if (ArrayBuffer.isView(chunk) && chunk.buffer instanceof ArrayBuffer) {
```
bghgary: Please accept `BufferSource` values by their internal brands rather than same-realm `instanceof` checks. A valid `ArrayBuffer` remains a `BufferSource` after its prototype changes, and cross-realm values likewise fail these checks; add regression coverage.

Polyfills/Compression/Source/CompressionPolyfill.js
```diff
@@ -0,0 +1,137 @@
...
+        function getState(states, value) {
+            var state = states.get(value);
+            if (!state) {
+                throw new TypeError("Illegal invocation");
+            }
+            return state;
+        }
+
+        function CompressionStream(format) {
+            if (!(this instanceof CompressionStream)) {
```
bghgary: Please reject every call made without construction rather than using `instanceof`; `CompressionStream.call(Object.create(CompressionStream.prototype), "gzip")` currently succeeds, while native constructors throw. Add the corresponding regression case for both constructors.

Polyfills/Compression/Source/Compression.cpp
```diff
@@ -0,0 +1,415 @@
...
+    void BABYLON_API Initialize(Napi::Env env)
+    {
+        Streams::Initialize(env);
+
+        Napi::HandleScope scope{env};
+        auto global = env.Global();
+        const auto compressionStream = global.Get("CompressionStream");
+        const auto decompressionStream = global.Get("DecompressionStream");
+        if (!compressionStream.IsUndefined() && !decompressionStream.IsUndefined())
+        {
```
bghgary: Please treat `null` the same as a missing constructor, as the Streams initializer does, and add coverage. With both globals set to `null`, this early return leaves both APIs unusable even though the host provides neither constructor.

Polyfills/Streams/README.md
```diff
@@ -0,0 +1,14 @@
...
+JavaScript engine does not provide them.
+
+The implementation is the ES5 ponyfill bundle from
+[`web-streams-polyfill` 4.3.0](https://github.com/MattiasBuelens/web-streams-polyfill/tree/add69059ed08eae6a18559aba49575a280d1529e),
+which is based on the WHATWG reference implementation. The vendored project
+tests this release against the Streams portion of WPT at revision
+[`c05b4473`](https://github.com/web-platform-tests/wpt/tree/c05b447326585237713013c66341eab2cdf967b6/streams).
+
+`Initialize` preserves any Streams constructors already supplied by the host
+engine and fills only missing globals.
```
bghgary: Please update this README and the public `Streams.h` contract to document the implemented replacement behavior: if any listed constructor is missing or `null`, initialization replaces the complete constructor suite. The current partial-preservation claims contradict `Streams::Initialize` and `Streams.ReplacesPartialHostSuiteAndIsIdempotent`.

Issue #67, opened 2023-12-04
Handle Base64 in uris


CedricGuillemet:
handle base64 data uris generically instead of specific case like https://github.com/BabylonJS/BabylonNative/pull/1328

bkaradzic-microsoft:
Being addressed by #212 (open) — *Support data URLs in the fetch polyfill*, which resolves both percent-encoded and Base64 `data:` URLs generically (WHATWG forgiving-base64 validation, default/explicit media types, fragment stripping, case-insensitive `base64`, missing padding, invalid-alphabet rejection), with cases ported from WPT `fetch/data-urls` and Chromium's `DataURL` tests — rather than the one-off decode this issue was filed against (BabylonJS/BabylonNative#1328).

Note on scope: #212 implements `data:` inside `Polyfills/Fetch/Source/Fetch.cpp`, so it covers the `fetch` path only. `XMLHttpRequest` and other `UrlRequest` consumers (image/video `src`, texture loaders) still won't resolve `data:`. The fully generic answer is the same shape as the `blob:` work in #207/#213 — a `data:` scheme resolver registered through `UrlLib::UrlRequest::RegisterSchemeResolver`, so every consumer gets it uniformly.

Leaving this open until that generic path exists; #212 is the first (and largest) piece.

PR #212, opened 2026-07-22
Support data URLs in the fetch polyfill

matthargett:
## Summary
- resolve percent-encoded and Base64 `data:` URLs locally through `fetch`
- preserve standard `Response` metadata and body-consumption behavior
- validate WHATWG forgiving Base64 without the previous general-purpose decoder dependency

Depends on #210.

## Memory
Percent decoding reserves once. Base64 URLs stream percent-decoded bytes through validation and decoding passes without materializing an intermediate buffer, then allocate the exact native payload size. The completed body follows the shared fetch response path and is copied once into JavaScript-owned storage.

## Coverage
Focused cases are adapted from [WPT fetch/data-urls](https://github.com/web-platform-tests/wpt/tree/master/fetch/data-urls) and [Chromium's DataURL tests](https://chromium.googlesource.com/chromium/src/+/main/net/base/data_url_unittest.cc). They cover default and explicit media types, fragment stripping, case-insensitive `base64`, ASCII whitespace, missing padding, byte-preserving percent decoding, and invalid alphabet/padding rejection.

Validated with 272 JavaScript cases plus 8 native tests on:
- JavaScriptCore ASan/UBSan
- JavaScriptCore TSan
- QuickJS Release


matthargett:

This is the first time the branch has had Windows/UWP, JSI, iOS-simulator and Android CI (upstream runs are behind the fork-workflow gate; the fork twin runs the full 24-job matrix), which surfaced a few portability items now fixed on the branch: the embedded Streams bundle split into ≤16 KB literals (MSVC C2026), `Streams` linked into the Android test host, a `size_t`→`double` narrowing MSVC treats as an error, DataView parts read through JS properties (JSI's `Napi::DataView` is an unimplemented stub), and a 60 s budget for the 40k-chunk stream tests, which straddled mocha's 2 s default on the iOS simulator. Also: the embedded Fetch polyfill hit the same MSVC literal cap and now uses adjacent ≤16 KB literals, and the Linux ASan job gained an LSan suppression for a WebKitGTK-internal allocation (see #210 for detail). Re #67: this PR covers `fetch()`; the generic `data:` scheme resolver through `UrlLib::UrlRequest::RegisterSchemeResolver` that @bkaradzic-microsoft described is the natural follow-up so XHR and texture loaders get it too. Fork twin: rebeckerspecialties/JsRuntimeHost#18.

matthargett:
Following up on #67: the branch now carries the generic path @bkaradzic-microsoft described — the WHATWG `data:` URL processor from this PR is lifted out of `Fetch.cpp` into `Polyfills/URL` (a `URLInternal` seam, like `BlobInternal`) and registered with `UrlLib::UrlRequest::RegisterSchemeResolver("data", …)` next to `blob:`, so XMLHttpRequest — the path Babylon.js' asset and texture loaders take — and every other UrlLib consumer resolve `data:` URLs too. A malformed one is a network error per spec (`handled == false`). `fetch` keeps its synchronous fast path on the shared decoder.

Tests: XMLHttpRequest text / base64→ArrayBuffer / malformed→status 0, a `data:` → Blob → object-URL round trip, and a base64 gzip `data:` URL piped through `DecompressionStream` (skipped where the compression polyfill isn't present — it's a sibling PR). Verified locally on macOS JavaScriptCore; fork twin: rebeckerspecialties/JsRuntimeHost#18.

Polyfills/Fetch/Source/FetchPolyfill.js
```diff
@@ -0,0 +1,704 @@
...
+
+    Object.defineProperty(Headers.prototype, Symbol.toStringTag, {
+        configurable: true,
+        value: "Headers"
+    });
+
+    function createResponseHeaders(init) {
+        var headers = new Headers(init);
+        var state = requireHeaderState(headers);
+        state.guard = "response";
```
bghgary: Apply an immutable guard to headers on fetch-created, error, and redirect responses. Their headers currently keep the ordinary response guard, so `set`, `append`, and `delete` succeed instead of throwing `TypeError`.

Polyfills/Fetch/Source/FetchPolyfill.js
```diff
@@ -0,0 +1,704 @@
...
+            var result = await reader.read();
+            if (result.done) {
+                pieces.push(decoder.decode());
+                return pieces.join("");
+            }
+            if (!(result.value instanceof Uint8Array)) {
+                throw new TypeError("Response body stream yielded a non-Uint8Array chunk");
+            }
+            if (result.value.byteLength !== 0) {
+                pieces.push(decoder.decode(result.value, { stream: true }));
```
bghgary: Preserve incomplete UTF-8 sequences across chunks. This repository's `TextDecoder` ignores `{ stream: true }`, so a character split between chunks becomes two replacement characters. Decode the combined bytes or add streaming decoder state.

Polyfills/Fetch/Source/FetchPolyfill.js
```diff
@@ -0,0 +1,704 @@
...
+            }
+
+            var cloneBody = null;
+            if (state.body !== null) {
+                var branches = state.body.tee();
+                state.body = branches[0];
+                state.bodyTracker = trackStream(state.body);
+                cloneBody = branches[1];
+            }
+            var clone = new Response(cloneBody, {
```
bghgary: Preserve network-error responses without routing status `0` through the public constructor. `Response.error().clone()` currently throws `RangeError` here instead of returning another error response.

Polyfills/Fetch/Source/FetchPolyfill.js
```diff
@@ -0,0 +1,704 @@
...
+    Object.defineProperty(Response.prototype, Symbol.toStringTag, {
+        configurable: true,
+        value: "Response"
+    });
+
+    function createFetchResponse(ResponseConstructor, bytes, init, metadata) {
+        var status = init && init.status;
+        var body = nullBodyStatuses.indexOf(status) === -1 ? streamFromBytes(bytes) : null;
+        var response = new ResponseConstructor(body, init);
+        var state = responseStates.get(response);
```
bghgary: Preserve fetched-response metadata without breaking the retained host `Response` identity. This private state is absent for host responses, so fetch returns constructor defaults for `url`, `redirected`, and `type`. Route metadata through a strategy compatible with the host constructor.

Polyfills/Streams/README.md
```diff
@@ -0,0 +1,14 @@
...
+reader, writer, controller, and queuing-strategy globals when the selected
+JavaScript engine does not provide them.
+
+The implementation is the ES5 ponyfill bundle from
+[`web-streams-polyfill` 4.3.0](https://github.com/MattiasBuelens/web-streams-polyfill/tree/add69059ed08eae6a18559aba49575a280d1529e),
+which is based on the WHATWG reference implementation. The vendored project
+tests this release against the Streams portion of WPT at revision
+[`c05b4473`](https://github.com/web-platform-tests/wpt/tree/c05b447326585237713013c66341eab2cdf967b6/streams).
+
+`Initialize` preserves any Streams constructors already supplied by the host
```
bghgary: Update this to match `Initialize`: if any constructor is missing, it replaces the complete Streams suite so the constructors remain compatible; it does not preserve the existing constructors and fill only the missing globals.

Polyfills/Blob/Source/Blob.cpp
```diff
@@ -19,45 +74,137 @@ namespace Babylon::Polyfills::Internal
                     InstanceMethod("text", &Blob::Text),
                     InstanceMethod("arrayBuffer", &Blob::ArrayBuffer),
                     InstanceMethod("bytes", &Blob::Bytes),
+                    InstanceMethod("slice", &Blob::Slice),
```
bghgary: Make the new Blob methods use a brand and backing path that also recognizes File. `Blob.prototype.slice.call(file)` and `.stream.call(file)` currently enter Blob's native callbacks with a File wrapper, so unwrapping `this` as Blob can miscast the native pointer.

Polyfills/Blob/Source/Blob.cpp
```diff
@@ -78,60 +228,360 @@ namespace Babylon::Polyfills::Internal
...
+            isBufferSource = true;
+        }
+        else if (blobPart.IsObject())
+        {
+            const auto object = blobPart.As<Napi::Object>();
+            const auto blobConstructor = Env().Global().Get("Blob").As<Napi::Function>();
+            if (object.InstanceOf(blobConstructor))
+            {
+                auto nativeBlob = object;
+                if (!object.Get("constructor").StrictEquals(blobConstructor))
```
bghgary: Read Blob parts from native backing state without invoking overridable `constructor`, `slice`, or `size` properties. A Blob subclass that overrides `slice()` makes `new Blob([subclassInstance])` run user code or throw, although Blob construction should copy its bytes directly.

Polyfills/URL/Source/DataUrl.cpp
```diff
@@ -0,0 +1,206 @@
...
+            std::string finalParameter{mediaType.substr(semicolon + 1)};
+            TrimAsciiWhitespace(finalParameter);
+            if (EqualsIgnoreCase(finalParameter, "base64"))
+            {
+                mediaType.resize(semicolon);
+                TrimAsciiWhitespace(mediaType);
+                base64 = true;
+            }
+        }
+        if (mediaType.empty())
```
bghgary: Parse and serialize the metadata as a MIME type, falling back to `text/plain;charset=US-ASCII` when parsing fails. Returning the trimmed token verbatim makes `data:x,foo` expose `content-type: x` instead of the required fallback.

Polyfills/Fetch/Source/Fetch.cpp
```diff
@@ -397,6 +301,47 @@ namespace Babylon::Polyfills::Internal
...
+                    {
+                        deferred.Reject(Napi::TypeError::New(env, error.what()).Value());
+                        return deferred.Promise();
+                    }
+
+                    if (dataUrl)
+                    {
+                        if (method != UrlLib::UrlMethod::Get || body.has_value())
+                        {
+                            throw std::runtime_error{"fetch: data URLs only support GET requests"};
```
bghgary: Do not reject POST requests or request bodies solely because the URL uses `data:`. The Fetch algorithm resolves this non-HTTP scheme independently of the HTTP method, so `fetch("data:,x", { method: "POST", body: "y" })` should still return the decoded resource.

Polyfills/Fetch/Source/Fetch.cpp
```diff
@@ -397,6 +301,47 @@ namespace Babylon::Polyfills::Internal
...
+                        deferred.Resolve(CreateFetchResponse(
+                            env,
+                            dataUrl->body.data(),
+                            dataUrl->body.size(),
+                            responseHeaders,
+                            200,
+                            "OK",
+                            dataUrl->url,
+                            false));
+                        return deferred.Promise();
```
bghgary: Validate request headers before returning from this fast path. `ApplyRequestHeaders` only runs below it, so invalid names or values on a `data:` request are silently accepted instead of rejecting `fetch`.

PR #217, opened 2026-07-29
Drain deferred V8 N-API finalizers after dispatch

matthargett:
## Summary

Drain deferred V8 N-API finalizers at a safe host boundary, with an 8 ms budget per dispatcher turn.

V8 weak callbacks enqueue non-experimental N-API finalizers for a safe second pass. JsRuntimeHost did not run that pass, so collected wrappers retained native resources until runtime teardown. This is independent of Dawn, WebGPU, and application rendering code.

## Reproduction

The profiling workload uses four packaged VRM files (from https://github.com/PolygonalMind/100Avatars ), loaded through the browser-shaped Blob path, with four thin instances per avatar. It replaces the batch ten times after 180 rendered frames:

```js
const engine = new BABYLON.NativeEngine();
const scene = new BABYLON.Scene(engine);
scene.createDefaultCamera(true, true, true);
scene.createDefaultLight(true);
engine.runRenderLoop(() => scene.render());

const avatarUrls = [
    "app:///Assets/avatar-a.vrm",
    "app:///Assets/avatar-b.vrm",
    "app:///Assets/avatar-c.vrm",
    "app:///Assets/avatar-d.vrm",
];

const matrices = new Float32Array(4 * 16);
for (let index = 0; index < 4; ++index) {
    BABYLON.Matrix.Translation((index - 1.5) * 1.5, 0, 0).copyToArray(matrices, index * 16);
}

const waitFrames = (count) =>
    new Promise((resolve) => {
        const observer = scene.onAfterRenderObservable.add(() => {
            if (--count === 0) {
                scene.onAfterRenderObservable.remove(observer);
                resolve();
            }
        });
    });

(async () => {
    let containers = [];
    for (let cycle = 0; cycle < 10; ++cycle) {
        for (const container of containers) {
            container.dispose();
        }
        containers = [];

        for (const url of avatarUrls) {
            const response = await fetch(url);
            const blob = await response.blob();
            const objectUrl = URL.createObjectURL(blob);
            try {
                const container = await BABYLON.SceneLoader.LoadAssetContainerAsync(
                    "",
                    objectUrl,
                    scene,
                    undefined,
                    ".glb"
                );
                container.addAllToScene();
                for (const mesh of container.meshes) {
                    if (mesh.getTotalVertices() > 0) {
                        mesh.thinInstanceSetBuffer("matrix", matrices, 16, true);
                    }
                }
                containers.push(container);
            } finally {
                URL.revokeObjectURL(objectUrl);
            }
        }

        await waitFrames(180);
    }
})();
```

The measured files were four distinct 6-10 MB VRMs from [PolygonalMind/100Avatars](https://github.com/PolygonalMind/100Avatars/tree/master/100Avatars_200).

## Validation

- Current main reached about 2 GB RSS / 1.9 GB native heap with the deferred queue undrained.
- Draining the queue kept the same workload near 429 MB RSS at 50 seconds.
- Per-turn budgeting reduced the original 109.49 ms maximum drain to short slices. After removing an unrelated synchronous resource flush in the application finalizer, the 10-cycle run measured p95 8.00 ms, p99 8.06 ms, and max 8.36 ms.
- V8 connected suite: 8 tests passed, including deterministic coverage that the queue yields between turns and then drains completely.
- Android HWASan soak: 600,064 transient wrappers completed with bounded memory and no finding.

### DIFF
--- a/Core/Node-API/Source/env_v8.cc
+++ b/Core/Node-API/Source/env_v8.cc
@@ -2,8 +2,15 @@
 #include <napi/js_native_api_types.h>
 #include "js_native_api_v8.h"
 
+#include <chrono>
+
 namespace Napi
 {
+  namespace
+  {
+    constexpr auto FinalizerDrainBudget{std::chrono::milliseconds{8}};
+  }
+
   Env Attach(v8::Local<v8::Context> isolate)
   {
     // second argument is module version
@@ -19,13 +26,20 @@ namespace Napi
   void DrainFinalizers(Env env)
   {
     napi_env env_ptr{env};
+    const auto start = std::chrono::steady_clock::now();
 
-    // Finalizers can mutate this queue, so continue until it is empty.
+    // Finalizers can mutate this queue. Bound each dispatcher turn so a large
+    // batch cannot monopolize the runtime thread.
     while (!env_ptr->pending_finalizers.empty())
     {
       auto* finalizer = *env_ptr->pending_finalizers.begin();
       env_ptr->pending_finalizers.erase(finalizer);
       finalizer->Finalize();
+
+      if (std::chrono::steady_clock::now() - start >= FinalizerDrainBudget)
+      {
+        break;
+      }
     }
   }
 
--- a/Core/Node-API/Source/env_v8.cc
+++ b/Core/Node-API/Source/env_v8.cc
@@ -16,6 +16,19 @@ namespace Napi
     env_ptr->DeleteMe();
   }
 
+  void DrainFinalizers(Env env)
+  {
+    napi_env env_ptr{env};
+
+    // Finalizers can mutate this queue, so continue until it is empty.
+    while (!env_ptr->pending_finalizers.empty())
+    {
+      auto* finalizer = *env_ptr->pending_finalizers.begin();
+      env_ptr->pending_finalizers.erase(finalizer);
+      finalizer->Finalize();
+    }
+  }
+
   v8::Local<v8::Context> GetContext(Env env)
   {
     napi_env env_ptr{env};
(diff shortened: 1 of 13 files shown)


Core/Node-API/Source/env_v8.cc
```diff
@@ -16,6 +16,19 @@ namespace Napi
     env_ptr->DeleteMe();
   }
 
+  void DrainFinalizers(Env env)
+  {
+    napi_env env_ptr{env};
+
+    // Finalizers can mutate this queue, so continue until it is empty.
+    while (!env_ptr->pending_finalizers.empty())
```
matthargett, line 33: hypothetically, this could go on forever in some pathological case with a semi-malicious native module. lmk if you want a safeguard counter check in there.

Core/Node-API/Source/env_v8.cc
```diff
@@ -16,6 +23,26 @@ namespace Napi
...
+
+    // Finalizers can mutate this queue. Bound each dispatcher turn so a large
+    // batch cannot monopolize the runtime thread.
+    while (!env_ptr->pending_finalizers.empty())
+    {
+      auto* finalizer = *env_ptr->pending_finalizers.begin();
+      env_ptr->pending_finalizers.erase(finalizer);
+      finalizer->Finalize();
+
+      if (std::chrono::steady_clock::now() - start >= FinalizerDrainBudget)
```
bghgary: Please ensure a non-empty remainder schedules another dispatcher turn instead of waiting for unrelated work. `blocking_tick` blocks once its captured batch is empty, and an idle probe left 15 of 16 finalizers pending after 200 ms; the test's observation dispatches currently provide the missing wakeups.

PR #10, opened 2023-03-02
Add AbortController and AbortSignal Polyfills

karisme:

PR #218, opened 2026-07-30
Fix napi_get_property_names conformance on JSC, Chakra and QuickJS

### LINKED ISSUE #216: napi_get_property_names: throws on JavaScriptCore; inconsistent enumerability/prototype semantics on Chakra and QuickJS
`napi_get_property_names` behaves differently on all three non-V8 backends, and **throws unconditionally on JavaScriptCore**. Since `Napi::Object::GetPropertyNames()` is a direct wrapper, that method is unusable on JSC for every N-API caller — and JSC is the default engine on macOS and iOS.

## Reference behaviour (V8)

`js_native_api_v8.cc` matches the Node-API specification: *enumerable*, string-keyed properties, **including the prototype chain**.

```cpp
obj->GetPropertyNames(
    context,
    v8::KeyCollectionMode::kIncludePrototypes,
    static_cast<v8::PropertyFilter>(
        v8::PropertyFilter::ONLY_ENUMERABLE |
        v8::PropertyFilter::SKIP_SYMBOLS),
    v8::IndexFilter::kIncludeIndices,
    v8::KeyConversionMode::kConvertToString);
```

## JavaScriptCore — throws

`Core/Node-API/Source/js_native_api_javascriptcore.cc`:

```cpp
CHECK_NAPI(napi_get_named_property(env, object_ctor, "getOwnPropertyNames", &function));
CHECK_NAPI(napi_call_function(env, object_ctor, function, 0, nullptr, result));
```

The `object` parameter is never used. The call passes argc `0` / argv `nullptr`, so it evaluates `Object.getOwnPropertyNames(undefined)`, which throws `TypeError`. Note there is also no `CHECK_ARG(env, object)`.

**Passing the object is necessary but not sufficient.** `Object.getOwnPropertyNames` is own-only and *includes non-enumerable* properties, so it disagrees with V8 on both axes. A conforming JSC implementation needs enumerable properties across the prototype chain — the semantics of `for...in` filtered to string keys.

## Chakra — own-only, includes non-enumerables

`js_native_api_chakra.cc` uses `JsGetOwnPropertyNames`, which is own-only and does not filter to enumerable properties. Wrong on both axes, though it does not throw.

## QuickJS — own-only

`js_native_api_quickjs.cc` uses `JS_GetOwnPropertyNames` with `JS_GPN_STRING_MASK | JS_GPN_ENUM_ONLY`. Enumerability is correct; the prototype chain is missing.

## Summary

| backend | enumerable-only | includes prototype chain | throws |
|---|---|---|---|
| V8 | yes | yes | no |
| JavaScriptCore | n/a | n/a | **yes** |
| Chakra | no | no | no |
| QuickJS | yes | no | no |

## Repro

```js
const proto = { inherited: 1 };
const obj = Object.create(proto);
obj.own = 2;
Object.defineProperty(obj, "hidden", { value: 3, enumerable: false });
// napi_get_property_names(obj) should yield exactly ["own", "inherited"].
```

On JSC this throws instead of returning. On Chakra it yields `["own", "hidden"]`. On QuickJS it yields `["own"]`.

## Impact / workaround

Encountered in Babylon Native while enumerating a plain JS data object (BabylonJS/BabylonNative#1797). The workaround is to fetch `Object.keys` from the global and call it via `napi_call_function`, which behaves consistently across engines for plain data objects:

```cpp
const auto objectCtor = env.Global().Get("Object").As<Napi::Object>();
const auto keys = objectCtor.Get("keys").As<Napi::Function>();
```

That is only equivalent for own-enumerable cases, so it is a local workaround rather than a fix.

bkaradzic-microsoft:
Fixes #216.

## The problem

`napi_get_property_names` should return the enumerable string-keyed properties of an object **and its prototype chain** — the same set a `for...in` loop visits. Only V8 got this right.

| backend | status |
|---|---|
| V8 | correct |
| JavaScriptCore | **always threw** — called `Object.getOwnPropertyNames` with `argc` 0, so the object argument was never passed |
| Chakra | own properties only, and included non-enumerable ones |
| QuickJS | own properties only |
| JSI | **unimplemented stub** that threw `"TODO"` |

## The fix

JSC, Chakra and QuickJS have no native equivalent of V8's prototype-walking key collection. Rather than write three subtly different walks, this adds one shared implementation (`Source/js_native_api_shared.{h,cc}`) built purely on the public `napi_*` API, and has all three delegate to it.

Per prototype level it combines:
- `Object.keys` — what `for...in` reports at that level
- `Object.getOwnPropertyNames` — the shadowing set, so a non-enumerable own property correctly hides a same-named enumerable property further up the chain

JSI keeps its own path — `jsi::Object::getPropertyNames` already has the right semantics, so it just forwards.

Two defects found in review are also fixed:

- **Cyclic prototype chains hung the process.** A `getPrototypeOf` Proxy trap can return an object already on the chain. V8 survives because its walk is recursive and dies with a `RangeError`; this walk is iterative, so it spun forever (reproduced: 56s of CPU, never returned — it would have taken CI down rather than failing a test). The walk now stops at an already-visited level. This is exact, not a bail-out: every level adds its full own-property set to the shadowing set before advancing, so revisiting a level can only re-encounter names that are already shadowed.
- **The returned status disagreed with `napi_get_last_error_info`.** Rejections returned `napi_object_expected` while the recorded error still read `napi_ok`, and the success path left stale errors in place. Both are now handled explicitly at all three call sites.

## Also fixed: `napi_get_prototype` on JavaScriptCore

The shared walk depends on it, and it was broken two ways. It ran the *result* of `JSObjectGetPrototype` through `JSValueToObject` — but at the top of a chain that value is `null`, so it threw instead of reporting the end of the chain, making the chain impossible to walk.

The conversion was simply on the wrong operand: it belongs on the **argument**, which is where V8 puts it. Moving it there also fixes a latent bug on the same line — the argument went to `ToJSObject`, which only `assert`s its input is an object, so a primitive tripped the assert in debug and reinterpreted a non-object as a `JSObjectRef` in release. Now it matches V8: a primitive yields its wrapper's prototype, and only `null`/`undefined` are rejected.

## Testing

13 script tests via a new `napiGetPropertyNames` global, covering own/inherited/shadowed/non-enumerable/symbol/class-method/array cases, equality with `for...in`, and the termination cases (self-referential proxy, two-object cycle, long acyclic chain, throwing trap). Plus a native test asserting the status and recorded error agree.

Verified locally on Chakra (235 passing, 0 failing); all 24 CI legs green.

As a **negative control**, restoring the old Chakra implementation fails 5 of the 9 conformance tests, so the tests are load-bearing. Reverting the termination fix hangs the suite; reverting the `last_error` fix fails the native test on exactly the two assertions it should.

## Notes

- **Hermes' `for...in` doesn't implement the shadowing rule** — it reports an inherited property that a non-enumerable own property should hide. Not fixed here (engine bug, tracked by #219); the test probes for the behaviour at runtime rather than naming engines, and the explicit expected-value assertion still runs everywhere.
- The three proxy-dependent tests are skipped on JavaScriptCore, whose `napi_get_prototype` reads the internal `[[Prototype]]` slot and never runs a trap — so the cycle is invisible there and that backend was never at risk. V8 behaves the same way and says so in a comment. The acyclic-chain test runs everywhere.
- The native error-reporting test covers only the three backends sharing the walk. V8's version is vendored upstream Node code that keeps neither half of the contract; that's upstream's to define.
- `napi_get_all_property_names` is still V8-only. Separate gap, left alone.

### DIFF
--- a/Core/Node-API/Source/js_native_api_javascriptcore.cc
+++ b/Core/Node-API/Source/js_native_api_javascriptcore.cc
@@ -968,9 +968,13 @@ napi_status napi_get_property_names(napi_env env,
   CHECK_ARG(env, object);
   CHECK_ARG(env, result);
 
-  // JavaScriptCore's `JSObjectCopyPropertyNames` walks the prototype chain but
-  // silently drops properties shadowed by a non-enumerable own property, so use
-  // the shared prototype-chain walk instead. It is written against the public
+  // JavaScriptCore's `JSObjectCopyPropertyNames` walks the prototype chain, but
+  // it does not apply the shadowing rule: `JSObject::getPropertyNames` calls
+  // `getOwnPropertyNames` per level with `DontEnumPropertiesMode::Exclude`, so
+  // a non-enumerable own property is never added to the array and so cannot
+  // suppress a same-named enumerable property further up the chain. The
+  // inherited name is reported where `for...in` correctly omits it. Use the
+  // shared prototype-chain walk instead. It is written against the public
   // `napi_*` surface and so cannot reach `napi_set_last_error`; do it here,
   // since `CHECK_NAPI` only propagates the status and the preceding call inside
   // the walk will have cleared the last error. The success path likewise has to
--- a/Core/Node-API/Source/js_native_api_javascriptcore.cc
+++ b/Core/Node-API/Source/js_native_api_javascriptcore.cc
@@ -1,4 +1,5 @@
 #include "js_native_api_javascriptcore.h"
+#include "js_native_api_shared.h"
 #include <algorithm>
 #include <cassert>
 #include <cmath>
@@ -964,13 +965,13 @@ napi_status napi_get_property_names(napi_env env,
                                     napi_value object,
                                     napi_value* result) {
   CHECK_ENV(env);
+  CHECK_ARG(env, object);
   CHECK_ARG(env, result);
 
-  napi_value global{}, object_ctor{}, function{};
-  CHECK_NAPI(napi_get_global(env, &global));
-  CHECK_NAPI(napi_get_named_property(env, global, "Object", &object_ctor));
-  CHECK_NAPI(napi_get_named_property(env, object_ctor, "getOwnPropertyNames", &function));
-  CHECK_NAPI(napi_call_function(env, object_ctor, function, 0, nullptr, result));
+  // JavaScriptCore's `JSObjectCopyPropertyNames` walks the prototype chain but
+  // silently drops properties shadowed by a non-enumerable own property, so use
+  // the shared prototype-chain walk instead.
+  CHECK_NAPI(napi_shared::GetEnumerablePropertyNames(env, object, result));
 
   return napi_ok;
 }
--- a/Core/Node-API/Source/js_native_api_javascriptcore.cc
+++ b/Core/Node-API/Source/js_native_api_javascriptcore.cc
@@ -1345,9 +1345,25 @@ napi_status napi_get_prototype(napi_env env,
   // `JSValueToObject` threw "TypeError: null is not an object" there instead of
   // reporting the end of the chain, which made the chain impossible to walk.
   // V8 likewise returns the raw prototype value.
-  *result = ToNapi(JSObjectGetPrototype(env->context, ToJSObject(env, object)));
+  //
+  // The conversion belongs on the argument rather than the result. V8 coerces
+  // there (`CHECK_TO_OBJECT`), so a primitive yields its wrapper's prototype
+  // and only `null`/`undefined` are rejected. Passing the argument straight to
+  // `ToJSObject` instead would assert in debug and, in release, reinterpret a
+  // non-object `JSValueRef` as a `JSObjectRef` -- so a primitive was undefined
+  // behaviour rather than a status.
+  const JSValueRef value{ToJSValue(object)};
+  if (JSValueIsNull(env->context, value) || JSValueIsUndefined(env->context, value)) {
+    return napi_set_last_error(env, napi_object_expected);
+  }
 
-  return napi_ok;
+  JSValueRef exception{};
+  const JSObjectRef self{JSValueToObject(env->context, value, &exception)};
+  CHECK_JSC(env, exception);
+
+  *result = ToNapi(JSObjectGetPrototype(env->context, self));
+
+  return napi_clear_last_error(env);
 }
 
 napi_status napi_create_object(napi_env env, napi_value* result) {
--- a/Core/Node-API/Source/js_native_api_javascriptcore.cc
+++ b/Core/Node-API/Source/js_native_api_javascriptcore.cc
@@ -1329,13 +1329,16 @@ napi_status napi_get_prototype(napi_env env,
                                napi_value object,
                                napi_value* result) {
   CHECK_ENV(env);
+  CHECK_ARG(env, object);
   CHECK_ARG(env, result);
 
-  JSValueRef exception{};
-  JSObjectRef prototype{JSValueToObject(env->context, JSObjectGetPrototype(env->context, ToJSObject(env, object)), &exception)};
-  CHECK_JSC(env, exception);
+  // `JSObjectGetPrototype` already yields a JSValueRef, and that value is
+  // `null` at the top of a prototype chain. Running it through
+  // `JSValueToObject` threw "TypeError: null is not an object" there instead of
+  // reporting the end of the chain, which made the chain impossible to walk.
+  // V8 likewise returns the raw prototype value.
+  *result = ToNapi(JSObjectGetPrototype(env->context, ToJSObject(env, object)));
 
-  *result = ToNapi(prototype);
   return napi_ok;
 }
 
--- a/Core/Node-API/Source/js_native_api_javascriptcore.cc
+++ b/Core/Node-API/Source/js_native_api_javascriptcore.cc
@@ -970,10 +970,18 @@ napi_status napi_get_property_names(napi_env env,
 
   // JavaScriptCore's `JSObjectCopyPropertyNames` walks the prototype chain but
   // silently drops properties shadowed by a non-enumerable own property, so use
-  // the shared prototype-chain walk instead.
-  CHECK_NAPI(napi_shared::GetEnumerablePropertyNames(env, object, result));
+  // the shared prototype-chain walk instead. It is written against the public
+  // `napi_*` surface and so cannot reach `napi_set_last_error`; do it here,
+  // since `CHECK_NAPI` only propagates the status and the preceding call inside
+  // the walk will have cleared the last error. The success path likewise has to
+  // clear it, so that a rejection recorded by an earlier call does not survive
+  // as the last error of a call that succeeded.
+  const napi_status status{napi_shared::GetEnumerablePropertyNames(env, object, result)};
+  if (status != napi_ok) {
+    return napi_set_last_error(env, status);
+  }
 
-  return napi_ok;
+  return napi_clear_last_error(env);
 }
 
 napi_status napi_set_property(napi_env env,
--- a/Core/Node-API/Source/js_native_api_quickjs.cc
+++ b/Core/Node-API/Source/js_native_api_quickjs.cc
@@ -1397,8 +1397,16 @@ napi_status napi_get_property_names(napi_env env, napi_value object, napi_value*
   CHECK_ARG(env, result);
 
   // `JS_GetOwnPropertyNames` is own-only, so use the shared prototype-chain
-  // walk instead.
-  CHECK_NAPI(napi_shared::GetEnumerablePropertyNames(env, object, result));
+  // walk instead. It is written against the public `napi_*` surface and so
+  // cannot reach `napi_set_last_error`; do it here, since `CHECK_NAPI` only
+  // propagates the status and the preceding call inside the walk will have
+  // cleared the last error. The success path likewise has to clear it, so that
+  // a rejection recorded by an earlier call does not survive as the last error
+  // of a call that succeeded.
+  const napi_status status{napi_shared::GetEnumerablePropertyNames(env, object, result)};
+  if (status != napi_ok) {
+    return napi_set_last_error(env, status);
+  }
 
   napi_clear_last_error(env);
   return napi_ok;
--- a/Core/Node-API/Source/js_native_api_quickjs.cc
+++ b/Core/Node-API/Source/js_native_api_quickjs.cc
@@ -1,4 +1,5 @@
 #include "js_native_api_quickjs.h"
+#include "js_native_api_shared.h"
 #include <napi/js_native_api.h>
 #if defined(__clang__)
 #pragma clang diagnostic push
@@ -1394,27 +1395,11 @@ napi_status napi_get_property_names(napi_env env, napi_value object, napi_value*
   CHECK_ENV(env);
   CHECK_ARG(env, object);
   CHECK_ARG(env, result);
-  
-  JSValue jsObject = ToJSValue(object);
-  
-  JSPropertyEnum* ptab;
-  uint32_t plen;
-  
-  if (JS_GetOwnPropertyNames(env->context, &ptab, &plen, jsObject, 
-      JS_GPN_STRING_MASK | JS_GPN_ENUM_ONLY) < 0) {
-    return napi_set_last_error(env, napi_generic_failure);
-  }
-  
-  JSValue arr = JS_NewArray(env->context);
-  
-  for (uint32_t i = 0; i < plen; i++) {
-    JSValue name = JS_AtomToString(env->context, ptab[i].atom);
-    JS_SetPropertyUint32(env->context, arr, i, name);
-  }
-  
-  JS_FreePropertyEnum(env->context, ptab, plen);
-  
-  *result = FromJSValue(env, arr);
+
+  // `JS_GetOwnPropertyNames` is own-only, so use the shared prototype-chain
+  // walk instead.
+  CHECK_NAPI(napi_shared::GetEnumerablePropertyNames(env, object, result));
+
   napi_clear_last_error(env);
   return napi_ok;
 }
--- a/Core/Node-API/Source/js_native_api_shared.cc
+++ b/Core/Node-API/Source/js_native_api_shared.cc
@@ -1,0 +1,114 @@
+#include "js_native_api_shared.h"
+
+#include <napi/js_native_api.h>
+
+#include <string>
+#include <unordered_set>
+#include <vector>
+
+namespace napi_shared {
+  namespace {
+    #define RETURN_IF_NOT_OK(expression)                    \
+      do {                                                  \
+        const napi_status status__{(expression)};           \
+        if (status__ != napi_ok) {                          \
+          return status__;                                  \
+        }                                                   \
+      } while (0)
+
+    napi_status GetUtf8Value(napi_env env, napi_value value, std::string& result) {
+      size_t length{};
+      RETURN_IF_NOT_OK(napi_get_value_string_utf8(env, value, nullptr, 0, &length));
+
+      std::vector<char> buffer(length + 1);
+      size_t copied{};
+      RETURN_IF_NOT_OK(napi_get_value_string_utf8(env, value, buffer.data(), buffer.size(), &copied));
+
+      result.assign(buffer.data(), copied);
+      return napi_ok;
+    }
+
+    napi_status IsObjectLike(napi_env env, napi_value value, bool& result) {
+      napi_valuetype type{};
+      RETURN_IF_NOT_OK(napi_typeof(env, value, &type));
+      result = (type == napi_object || type == napi_function || type == napi_external);
+      return napi_ok;
+    }
+
+    // Appends every element of the string array `names` to `shadowed`.
+    napi_status AddAll(napi_env env, napi_value names, std::unordered_set<std::string>& shadowed) {
+      uint32_t count{};
+      RETURN_IF_NOT_OK(napi_get_array_length(env, names, &count));
+
+      std::string key{};
+      for (uint32_t index = 0; index < count; ++index) {
+        napi_value name{};
+        RETURN_IF_NOT_OK(napi_get_element(env, names, index, &name));
+        RETURN_IF_NOT_OK(GetUtf8Value(env, name, key));
+        shadowed.insert(std::move(key));
+      }
+
+      return napi_ok;
+    }
+  }
+
+  napi_status GetEnumerablePropertyNames(napi_env env, napi_value object, napi_value* result) {
+    // `Object.keys` reports one level's own enumerable string-keyed properties
+    // in specification order, which is exactly what `for...in` visits at that
+    // level. `Object.getOwnPropertyNames` additionally reports the
+    // non-enumerable ones: `for...in` does not visit those, but they still
+    // shadow same-named properties further up the prototype chain, so they have
+    // to be tracked as well.
+    napi_value global{};
+    napi_value objectConstructor{};
+    napi_value keys{};
+    napi_value getOwnPropertyNames{};
+    RETURN_IF_NOT_OK(napi_get_global(env, &global));
+    RETURN_IF_NOT_OK(napi_get_named_property(env, global, "Object", &objectConstructor));
+    RETURN_IF_NOT_OK(napi_get_named_property(env, objectConstructor, "keys", &keys));
+    RETURN_IF_NOT_OK(napi_get_named_property(env, objectConstructor, "getOwnPropertyNames", &getOwnPropertyNames));
+
+    napi_value names{};
+    RETURN_IF_NOT_OK(napi_create_array(env, &names));
+    uint32_t nameCount{};
+
+    std::unordered_set<std::string> shadowed{};
+    std::string key{};
+
+    napi_value current{};
+    RETURN_IF_NOT_OK(napi_coerce_to_object(env, object, &current));
+
+    while (true) {
+      bool isObjectLike{};
+      RETURN_IF_NOT_OK(IsObjectLike(env, current, isObjectLike));
+      if (!isObjectLike) {
+        break;
+      }
+
+      napi_value ownEnumerableNames{};
+      RETURN_IF_NOT_OK(napi_call_function(env, objectConstructor, keys, 1, &current, &ownEnumerableNames));
+
+      uint32_t ownEnumerableCount{};
+      RETURN_IF_NOT_OK(napi_get_array_length(env, ownEnumerableNames, &ownEnumerableCount));
+      for (uint32_t index = 0; index < ownEnumerableCount; ++index) {
+        napi_value name{};
+        RETURN_IF_NOT_OK(napi_get_element(env, ownEnumerableNames, index, &name));
+        RETURN_IF_NOT_OK(GetUtf8Value(env, name, key));
+        if (shadowed.find(key) == shadowed.end()) {
+          RETURN_IF_NOT_OK(napi_set_element(env, names, nameCount++, name));
+        }
+      }
+
+      napi_value ownNames{};
+      RETURN_IF_NOT_OK(napi_call_function(env, objectConstructor, getOwnPropertyNames, 1, &current, &ownNames));
+      RETURN_IF_NOT_OK(AddAll(env, ownNames, shadowed));
+
+      RETURN_IF_NOT_OK(napi_get_prototype(env, current, &current));
+    }
+
+    *result = names;
+    return napi_ok;
+  }
+
+  #undef RETURN_IF_NOT_OK
+}
--- a/Core/Node-API/Source/js_native_api_shared.cc
+++ b/Core/Node-API/Source/js_native_api_shared.cc
@@ -111,11 +111,18 @@ namespace napi_shared {
         }
       }
 
-      napi_value ownNames{};
-      RETURN_IF_NOT_OK(napi_call_function(env, objectConstructor, getOwnPropertyNames, 1, &current, &ownNames));
-      RETURN_IF_NOT_OK(AddAll(env, ownNames, shadowed));
+      napi_value next{};
+      RETURN_IF_NOT_OK(napi_get_prototype(env, current, &next));
 
-      RETURN_IF_NOT_OK(napi_get_prototype(env, current, &current));
+      bool hasNextLevel{};
+      RETURN_IF_NOT_OK(IsObjectLike(env, next, hasNextLevel));
+      if (hasNextLevel) {
+        napi_value ownNames{};
+        RETURN_IF_NOT_OK(napi_call_function(env, objectConstructor, getOwnPropertyNames, 1, &current, &ownNames));
+        RETURN_IF_NOT_OK(AddAll(env, ownNames, shadowed));
+      }
+
+      current = next;
     }
 
     *result = names;
--- a/Core/Node-API/Source/js_native_api_shared.cc
+++ b/Core/Node-API/Source/js_native_api_shared.cc
@@ -75,6 +75,18 @@ namespace napi_shared {
     std::unordered_set<std::string> shadowed{};
     std::string key{};
 
+    // `ToObject` is what the specification (and the V8 implementation) applies
+    // to the argument, so a primitive is wrapped and its properties reported.
+    // `null` and `undefined` have no wrapper, and V8 reports that as
+    // `napi_object_expected`; check explicitly rather than relying on
+    // `napi_coerce_to_object`, whose behaviour for those two values differs
+    // between engines (QuickJS yields an empty object, JavaScriptCore throws).
+    napi_valuetype type{};
+    RETURN_IF_NOT_OK(napi_typeof(env, object, &type));
+    if (type == napi_null || type == napi_undefined) {
+      return napi_object_expected;
+    }
+
     napi_value current{};
     RETURN_IF_NOT_OK(napi_coerce_to_object(env, object, &current));
 
--- a/Core/Node-API/Source/js_native_api_shared.cc
+++ b/Core/Node-API/Source/js_native_api_shared.cc
@@ -50,6 +50,21 @@ namespace napi_shared {
 
       return napi_ok;
     }
+
+    // Whether `value` is strictly equal to something already in `seen`.
+    napi_status Contains(napi_env env, const std::vector<napi_value>& seen, napi_value value, bool& result) {
+      for (const napi_value candidate : seen) {
+        bool equal{};
+        RETURN_IF_NOT_OK(napi_strict_equals(env, candidate, value, &equal));
+        if (equal) {
+          result = true;
+          return napi_ok;
+        }
+      }
+
+      result = false;
+      return napi_ok;
+    }
   }
 
   napi_status GetEnumerablePropertyNames(napi_env env, napi_value object, napi_value* result) {
@@ -73,6 +88,7 @@ namespace napi_shared {
     uint32_t nameCount{};
 
     std::unordered_set<std::string> shadowed{};
+    std::vector<napi_value> visited{};
     std::string key{};
 
     // `ToObject` is what the specification (and the V8 implementation) applies
@@ -96,6 +112,24 @@ namespace napi_shared {
       if (!isObjectLike) {
         break;
       }
+
+      // A `getPrototypeOf` Proxy trap can return an object that is already on
+      // the chain -- nothing in the specification forbids it, so
+      // `Object.getPrototypeOf(p) === p` is reachable from script -- which
+      // makes this walk cyclic. V8 recurses and so terminates with a
+      // `RangeError`; this loop is iterative and would spin forever.
+      //
+      // Stopping at the repeat is exact rather than a bail-out: every level
+      // adds its own property names to `shadowed` before the walk continues,
+      // so a level visited a second time can only re-encounter names that are
+      // already shadowed. Breaking here therefore yields the same result the
+      // non-terminating walk converges on.
+      bool alreadyVisited{};
+      RETURN_IF_NOT_OK(Contains(env, visited, current, alreadyVisited));
+      if (alreadyVisited) {
+        break;
+      }
+      visited.push_back(current);
 
       napi_value ownEnumerableNames{};
       RETURN_IF_NOT_OK(napi_call_function(env, objectConstructor, keys, 1, &current, &ownEnumerableNames));
--- a/Core/Node-API/Source/js_native_api_shared.h
+++ b/Core/Node-API/Source/js_native_api_shared.h
@@ -1,0 +1,22 @@
+#pragma once
+
+#include <napi/js_native_api_types.h>
+
+// Engine-agnostic pieces of the Node-API surface, implemented purely in terms
+// of the public `napi_*` entry points so that every backend behaves the same.
+// Backends whose engine offers a faithful native equivalent should keep using
+// it; these helpers exist for the ones that do not.
+namespace napi_shared {
+  // Implements `napi_get_property_names` semantics: the names of all
+  // enumerable string-keyed properties of `object` and of its prototype chain,
+  // as an array of strings, matching a `for...in` enumeration.
+  //
+  // V8 gets this from a single `GetPropertyNames` call configured with
+  // `kIncludePrototypes | ONLY_ENUMERABLE | SKIP_SYMBOLS`. JavaScriptCore,
+  // ChakraCore and QuickJS have no equivalent, so this walks the prototype
+  // chain explicitly. See https://github.com/BabylonJS/JsRuntimeHost/issues/216.
+  //
+  // `object` is coerced with `napi_coerce_to_object`, as V8's `CHECK_TO_OBJECT`
+  // does. Callers are expected to have already validated `env` and `result`.
+  napi_status GetEnumerablePropertyNames(napi_env env, napi_value object, napi_value* result);
+}
--- a/Core/Node-API/Source/js_native_api_shared.h
+++ b/Core/Node-API/Source/js_native_api_shared.h
@@ -13,7 +13,7 @@ namespace napi_shared {
   //
   // V8 gets this from a single `GetPropertyNames` call configured with
   // `kIncludePrototypes | ONLY_ENUMERABLE | SKIP_SYMBOLS`. JavaScriptCore,
-  // ChakraCore and QuickJS have no equivalent, so this walks the prototype
+  // Chakra and QuickJS have no equivalent, so this walks the prototype
   // chain explicitly. See https://github.com/BabylonJS/JsRuntimeHost/issues/216.
   //
   // `object` is coerced with `napi_coerce_to_object`, as V8's `CHECK_TO_OBJECT`
--- a/Tests/UnitTests/Scripts/tests.ts
+++ b/Tests/UnitTests/Scripts/tests.ts
@@ -1637,6 +1637,14 @@ describe("napi_get_property_names (#216)", function () {
         return keys;
     }
 
+    // Some engines' own `for...in` does not implement the shadowing rule that
+    // the tests below rely on -- Hermes reports an inherited property that a
+    // non-enumerable own property is supposed to hide. Probe for that rather
+    // than name engines, and only use `for...in` as an oracle where it holds.
+    const shadowingProbe = Object.create({ probe: 1 });
+    Object.defineProperty(shadowingProbe, "probe", { value: 2, enumerable: false });
+    const forInHonoursShadowing = forIn(shadowingProbe).length === 0;
+
     it("returns own enumerable string keys", function () {
         expect(napiGetPropertyNames({ a: 1, b: 2 })).to.deep.equal(["a", "b"]);
     });
@@ -1701,8 +1709,10 @@ describe("napi_get_property_names (#216)", function () {
         object[Symbol("s")] = 4;
         Object.defineProperty(object, "deep", { value: 5, enumerable: false });
 
-        expect(napiGetPropertyNames(object)).to.deep.equal(forIn(object));
         expect(napiGetPropertyNames(object)).to.deep.equal(["own", "middle"]);
+        if (forInHonoursShadowing) {
+            expect(napiGetPropertyNames(object)).to.deep.equal(forIn(object));
+        }
     });
 });
 
--- a/Tests/UnitTests/Scripts/tests.ts
+++ b/Tests/UnitTests/Scripts/tests.ts
@@ -1784,16 +1784,29 @@ describe("napi_get_property_names (#216)", function () {
     const usesSharedWalk = napiEngine === "Chakra" || napiEngine === "QuickJS" || napiEngine === "JavaScriptCore";
     const describeCycles = usesSharedWalk ? describe : describe.skip;
 
+    // A cycle can only be built with a `getPrototypeOf` trap, so observing one
+    // additionally requires `napi_get_prototype` to consult that trap.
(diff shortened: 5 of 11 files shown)


Tests/UnitTests/Scripts/tests.ts
```diff
@@ -1621,6 +1622,91 @@ describe("napi class prototype isolation (#172)", function () {
...
+        const keys: string[] = [];
+        for (const key in object) {
+            keys.push(key);
+        }
+        return keys;
+    }
+
+    it("returns own enumerable string keys", function () {
+        expect(napiGetPropertyNames({ a: 1, b: 2 })).to.deep.equal(["a", "b"]);
+    });
```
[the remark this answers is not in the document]
bkaradzic-microsoft: Fair point — that was untested, and you're right that the harness structurally could not test it: `Napi::Object::GetPropertyNames` can only be called on an already-constructed `Napi::Object`, so the coercion path was unreachable from C++.

Fixed in c3def88 by exposing the C entry point directly as `napiGetPropertyNamesRaw` and passing the raw value through, plus six tests.

One caveat on your suggestion to just call the C API in the shared harness: `Core/Node-API-JSI` implements the `Napi::` C++ surface directly on top of JSI and has no `js_native_api_*.cc` at all, so there is no C API to call there. The new global is therefore compiled out under the existing `JSRUNTIMEHOST_NAPI_ENGINE_JSI` define and the coercion tests skip themselves on that backend (225 passing / 6 pending) while running everywhere else.

Writing the tests also turned up a real divergence for `null` and `undefined`, which have no object wrapper: V8 reports `napi_object_expected`, QuickJS's `napi_coerce_to_object` uses `Object(value)` and cheerfully returns an empty object, and JavaScriptCore's throws. The shared implementation now checks `napi_typeof` explicitly so all three match V8.

Local results: ChakraCore 231 passing / 0 failing, QuickJS 231 / 0, JSI 225 / 0 + 6 pending. Negative control: with the null/undefined check removed, QuickJS fails exactly the two tests that cover it.

Tests/UnitTests/Shared/Shared.cpp
```diff
@@ -100,6 +100,16 @@ TEST(JavaScript, All)
...
 
         env.Global().Set("hostPlatform", Napi::Value::From(env, JSRUNTIMEHOST_PLATFORM));
+
+        // Exposes napi_get_property_names, via its C++ wrapper, so that the
+        // script tests can compare it against `for...in`. See
+        // https://github.com/BabylonJS/JsRuntimeHost/issues/216.
+        auto getPropertyNamesCallback = Napi::Function::New(
+            env, [](const Napi::CallbackInfo& info) -> Napi::Value {
+                return info[0].As<Napi::Object>().GetPropertyNames();
+            },
```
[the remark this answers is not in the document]
bkaradzic-microsoft: Fair point — that was untested, and you're right that the harness structurally could not test it: `Napi::Object::GetPropertyNames` can only be called on an already-constructed `Napi::Object`, so the coercion path was unreachable from C++.

Fixed in c3def88 by exposing the C entry point directly as `napiGetPropertyNamesRaw` and passing the raw value through, plus six tests.

One caveat on your suggestion to just call the C API in the shared harness: `Core/Node-API-JSI` implements the `Napi::` C++ surface directly on top of JSI and has no `js_native_api_*.cc` at all, so there is no C API to call there. The new global is therefore compiled out under the existing `JSRUNTIMEHOST_NAPI_ENGINE_JSI` define and the coercion tests skip themselves on that backend (225 passing / 6 pending) while running everywhere else.

Writing the tests also turned up a real divergence for `null` and `undefined`, which have no object wrapper: V8 reports `napi_object_expected`, QuickJS's `napi_coerce_to_object` uses `Object(value)` and cheerfully returns an empty object, and JavaScriptCore's throws. The shared implementation now checks `napi_typeof` explicitly so all three match V8.

Local results: ChakraCore 231 passing / 0 failing, QuickJS 231 / 0, JSI 225 / 0 + 6 pending. Negative control: with the null/undefined check removed, QuickJS fails exactly the two tests that cover it.

Core/Node-API/Source/js_native_api_shared.h
```diff
@@ -0,0 +1,22 @@
...
+// it; these helpers exist for the ones that do not.
+namespace napi_shared {
+  // Implements `napi_get_property_names` semantics: the names of all
+  // enumerable string-keyed properties of `object` and of its prototype chain,
+  // as an array of strings, matching a `for...in` enumeration.
+  //
+  // V8 gets this from a single `GetPropertyNames` call configured with
+  // `kIncludePrototypes | ONLY_ENUMERABLE | SKIP_SYMBOLS`. JavaScriptCore,
+  // ChakraCore and QuickJS have no equivalent, so this walks the prototype
+  // chain explicitly. See https://github.com/BabylonJS/JsRuntimeHost/issues/216.
```
bghgary, line 17: [Reviewed by Copilot on behalf of @bghgary]

This is Chakra, not ChakraCore - the backend links `chakrart.lib` and includes `<jsrt.h>`, and there is no ChakraCore in this repo. Same substitution in the PR title, the description, and `Tests/UnitTests/Scripts/tests.ts`.

```suggestion
  // V8 gets this from a single `GetPropertyNames` call configured with
  // `kIncludePrototypes | ONLY_ENUMERABLE | SKIP_SYMBOLS`. JavaScriptCore,
  // Chakra and QuickJS have no equivalent, so this walks the prototype
  // chain explicitly. See https://github.com/BabylonJS/JsRuntimeHost/issues/216.
```

Tests/UnitTests/Scripts/tests.ts
```diff
@@ -1621,6 +1624,145 @@ describe("napi class prototype isolation (#172)", function () {
...
+    // Node-API to call.
+    const describeCoercion = typeof napiGetPropertyNamesRaw === "function" ? describe : describe.skip;
+
+    describeCoercion("argument coercion", function () {
+        // Hermes' Node-API is not implemented in this repository -- it comes from
+        // the Hermes dependency itself -- and it rejects primitives outright
+        // rather than applying ToObject. Hermes is an experimental engine here,
+        // so assert the specified behaviour everywhere it is ours to control and
+        // skip the wrapping cases on Hermes rather than weakening them.
+        const describePrimitives = napiEngine === "Hermes" ? describe.skip : describe;
```
bghgary, line 1736: [Reviewed by Copilot on behalf of @bghgary]

This skip is permanent and silent - if Hermes ever adds the coercion, nothing notices and the cases stay skipped.

Asserting the current behaviour instead turns it into a tripwire that fires when the Hermes pin is bumped:

```ts
const describePrimitives = napiEngine === "Hermes" ? describe.skip : describe;

if (napiEngine === "Hermes") {
    it("rejects primitives instead of applying ToObject (upstream gap)", function () {
        expect(() => napiGetPropertyNamesRaw("ab")).to.throw();
    });
}
```

https://github.com/BabylonJS/JsRuntimeHost/issues/219 tracks this gap and the `for...in` one that `forInHonoursShadowing` handles. Worth linking from both.

Core/Node-API/Source/js_native_api_shared.cc
```diff
@@ -0,0 +1,126 @@
...
+        if (shadowed.find(key) == shadowed.end()) {
+          RETURN_IF_NOT_OK(napi_set_element(env, names, nameCount++, name));
+        }
+      }
+
+      napi_value ownNames{};
+      RETURN_IF_NOT_OK(napi_call_function(env, objectConstructor, getOwnPropertyNames, 1, &current, &ownNames));
+      RETURN_IF_NOT_OK(AddAll(env, ownNames, shadowed));
+
+      RETURN_IF_NOT_OK(napi_get_prototype(env, current, &current));
```
bghgary: [Reviewed by Copilot on behalf of @bghgary]

`shadowed` is rebuilt at the last level too, where nothing can consult it - for `{a: 1, b: 2}` that is 12 of 16 name materializations wasted on `Object.prototype`, each costing two `napi_get_value_string_utf8` calls and an allocation.

Fetching the prototype first and only populating `shadowed` when the walk continues avoids it:

```cpp
napi_value next{};
RETURN_IF_NOT_OK(napi_get_prototype(env, current, &next));

bool hasNextLevel{};
RETURN_IF_NOT_OK(IsObjectLike(env, next, hasNextLevel));
if (hasNextLevel) {
    napi_value ownNames{};
    RETURN_IF_NOT_OK(napi_call_function(env, objectConstructor, getOwnPropertyNames, 1, &current, &ownNames));
    RETURN_IF_NOT_OK(AddAll(env, ownNames, shadowed));
}

current = next;
```

Nit.

bghgary (commented):

LGTM. Comments inline.

Core/Node-API/Source/js_native_api_quickjs.cc
```diff
@@ -1394,27 +1395,11 @@ napi_status napi_get_property_names(napi_env env, napi_value object, napi_value*
...
-    JS_SetPropertyUint32(env->context, arr, i, name);
-  }
-  
-  JS_FreePropertyEnum(env->context, ptab, plen);
-  
-  *result = FromJSValue(env, arr);
+
+  // `JS_GetOwnPropertyNames` is own-only, so use the shared prototype-chain
+  // walk instead.
+  CHECK_NAPI(napi_shared::GetEnumerablePropertyNames(env, object, result));
```
matthargett: GetEnumerablePropertyNames returns napi_object_expected directly for null/undefined without setting env->last_error. This caller (and others in this PR's commits) uses CHECK_NAPI,and that contract assumes the callee already set last_error. On QuickJS the preceding napi_typeof in the shared macro clears the last error, so the returned status and napi_get_last_error_info()->error_code disagree. I'd recommend adding a unit test for both of these gaps, as the e2e failure with only one of the JS runtimes would be pretty subtle.

Core/Node-API/Source/js_native_api_shared.cc
```diff
@@ -0,0 +1,126 @@
...
+    // non-enumerable ones: `for...in` does not visit those, but they still
+    // shadow same-named properties further up the prototype chain, so they have
+    // to be tracked as well.
+    napi_value global{};
+    napi_value objectConstructor{};
+    napi_value keys{};
+    napi_value getOwnPropertyNames{};
+    RETURN_IF_NOT_OK(napi_get_global(env, &global));
+    RETURN_IF_NOT_OK(napi_get_named_property(env, global, "Object", &objectConstructor));
+    RETURN_IF_NOT_OK(napi_get_named_property(env, objectConstructor, "keys", &keys));
```
matthargett, line 83: grabbing global.Object.keys and Object.getOwnPropertyNames on every call allows user monkey-patching to change native Node-API behavior. if a BabylonNative app will be executing user-generated code, this is an attack surface. PR #116 already established the opposite invariant by capturing canonical Function.prototype.call. is there a reason not to use protected pristine intrinsics or engine-native key enumeration beyond sharing code? To pin this, I suggest adding a poisoned-intrinsics regression integration test. If you all have an internal fuzzer harness, you should ask why it didn't find/highlight this.

Core/Node-API/Source/js_native_api_quickjs.cc
matthargett: The prototype loop needs cycle detection. A Proxy whose getPrototypeOf trap returns itself currently causes an infinite native loop. A throwing proxy trap should also preserve the original pending exception.

Core/Node-API/Source/js_native_api_shared.cc
bkaradzic-microsoft: Applied in cbe9c16: the walk now fetches the prototype first and only materializes/adds own names when there is another object-like level that can consult the shadow set.

Tests/UnitTests/Scripts/tests.ts
bkaradzic-microsoft: Added the Hermes-specific rejection assertion as a tripwire in cbe9c16, while keeping the conformance cases skipped there. I also linked #219 from both the primitive-coercion and for-in shadowing notes.

Core/Node-API/Source/js_native_api_shared.h
bkaradzic-microsoft: Updated the source/test terminology and the PR title and description from ChakraCore to Chakra in cbe9c16. Thanks for catching that.

matthargett (commented):
thanks for coming at N-API! hopefully this means you can review my existing work in this area :D  a few review notes:
* This does not duplicate the implementation work in #189, #217, rebeckerspecialties/JsRuntimeHost#8, or #10 but it *does overlap* with PR #116’s already-vendored test_object conformance case, although that directory is not currently enabled.

* Please use PR #189’s JSR_NAPI_ENGINE_* definition family rather than adding a parallel engine-name macro. The current merge of this PR branch against my N-API stack has one Android CMake conflict for exactly this reason.

* Make a focused v1-compatible NodeApi/CTS addon  durable ABI test, retaining this PR’s additional shadowing, coercion, poisoned-intrinsic, proxy-cycle, and JSI cases.

* Please design the shared collector so it can support v6 napi_get_all_property_names, or link a required follow-up; the combined v7 stack (#189 ) otherwise still exposes a non-V8 symbol gap.


a couple of other security things to guard against:
* rethrow the pending value in the raw test bridge instead of clearing and replacing it with a generic error
* avoid UTF-8 strings as property-key identity because distinct lone-surrogate keys may collapse

Core/Node-API/Source/js_native_api_quickjs.cc
bkaradzic-microsoft: You are right, and this was the most serious of the three. Fixed in 0d7ab06.

I confirmed it is reachable rather than theoretical. Nothing in the specification forbids a `getPrototypeOf` trap from returning the proxy itself -- the invariants on that trap constrain only the non-extensible case -- so `Object.getPrototypeOf(p) === p` evaluates to `true` in Node today. V8 survives it because its walk is recursive and dies on the stack with a `RangeError`; this walk is iterative, so it spun. I reproduced it against a real build before changing anything: the test process burned 56 seconds of CPU and never returned, and I had to kill it. Worth noting that mocha's 5s timeout cannot preempt this, because control never re-enters the engine -- the hang would have taken CI down, not failed a test.

One point I want to be explicit about, since it is the part that makes the fix cheap: stopping at the first repeated level is **exact**, not a bail-out. Every level adds its full `Object.getOwnPropertyNames` set to `shadowed` before the walk advances, and in a cycle that branch is always taken because the next value is object-like. So a level reached a second time can only re-encounter names that are already shadowed. Breaking there yields precisely the fixed point the non-terminating walk converges on -- no result is lost, and no result differs from what an infinitely patient walk would produce. The comment in the source says this so the next reader does not have to re-derive it.

Four script tests cover it: a self-referential proxy, a two-object cycle, a long acyclic chain to show the termination check does not truncate legitimate chains, and a throwing `getPrototypeOf` trap, which propagates as a pending exception rather than hanging. Reverting the fix hangs the suite.

bkaradzic-microsoft: Correct, fixed in 0d7ab06.

The mechanism, for the record, since it is a little indirect: the shared walk is written against the public `napi_*` surface, so it has no way to reach `napi_set_last_error`. `CHECK_NAPI` (`js_native_api_quickjs.h:111`) only propagates the status -- the house contract is that the callee already recorded it -- and the `napi_typeof` the walk performs immediately before the rejection calls `napi_clear_last_error` on its way out (`js_native_api_quickjs.cc:596`). So a caller saw `napi_object_expected` returned while `napi_get_last_error_info()->error_code` still read `napi_ok`.

Writing the test turned up the mirror-image problem, which I had not expected: the *success* path had no clear either, so a successful `napi_get_property_names` left whatever error a previous call had recorded in place. Both directions are now handled explicitly at all three call sites.

Covered by a native regression test, `NodeApi.GetPropertyNamesReportsLastErrorConsistently`, which asserts the returned status and the recorded error agree for `null` and `undefined` and that the success path clears. I ran it against a reverted build as a negative control -- it fails on exactly the two assertions it should.

Core/Node-API/Source/js_native_api_shared.cc
bkaradzic-microsoft: I looked into this one carefully and I am going to push back, though not on the premise.

The premise is sound: script can redefine `Object.keys`, and a native caller of `napi_get_property_names` would then observe the redefined behaviour, where on V8 it would not. That is a real difference.

Where I disagree is that it is a property of this change. `js_native_api_javascriptcore.cc:1073-1075` already resolves `global -> "Object" -> "hasOwnProperty"` at runtime, and has for as long as `napi_has_own_property` has existed on that backend. This PR did not introduce the pattern, and fixing it here would leave the identical exposure one function away, which is the sort of half-measure that reads as fixed and is not.

Two of the supporting points do not hold up. PR #116 has not "established the opposite invariant" -- it is still open and unmerged, so there is nothing to be consistent with yet. And "use engine-native key enumeration instead" is not available on JavaScriptCore: its public C API offers only `JSObjectCopyPropertyNames`, which walks the prototype chain itself and silently drops properties shadowed by a non-enumerable own property. That is precisely the bug #216 reported and precisely why this PR does the walk in shared code rather than per-engine. Taking the native route would mean re-forking the three implementations this change just unified, in order to trade a documented, spec-visible correctness bug for a hardening property that the surrounding code does not have anyway.

So: worth doing, worth doing repo-wide against the whole intrinsic-dependent surface, and worth doing with a considered mechanism -- a per-env snapshot of the intrinsics taken at init, rather than ad-hoc lookups. That is a different change from this one, and I would rather it be scoped and reviewed on its own than smuggled in here. I am deliberately not filing an issue for it; if the team wants it, it should be prioritised as its own piece of work rather than parked in the backlog.

matthargett: TL;DR: please merge #116 first, then this PR, then we/I can do a hardening pass across the repo to prevent prototype injection attacks.

ok, I agree that the repo-wide pristine-intrinsic hardening can be handled as a separate PR, but I want to clarify two points after some cross-checking rather than leave misunderstandings for other folks/agents reading.

First, you are correct that #116 has not merged. “Established” was imprecise wording on my part: I meant that #116 has already implemented the invariant I believe this project should adopt, and that I have deployed in my N-API modules in my BabylonNative app, not that the invariant is already present on main. The existing PR captures the canonical Function.prototype.call once during environment initialization so that user mutation cannot alter napi_call_function behavior, and it introduces the portable Node-API conformance harness. Doing work in N-APi without at least merging the conformance suite PR first will probably yield more friction as time goes on -- unless you and the other project maintainers are purposefully deviating from Node API conformance?

I checked the JavaScriptCore claim because I was pretty sure that statement was incorrect. Looking at Apple’s public SDK (rather than WebKit SPI) and Bun's fork of JSC, [JSObjectCopyPropertyNames](https://developer.apple.com/documentation/javascriptcore/jsobjectcopypropertynames%28_%3A_%3A%29) and [JSObjectGetPrototype](https://developer.apple.com/documentation/javascriptcore/jsobjectgetprototype%28_%3A_%3A%29) are both public and available at my deployment floors—Apple currently lists iOS 16+ and macOS 10.5+. Looking at recent adjustments to CI, it looks like Microsoft's deployment floors are higher than that even?

I *do* see that even the current public JSC API has no configurable own/prototype, enumerable, string/symbol, descriptor-aware enumeration operation equivalent to V8’s key-collection modes. I also verified the behavior against shipping JSC: given an enumerable inherited property shadowed by a non-enumerable own property, JSObjectCopyPropertyNames returns the inherited name while for...in correctly omits it. Therefore, it is not a conforming replacement for the shared walk. The shadowing wording above is slightly ambiguous: JSC does not drop the shadowed inherited property; it fails to let the non-enumerable own property suppress that inherited name. Also, #216 reported the existing JSC backend’s unconditional throw and the cross-backend prototype/enumerability differences; this shadowing case is useful additional conformance coverage discovered while fixing it.

sorry for the wall of text, but you/agent got some things wrong and there were some nuances that are important to get full understanding on if we're going to do a good job of supporting N-API across the diversity of open souce JS VMs that's been expanded over the last several months (QuickJS, Hermes, etc).

Core/Node-API/Source/js_native_api_javascriptcore.cc
```diff
@@ -1328,13 +1337,16 @@ napi_status napi_get_prototype(napi_env env,
                                napi_value object,
                                napi_value* result) {
   CHECK_ENV(env);
+  CHECK_ARG(env, object);
   CHECK_ARG(env, result);
```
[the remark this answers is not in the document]
bkaradzic-microsoft: Good catch — the mechanism is exactly as described, and it is fixed in fdd36a9, though not quite the way suggested.

`ToJSObject` only `assert`s that its input is an object, so a primitive tripped the assert in debug and, in release, reinterpreted a non-object `JSValueRef` as a `JSObjectRef` before handing it to `JSObjectGetPrototype`. That is undefined behaviour rather than a status, so it is worth closing even though nothing in this repository can reach it: the shared prototype walk is the only caller, and it recurses only into values it has already established are object-like.

I did not return `napi_object_expected` for every non-object, because that would diverge from V8. V8's `napi_get_prototype` uses `CHECK_TO_OBJECT`, which *coerces* — `napi_get_prototype(5)` returns `Number.prototype` with `napi_ok`, and only `null` and `undefined` are rejected. The fix matches that.

The neater part is that this function already performed the conversion; it was simply on the wrong operand. It ran the *result* of `JSObjectGetPrototype` through `JSValueToObject`, which is what made the top of a chain throw `TypeError: null is not an object` and made the chain impossible to walk. Moving it to the argument fixes both defects with one change.

For the record, the two earlier comments on this PR (`3679902660`, `3679902688`) were addressed in c3def88, which switched the test helper to call the C API directly with the original argument and added the primitive-coercion cases.

Core/Node-API/Source/js_native_api_shared.cc
bkaradzic-microsoft: You're right on the shadowing point, and I've corrected it in 26a4a99 — thanks for pushing on it.

I checked it against WebKit rather than just taking the correction: `JSObject::getPropertyNames` walks the chain calling `getOwnPropertyNames` per level with `DontEnumPropertiesMode::Exclude`. Because non-enumerable own properties are never added to the array in that mode, they cannot suppress a same-named enumerable property further up — so the inherited name survives where `for...in` correctly omits it. That is the opposite of what my comment said, and "silently drops properties shadowed by a non-enumerable own property" described the *correct* behaviour as if it were the bug. Fixed in the source comment and in the PR description, with the mechanism spelled out so the next reader doesn't have to re-derive it. The conclusion and the code are unchanged. This case is covered by the `omits an inherited property shadowed by a non-enumerable own property` test.

Two smaller clarifications, in the same spirit of not leaving misunderstandings behind:

I didn't claim `JSObjectCopyPropertyNames` or `JSObjectGetPrototype` are unavailable — what I said was that the public C API has no own-only enumeration, and offers only `JSObjectCopyPropertyNames`, which chain-walks. That phrasing presupposes it's available; it's the *own-only* capability that's missing. You reached the same conclusion independently ("no configurable own/prototype, enumerable, string/symbol, descriptor-aware enumeration operation equivalent to V8's key-collection modes"), so I think we agree on the substance and this was just my wording being read more strongly than intended.

And no, there's no deliberate deviation from Node-API conformance — this PR is a conformance fix, and CI now covers the shadowing rule, symbol exclusion, coercion, prototype-chain ordering and `for...in` equivalence across all six backends. Thank you for the review; the cycle and `last_error` findings were both real and both worth the round trip.

Merge order for #116 is bghgary's call, not mine, so I'll leave that to him. I have no objection to it landing first; I'd just rather not block a fix for a reported bug on it. The two don't conflict — nothing here touches `napi_call_function` or intrinsic capture — so whichever order they land in, the rebase is trivial.

Core/Node-API/Source/js_native_api_shared.cc
```diff
@@ -0,0 +1,167 @@
...
+        napi_value name{};
+        RETURN_IF_NOT_OK(napi_get_element(env, ownEnumerableNames, index, &name));
+        RETURN_IF_NOT_OK(GetUtf8Value(env, name, key));
+        if (shadowed.find(key) == shadowed.end()) {
+          RETURN_IF_NOT_OK(napi_set_element(env, names, nameCount++, name));
+        }
+      }
+
+      napi_value next{};
+      RETURN_IF_NOT_OK(napi_get_prototype(env, current, &next));
```
bghgary: Please preserve `for...in`-equivalent Proxy prototype behavior on JSC. This new walk delegates each step to `napi_get_prototype`, whose JSC implementation bypasses `getPrototypeOf` traps, so trapped chains are invisible and the new tests skip that backend.

Core/Node-API/Source/js_native_api_shared.cc
```diff
@@ -0,0 +1,167 @@
...
+      // `RangeError`; this loop is iterative and would spin forever.
+      //
+      // Stopping at the repeat is exact rather than a bail-out: every level
+      // adds its own property names to `shadowed` before the walk continues,
+      // so a level visited a second time can only re-encounter names that are
+      // already shadowed. Breaking here therefore yields the same result the
+      // non-terminating walk converges on.
+      bool alreadyVisited{};
+      RETURN_IF_NOT_OK(Contains(env, visited, current, alreadyVisited));
+      if (alreadyVisited) {
```
bghgary: Please preserve V8's failure behavior for cyclic Proxy prototype chains instead of returning a finite result. V8's `napi_get_property_names` and `for...in` both terminate these cycles with `RangeError`, while this branch silently succeeds and the new cycle tests require that incompatible behavior.

Core/Node-API/Source/js_native_api_shared.cc
```diff
@@ -0,0 +1,167 @@
...
+
+      napi_value ownEnumerableNames{};
+      RETURN_IF_NOT_OK(napi_call_function(env, objectConstructor, keys, 1, &current, &ownEnumerableNames));
+
+      uint32_t ownEnumerableCount{};
+      RETURN_IF_NOT_OK(napi_get_array_length(env, ownEnumerableNames, &ownEnumerableCount));
+      for (uint32_t index = 0; index < ownEnumerableCount; ++index) {
+        napi_value name{};
+        RETURN_IF_NOT_OK(napi_get_element(env, ownEnumerableNames, index, &name));
+        RETURN_IF_NOT_OK(GetUtf8Value(env, name, key));
```
bghgary: Please compare property names with JavaScript string identity rather than their UTF-8 bytes. Distinct lone-surrogate names can encode to the same replacement bytes, so one can incorrectly shadow the other.

Core/Node-API/Source/js_native_api_shared.cc
```diff
@@ -0,0 +1,167 @@
...
+      }
+
+      napi_value next{};
+      RETURN_IF_NOT_OK(napi_get_prototype(env, current, &next));
+
+      bool hasNextLevel{};
+      RETURN_IF_NOT_OK(IsObjectLike(env, next, hasNextLevel));
+      if (hasNextLevel) {
+        napi_value ownNames{};
+        RETURN_IF_NOT_OK(napi_call_function(env, objectConstructor, getOwnPropertyNames, 1, &current, &ownNames));
```
bghgary: Please derive enumerable names and shadowing names from one `ownKeys` snapshot per level. Calling `Object.keys` and `Object.getOwnPropertyNames` separately invokes a Proxy's `ownKeys` trap twice, so a stateful or throwing second trap can change or fail an enumeration that `for...in` performs once.

Core/Node-API/Source/js_native_api_shared.cc
bghgary: Please capture pristine intrinsics instead of looking up `Object.keys` and `Object.getOwnPropertyNames` on every call. Script can replace either function and forge or break `napi_get_property_names`, unlike the engine-native implementations this replaces.

matthargett: while I agree that the repo-wide pristine-intrinsic hardening can be handled as a separate PR, I want to clarify a couple points rather than leave the statements above as written.

First, you are correct that #116 has not merged. “Established” was assumptive wording on my part: I meant that #116 has already implemented the invariant I believe BabylonNative should adopt, not that the invariant is already present on main. It captures the canonical Function.prototype.call once during environment initialization so that user mutation cannot alter napi_call_function behavior, and it introduces the portable Node-API conformance harness.

My requested landing order is that #116 and its prerequisites merge before #218. An open prerequisite is still relevant when deciding the architecture and ordering of what comes after. its being unmerged is not a reason to establish the opposite pattern on main first.

Second, I double checked the JavaScriptCore claim against Apple’s public frameworks rather than WebKit SPI. [JSObjectCopyPropertyNames](https://developer.apple.com/documentation/javascriptcore/jsobjectcopypropertynames%28_%3A_%3A%29) and [JSObjectGetPrototype](https://developer.apple.com/documentation/javascriptcore/jsobjectgetprototype%28_%3A_%3A%29) are both public and available at our minimum OS targets—Apple currently lists iOS 16+ and macOS 10.5+, so iOS 18 and macOS 12 are covered.

that said, I agree with your overall conclusion: even the current public JSC API has no configurable own/prototype, enumerable, string/symbol, descriptor-aware enumeration operation equivalent to V8’s key-collection modes. I also verified the behavior against shipping JSC on my local devices: given an enumerable inherited property shadowed by a non-enumerable own property, JSObjectCopyPropertyNames returns the inherited name while `for...in` correctly omits it. so, annoyingly, it is not a conforming replacement for the shared walk.

The shadowing wording above is a little ambiguous: JSC does not drop the shadowed inherited property; it fails to let the non-enumerable own property suppress that inherited name. on top of that, #216 reported the existing JSC backend’s unconditional throw and the cross-backend prototype/enumerability differences; this shadowing case is useful additional conformance coverage discovered while fixing it.

Engine-native enumeration was one proposed alternative, not mandaory. Rejecting that alternative on JSC shouldnt require three forked implementations: the shared walk can remain shared while obtaining Object.keys and Object.getOwnPropertyNames from per-environment pristine snapshots in a separate hardening PR.

lastly (phew), the existing napi_has_own_property implementation is not a good behavioral precedent IMO. In addition to resolving a mutable intrinsic, it invokes hasOwnProperty with zero arguments and never uses the supplied object or key, making it a separate conformance defect. I think that strengthens the case for addressing the whole intrinsic-dependent surface together.

I am okay with keeping #218 focused on the correctness fix, provided the hardening remains explicit follow-up work. I would still prefer #116 to land first so the conformance harness and pristine-intrinsic pattern become the baseline against which #218 and its follow-up are rebased and tested.

Core/Node-API/Source/js_native_api_shared.cc
bghgary: Please preserve `for...in`-equivalent Proxy prototype behavior on JSC. This new walk delegates each step to `napi_get_prototype`, whose JSC implementation bypasses `getPrototypeOf` traps, so trapped chains are invisible and the new tests skip that backend.

Core/Node-API/Source/js_native_api_shared.cc
bghgary: Please preserve V8's failure behavior for cyclic Proxy prototype chains instead of returning a finite result. V8's `napi_get_property_names` and `for...in` both terminate these cycles with `RangeError`, while this branch silently succeeds and the new cycle tests require that incompatible behavior.

Core/Node-API/Source/js_native_api_shared.cc
bghgary: Please compare property names with JavaScript string identity rather than their UTF-8 bytes. Distinct lone-surrogate names can encode to the same replacement bytes, so one can incorrectly shadow the other.

Core/Node-API/Source/js_native_api_shared.cc
bghgary: Please derive enumerable names and shadowing names from one `ownKeys` snapshot per level. Calling `Object.keys` and `Object.getOwnPropertyNames` separately invokes a Proxy's `ownKeys` trap twice, so a stateful or throwing second trap can change or fail an enumeration that `for...in` performs once.

Tests/UnitTests/Shared/Shared.cpp
```diff
@@ -101,6 +103,50 @@ TEST(JavaScript, All)
...
+                    // A failed call may or may not have left a JavaScript
+                    // exception pending; surface either as a thrown error so
+                    // that the script tests can assert on it uniformly.
+                    bool isExceptionPending{};
+                    if (napi_is_exception_pending(rawEnv, &isExceptionPending) == napi_ok && isExceptionPending)
+                    {
+                        napi_value error{};
+                        napi_get_and_clear_last_exception(rawEnv, &error);
+                    }
+
```
bghgary: Please preserve and assert the original pending exception instead of clearing it and throwing a generic status error. As written, the throwing-Proxy test passes even if `napi_get_property_names` discards or replaces the trap's exception, so it does not verify the propagation claim.

Core/Node-API/Source/js_native_api_shared.cc
bkaradzic-microsoft: @matthargett Thanks for the clarification. Agreed on the distinction: the public JSC APIs are available; the issue is that their enumeration semantics do not provide the descriptor-aware, own-property collection this walk needs. The shared walk can remain shared while its intrinsic functions come from per-environment snapshots. I also checked the current JSC `napi_has_own_property`: it still calls `hasOwnProperty` with the wrong receiver and no key argument. That is a separate correctness defect, not a sound precedent for this implementation.

To make the follow-up explicit: capture the required intrinsics before application code runs, use those captured functions throughout the intrinsic-dependent surface, correct the `napi_has_own_property` receiver/argument handling, and add poisoned-intrinsic and own-property conformance cases. Current head `d9383afb56383b9ca5650eef256efc09908ffa0c` still looks up mutable `Object.keys` / `Object.getOwnPropertyNames`; this PR should not be described as implementing that hardening. Your preference for #116 and its prerequisites to land first is understood; the final landing order remains @bghgary's decision, and an unmerged prerequisite is still relevant to that decision.

I also rechecked the remaining items in your overall review rather than treating the resolved inline threads as completion. The collector still uses UTF-8 strings for key identity, and the raw test bridge still clears a pending exception and replaces it with a generic error. Those two correctness concerns remain outstanding on this head; they should not be folded into the deferred repo-wide hardening. The portable CTS/ABI coverage, alignment with #189's engine definitions, and non-V8 `napi_get_all_property_names` support/follow-up are likewise not completed by this PR. The current green CI does not establish coverage for those gaps.

Core/Node-API/Source/js_native_api_shared.cc
bghgary: Please preserve `for...in`-equivalent Proxy prototype behavior on JSC. This new walk delegates each step to `napi_get_prototype`, whose JSC implementation bypasses `getPrototypeOf` traps, so trapped chains are invisible and the new tests skip that backend.

Core/Node-API/Source/js_native_api_shared.cc
bghgary, line 86: Please preserve V8's failure behavior for cyclic Proxy prototype chains instead of returning a finite result. V8's `napi_get_property_names` and `for...in` both terminate these cycles with `RangeError`, while this branch silently succeeds and the new cycle tests require that incompatible behavior.

Core/Node-API/Source/js_native_api_shared.cc
bghgary: Please compare property names with JavaScript string identity rather than their UTF-8 bytes. Distinct lone-surrogate names can encode to the same replacement bytes, so one can incorrectly shadow the other.

Core/Node-API/Source/js_native_api_shared.cc
bghgary: Please derive enumerable names and shadowing names from one `ownKeys` snapshot per level. Calling `Object.keys` and `Object.getOwnPropertyNames` separately invokes a Proxy's `ownKeys` trap twice, so a stateful or throwing second trap can change or fail an enumeration that `for...in` performs once.

Tests/UnitTests/Shared/Shared.cpp
bghgary, line 144: Please preserve and assert the original pending exception instead of clearing it and throwing a generic status error. As written, the throwing-Proxy test passes even if `napi_get_property_names` discards or replaces the trap's exception, so it does not verify the propagation claim.

bghgary (changes requested):

Concerns inline.

Core/Node-API/Source/js_native_api_shared.cc
bkaradzic-microsoft: Addressed in `2bb1fdd`: the shared walk now calls `Object.getPrototypeOf` through JavaScript instead of `napi_get_prototype`, so JSC observes `getPrototypeOf` traps. The proxy cycle and throwing-trap tests now run on JSC as well.

Core/Node-API/Source/js_native_api_shared.cc
bkaradzic-microsoft: Addressed in `2bb1fdd`: revisiting an object now throws a JavaScript `RangeError`, matching V8 instead of returning the finite prefix. Both one-object and two-object cycle regressions assert this behavior.

Core/Node-API/Source/js_native_api_shared.cc
bkaradzic-microsoft: Addressed in `2bb1fdd`: shadowed names are retained as `napi_value`s and compared with `napi_strict_equals`, avoiding lossy UTF-8 conversion. Added coverage for distinct high/low lone-surrogate names.

Core/Node-API/Source/js_native_api_shared.cc
bkaradzic-microsoft: Addressed in `2bb1fdd`: each level now takes one `Object.getOwnPropertyNames` snapshot and checks each returned key with `Object.getOwnPropertyDescriptor`; `Object.keys` is no longer called. Added a proxy regression asserting exactly one `ownKeys` invocation.

Tests/UnitTests/Shared/Shared.cpp
bkaradzic-microsoft: Addressed in `2bb1fdd`: the raw test bridge now rethrows the value returned by `napi_get_and_clear_last_exception` rather than replacing it with a status error. The throwing-proxy regression now asserts the original `"trap"` message.

Core/Node-API/Source/js_native_api_shared.cc
```diff
@@ -0,0 +1,140 @@
...
+    napi_value global{};
+    napi_value objectConstructor{};
+    napi_value getOwnPropertyNames{};
+    napi_value getOwnPropertyDescriptor{};
+    napi_value getPrototypeOf{};
+    RETURN_IF_NOT_OK(napi_get_global(env, &global));
+    RETURN_IF_NOT_OK(napi_get_named_property(env, global, "Object", &objectConstructor));
+    RETURN_IF_NOT_OK(napi_get_named_property(env, objectConstructor, "getOwnPropertyNames", &getOwnPropertyNames));
+    RETURN_IF_NOT_OK(napi_get_named_property(env, objectConstructor, "getOwnPropertyDescriptor", &getOwnPropertyDescriptor));
+    RETURN_IF_NOT_OK(napi_get_named_property(env, objectConstructor, "getPrototypeOf", &getPrototypeOf));
```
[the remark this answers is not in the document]
bkaradzic-microsoft: Fixed in eaf8cfe. JavaScriptCore, Chakra, and QuickJS now capture strong references to the Object constructor and the three enumeration intrinsics when their environments attach, and release them on detach. The walk uses those references (including for primitive coercion), so replacing globalThis.Object or any of its methods afterward cannot forge names or break enumeration. I also replaced linear name deduplication with UTF-16 hashing, preserving distinct lone-surrogate keys, and corrected the JavaScriptCore UTF-16 copy length. Added regressions for overwritten globals and 1,024 mixed enumerable/non-enumerable properties; the local JavaScriptCore and QuickJS suites each pass 251 tests. Chakra validation is pending CI.

Core/Node-API/Source/env_chakra.cc
```diff
@@ -44,6 +50,10 @@ namespace Napi
     void Detach(Env env)
     {
         napi_env env_ptr{env};
+        if (napi_shared::ReleasePropertyNameIntrinsics(env_ptr, env_ptr->property_name_intrinsics) != napi_ok)
```
bghgary: Please release these references while the Chakra context and runtime are live. `AppRuntime_Chakra` disposes the runtime before `Detach`, so `napi_delete_reference` calls `JsRelease` after teardown. Both Win32 Chakra jobs now crash on the first runtime test.

Core/Node-API/Source/env_javascriptcore.cc
```diff
@@ -1,18 +1,28 @@
...
+            delete env_ptr;
+            throw std::runtime_error{"Napi::Attach: failed to capture property-name intrinsics"};
+        }
         return {env_ptr};
     }
 
     void Detach(Napi::Env env)
     {
         napi_env env_ptr{env};
+        if (napi_shared::ReleasePropertyNameIntrinsics(env_ptr, env_ptr->property_name_intrinsics) != napi_ok)
```
bghgary: Please avoid leaving reference sentinels on the captured built-ins after `Detach`. `napi_ref__::init` installs non-deletable sentinels whose finalizers dereference the environment, but deleting the references only unprotects the values. A caller-retained context can collect the sentinels after the environment is deleted.

bghgary: Please avoid leaving reference sentinels on the captured built-ins after `Detach`. `napi_ref__::init` installs non-deletable sentinels whose finalizers dereference the environment, but deleting the references only unprotects the values. A caller-retained context can collect the sentinels after the environment is deleted.

Core/Node-API/Source/js_native_api_shared.cc
```diff
@@ -0,0 +1,226 @@
...
+#ifdef _WIN32
+        if (tracing) {
+          std::fprintf(stderr, "PROPERTY_NAMES_TRACE: length=%zu\n", length);
+          std::fflush(stderr);
+        }
+#endif
+        std::u16string key(length + 1, u'\0');
+        trace("key allocated");
+        size_t copied{};
+        RETURN_IF_NOT_OK(napi_get_value_string_utf16(env, name, key.data(), key.size(), &copied));
```
bghgary: Please correct the Chakra UTF-16 copy size before using this path. Its `JsCopyStringUtf16` wrapper calls `memcpy_s(buffer, bufferSize, ..., stringLength * sizeof(wchar_t))`, but the caller supplies `bufferSize` in UTF-16 units, not bytes. Every nonempty name trips the invalid-parameter handler; both Chakra jobs fast-fail after `utf16 length`, before `utf16 copied`.

Tests/UnitTests/Scripts/tests.ts
```diff
@@ -1696,6 +1698,272 @@ describe("napi class prototype isolation (#172)", function () {
...
+        const prototype = Object.getPrototypeOf;
+        const object = Object.create({ inherited: 1 });
+        object.own = 2;
+        let names: string[] | undefined;
+
+        try {
+            Object.getOwnPropertyNames = () => ["forged"];
+            Object.getOwnPropertyDescriptor = () => ({ enumerable: false });
+            Object.getPrototypeOf = () => null;
+            Reflect.set(globalThis, "Object", {});
```
bghgary: Please use a Chakra-compatible way to access the global object in both new tests. Both Win32 Chakra jobs now pass the UTF-16 copy test but fail these two tests with ReferenceError: 'globalThis' is not defined.

PR #221, opened 2026-08-06
XMLHttpRequest: implement the `on<event>` handler properties

bkaradzic-microsoft:
### Problem

`RaiseEvent` dispatches only to handlers registered via `addEventListener`. The
DOM `on<event>` properties had no accessors, so they were stored as plain
expandos and never called:

```js
request.onreadystatechange = function () { /* never runs */ };
request.onerror = function () { /* never runs */ };
```

The failure is silent: the request completes and the state is correct
(`readyState === 4`, `status === 200`), but no callback fires and no error is
reported. The caller waits forever.

This surfaced in the BabylonNative Playground suite, where the tests that fetch
their scene script over XHR use `onreadystatechange`. All of them hung until
timeout and were marked excluded on every graphics API, with a misattributed
"scene never becomes ready" reason.

### Changes

**1. `on<event>` handler properties.** Adds `onreadystatechange`, `onload`,
`onerror`, `onloadend` and `onabort` as accessors. They share one listener list
per event type with `addEventListener`, tagged so the single `on<event>` entry
can be found:

```cpp
struct Listener
{
    Napi::FunctionReference callback;
    bool isEventHandler;
};

std::unordered_map<std::string, std::vector<Listener>> m_listeners;
```

One list is what the DOM specifies, and it gets the observable details right:
dispatch follows registration order across both styles; reassigning the
property keeps its position rather than moving to the end; `xhr.onload = f` and
`addEventListener("load", f)` are independent registrations, so `f` is called
twice; `removeEventListener` does not remove an `on<event>` handler (assigning
`null` does). Per WebIDL these are `[LegacyTreatNonObjectAsNull]`, so
`xhr.onload = 0` clears the handler rather than throwing.

**2. Raise `load` on success.** It was never raised, so neither `onload` nor
`addEventListener("load", ...)` could fire — only `loadend`, plus `error` on
failure.

**3. Raise `abort` when a request is aborted.** `Abort()` only forwarded to
`UrlLib`, and the continuation reported the cancelled transfer as a transport
`error`. A cancelled request now dispatches `abort` + `loadend`, never `error`.

**4. A duplicate `addEventListener` is a no-op.** Re-adding an identical
`(type, callback)` pair threw `Cannot add the same event handler twice`; per DOM
the second add is silently ignored. The scan still skips the `on<event>` entry,
so the two-registration case above is unchanged.

**5. `error` is transport-level only.**

```cpp
const bool failed = result.has_error() || statusCode == 0;
```

Previously any non-2xx status took the error branch, so a 404 fired `error` and
`load` never fired. Per spec `error` means the transfer did not complete; a 404
*is* a completed exchange, so it dispatches `load` and callers branch on
`xhr.status`.

The condition also covers a missing local file on UWP, where `UrlLib` leaves
the status at 0. `UrlStatusCode::None` (0) is only ever the initial value and
the `ResetForOpen` reset — every path that produces a response assigns an
explicit code, including non-HTTP ones, where a local file read sets `Ok`. So
`statusCode == 0` means precisely "no response was obtained".

### Tests

12 regression tests in `Tests/UnitTests/Scripts/tests.ts` covering: each
handler property firing (and reading back, being replaced, cleared, and
coercing a non-callable to `null`); 404 dispatching `load` rather than `error`
via both registration styles; abort dispatching `abort`; registration-order
dispatch; a reassigned handler keeping its position; a function registered both
ways being called twice; a duplicate `addEventListener` not throwing and firing
once; and `removeEventListener` not removing an `on<event>` handler.

All 227 unit tests pass on Windows.

> [!NOTE]
> `Tests/UnitTests/dist/` is gitignored and CMake only copies it, so `npm run
> build` in `Tests/` is required after editing `tests.ts` — otherwise the build
> silently tests a stale bundle.

### Compatibility

The `on<event>` properties, `load` and `abort` are additive: they cover cases
that previously could not fire at all.

**One behavioral change worth a close look:** a non-2xx response now fires
`load` instead of `error`. Code relying on `onerror` to observe an HTTP 404
must check `xhr.status` inside `onload` instead. This matches browsers, and
`loadend` still fires either way, so anything settling on `loadend` is
unaffected.

### DIFF
--- a/Polyfills/XMLHttpRequest/Source/XMLHttpRequest.cpp
+++ b/Polyfills/XMLHttpRequest/Source/XMLHttpRequest.cpp
@@ -292,10 +292,14 @@ namespace Babylon::Polyfills::Internal
         {
             // Deliberately skips the `on<event>` entry: `xhr.onload = f` followed by
             // `xhr.addEventListener("load", f)` is two independent registrations, and a browser
-            // calls `f` twice rather than rejecting the second.
+            // calls `f` twice rather than collapsing them.
             if (!listener.isEventHandler && listener.callback.Value() == eventHandler)
             {
-                throw Napi::Error::New(info.Env(), "Cannot add the same event handler twice");
+                // Per DOM, re-adding an identical (type, callback, capture) triple is a silent
+                // no-op rather than an error: "If eventTarget's event listener list does not
+                // contain an event listener whose type is listener's type [...] then append
+                // listener". The listener stays registered once and is dispatched once.
+                return;
             }
         }
 
--- a/Polyfills/XMLHttpRequest/Source/XMLHttpRequest.cpp
+++ b/Polyfills/XMLHttpRequest/Source/XMLHttpRequest.cpp
@@ -89,8 +89,10 @@ namespace Babylon::Polyfills::Internal
     {
         const char* eventType = EVENT_TYPE_NAMES[static_cast<size_t>(Index)];
 
-        // Assigning null/undefined clears the handler, matching the DOM behavior where
-        // `xhr.onload = null` detaches the previously assigned handler.
+        // `EventHandler` attributes are declared [LegacyTreatNonObjectAsNull] in WebIDL, so a
+        // non-callable assignment is coerced to null rather than throwing: `xhr.onload = 0`
+        // leaves `xhr.onload === null`. We extend that to non-callable objects too -- storing a
+        // value we could never invoke would only defer the failure to dispatch time.
         if (!value.IsFunction())
         {
             m_onEventHandlerRefs.erase(eventType);
@@ -296,6 +298,10 @@ namespace Babylon::Polyfills::Internal
 
     void XMLHttpRequest::Abort(const Napi::CallbackInfo&)
     {
+        // Record the caller's intent so the in-flight continuation reports this as an abort
+        // rather than a transport error. If no request is in flight this is inert, matching the
+        // DOM, where abort() on an unsent request produces no observable events.
+        m_aborted = true;
         m_request.Abort();
     }
 
@@ -366,7 +372,13 @@ namespace Babylon::Polyfills::Internal
                 const bool failed = result.has_error() || statusCode < 200 || statusCode >= 300;
 
                 SetReadyState(ReadyState::Done);
-                if (failed)
+                if (m_aborted)
+                {
+                    // A cancelled request is not a transport failure: the DOM reports it as
+                    // 'abort' + 'loadend' and never raises 'error'.
+                    RaiseEvent(EventType::Abort);
+                }
+                else if (failed)
                 {
                     RaiseEvent(EventType::Error);
                 }
@@ -393,21 +405,44 @@ namespace Babylon::Polyfills::Internal
     {
         std::string traceName = (std::ostringstream{} << "XMLHttpRequest::RaiseEvent [" << eventType << "] [" << m_url << "]").str();
         arcana::trace_region raiseEventRegion{traceName.c_str()};
+
+        Napi::Env env = Env();
+
+        // Snapshot the handlers before dispatching. A handler may call addEventListener,
+        // removeEventListener, or reassign an on<event> property while it runs, which would
+        // otherwise reallocate the vector or rehash the map out from under this dispatch.
+        // (Mirrors FileReader::Dispatch.)
+        std::vector<Napi::Function> handlers{};
+
+        const auto onIt = m_onEventHandlerRefs.find(eventType);
+        if (onIt != m_onEventHandlerRefs.end() && !onIt->second.IsEmpty())
+        {
+            handlers.push_back(onIt->second.Value());
+        }
+
         const auto it = m_eventHandlerRefs.find(eventType);
         if (it != m_eventHandlerRefs.end())
         {
-            const auto& eventHandlerRefs = it->second;
-            for (const auto& eventHandlerRef : eventHandlerRefs)
+            handlers.reserve(handlers.size() + it->second.size());
+            for (const auto& eventHandlerRef : it->second)
             {
-                eventHandlerRef.Call({});
+                if (!eventHandlerRef.IsEmpty())
+                {
+                    handlers.push_back(eventHandlerRef.Value());
+                }
             }
         }
 
-        // The DOM dispatches the `on<event>` handler alongside any addEventListener handlers.
-        const auto onIt = m_onEventHandlerRefs.find(eventType);
-        if (onIt != m_onEventHandlerRefs.end())
+        for (const auto& handler : handlers)
         {
-            onIt->second.Call({});
+            handler.Call({});
+
+            // A throwing handler must not abort the remaining dispatch, and the exception must
+            // not escape into the native completion continuation that called us.
+            if (env.IsExceptionPending())
+            {
+                env.GetAndClearPendingException();
+            }
         }
     }
 }
--- a/Polyfills/XMLHttpRequest/Source/XMLHttpRequest.cpp
+++ b/Polyfills/XMLHttpRequest/Source/XMLHttpRequest.cpp
@@ -2,6 +2,7 @@
 #include <Babylon/JsRuntime.h>
 #include <Babylon/Polyfills/XMLHttpRequest.h>
 #include <arcana/tracing/trace_region.h>
+#include <algorithm>
 #include <cstring>
 #include <sstream>
 
@@ -75,19 +76,28 @@ namespace Babylon::Polyfills::Internal
     template<XMLHttpRequest::EventIndex Index>
     Napi::Value XMLHttpRequest::GetEventHandler(const Napi::CallbackInfo&)
     {
-        const auto it = m_onEventHandlerRefs.find(EVENT_TYPE_NAMES[static_cast<size_t>(Index)]);
-        if (it == m_onEventHandlerRefs.end())
+        const auto it = m_listeners.find(EVENT_TYPE_NAMES[static_cast<size_t>(Index)]);
+        if (it != m_listeners.end())
         {
-            return Env().Null();
+            for (const auto& listener : it->second)
+            {
+                if (listener.isEventHandler)
+                {
+                    return listener.callback.Value();
+                }
+            }
         }
 
-        return it->second.Value();
+        return Env().Null();
     }
 
     template<XMLHttpRequest::EventIndex Index>
-    void XMLHttpRequest::SetEventHandler(const Napi::CallbackInfo& info, const Napi::Value& value)
+    void XMLHttpRequest::SetEventHandler(const Napi::CallbackInfo&, const Napi::Value& value)
     {
-        const char* eventType = EVENT_TYPE_NAMES[static_cast<size_t>(Index)];
+        auto& listeners = m_listeners[EVENT_TYPE_NAMES[static_cast<size_t>(Index)]];
+        const auto it = std::find_if(listeners.begin(), listeners.end(), [](const Listener& listener) {
+            return listener.isEventHandler;
+        });
 
         // `EventHandler` attributes are declared [LegacyTreatNonObjectAsNull] in WebIDL, so a
         // non-callable assignment is coerced to null rather than throwing: `xhr.onload = 0`
@@ -95,12 +105,24 @@ namespace Babylon::Polyfills::Internal
         // value we could never invoke would only defer the failure to dispatch time.
         if (!value.IsFunction())
         {
-            m_onEventHandlerRefs.erase(eventType);
+            if (it != listeners.end())
+            {
+                listeners.erase(it);
+            }
+
             return;
         }
 
-        m_onEventHandlerRefs[eventType] = Napi::Persistent(value.As<Napi::Function>());
-        (void)info;
+        if (it != listeners.end())
+        {
+            // Replace in place so reassignment keeps this listener's position in the
+            // dispatch order.
+            it->callback = Napi::Persistent(value.As<Napi::Function>());
+        }
+        else
+        {
+            listeners.push_back(Listener{Napi::Persistent(value.As<Napi::Function>()), true});
+        }
     }
 
     void XMLHttpRequest::Initialize(Napi::Env env)
@@ -265,31 +287,36 @@ namespace Babylon::Polyfills::Internal
         const std::string eventType = info[0].As<Napi::String>().Utf8Value();
         const Napi::Function eventHandler = info[1].As<Napi::Function>();
 
-        const auto& eventHandlerRefs = m_eventHandlerRefs[eventType];
-        for (auto it = eventHandlerRefs.begin(); it != eventHandlerRefs.end(); ++it)
+        auto& listeners = m_listeners[eventType];
+        for (const auto& listener : listeners)
         {
-            if (it->Value() == eventHandler)
+            // Deliberately skips the `on<event>` entry: `xhr.onload = f` followed by
+            // `xhr.addEventListener("load", f)` is two independent registrations, and a browser
+            // calls `f` twice rather than rejecting the second.
+            if (!listener.isEventHandler && listener.callback.Value() == eventHandler)
             {
                 throw Napi::Error::New(info.Env(), "Cannot add the same event handler twice");
             }
         }
 
-        m_eventHandlerRefs[eventType].push_back(Napi::Persistent(eventHandler));
+        listeners.push_back(Listener{Napi::Persistent(eventHandler), false});
     }
 
     void XMLHttpRequest::RemoveEventListener(const Napi::CallbackInfo& info)
     {
         const std::string eventType = info[0].As<Napi::String>().Utf8Value();
         const Napi::Function eventHandler = info[1].As<Napi::Function>();
-        const auto itType = m_eventHandlerRefs.find(eventType);
-        if (itType != m_eventHandlerRefs.end())
+        const auto itType = m_listeners.find(eventType);
+        if (itType != m_listeners.end())
         {
-            auto& eventHandlerRefs = itType->second;
-            for (auto it = eventHandlerRefs.begin(); it != eventHandlerRefs.end(); ++it)
+            auto& listeners = itType->second;
+            for (auto it = listeners.begin(); it != listeners.end(); ++it)
             {
-                if (it->Value() == eventHandler)
+                // removeEventListener never removes an `on<event>` handler; that is done by
+                // assigning null to the property.
+                if (!it->isEventHandler && it->callback.Value() == eventHandler)
                 {
-                    eventHandlerRefs.erase(it);
+                    listeners.erase(it);
                     break;
                 }
             }
@@ -369,7 +396,15 @@ namespace Babylon::Polyfills::Internal
                 // success-only continuation here skipped readyState=Done / loadend / error and let the JS observer
                 // hang.
                 const auto statusCode = arcana::underlying_cast(m_request.StatusCode());
-                const bool failed = result.has_error() || statusCode < 200 || statusCode >= 300;
+                // `error` is reserved for transport-level failure. A completed HTTP transaction
+                // that returned a non-2xx status (e.g. 404) is still a successful exchange, so it
+                // dispatches `load` and the caller branches on `xhr.status` inside the handler.
+                // UrlStatusCode::None (0) is UrlLib's "no response was obtained" sentinel: it is
+                // only ever the initial value and the reset in ResetForOpen, because every path
+                // that produces a response assigns an explicit code -- including the non-HTTP
+                // ones, where local file reads set Ok. That keeps the missing-local-file-on-UWP
+                // case (status left at 0) reporting `error`.
+                const bool failed = result.has_error() || statusCode == 0;
 
                 SetReadyState(ReadyState::Done);
                 if (m_aborted)
@@ -390,8 +425,7 @@ namespace Babylon::Polyfills::Internal
 
                 // Assume the XMLHttpRequest will only be used for a single request and clear the event handlers.
                 // Single use seems to be the standard pattern, and we need to release our strong refs to event handlers.
-                m_eventHandlerRefs.clear();
-                m_onEventHandlerRefs.clear();
+                m_listeners.clear();
             });
     }
 
@@ -414,21 +448,17 @@ namespace Babylon::Polyfills::Internal
         // (Mirrors FileReader::Dispatch.)
         std::vector<Napi::Function> handlers{};
 
-        const auto onIt = m_onEventHandlerRefs.find(eventType);
-        if (onIt != m_onEventHandlerRefs.end() && !onIt->second.IsEmpty())
+        const auto it = m_listeners.find(eventType);
+        if (it != m_listeners.end())
         {
-            handlers.push_back(onIt->second.Value());
-        }
-
-        const auto it = m_eventHandlerRefs.find(eventType);
-        if (it != m_eventHandlerRefs.end())
-        {
-            handlers.reserve(handlers.size() + it->second.size());
-            for (const auto& eventHandlerRef : it->second)
+            // One pass over the single list, so handlers run in registration order regardless of
+            // whether they arrived via addEventListener or an `on<event>` property.
+            handlers.reserve(it->second.size());
+            for (const auto& listener : it->second)
             {
-                if (!eventHandlerRef.IsEmpty())
+                if (!listener.callback.IsEmpty())
                 {
-                    handlers.push_back(eventHandlerRef.Value());
+                    handlers.push_back(listener.callback.Value());
                 }
             }
         }
--- a/Polyfills/XMLHttpRequest/Source/XMLHttpRequest.cpp
+++ b/Polyfills/XMLHttpRequest/Source/XMLHttpRequest.cpp
@@ -59,7 +59,46 @@ namespace Babylon::Polyfills::Internal
             constexpr const char* ReadyStateChange = "readystatechange";
             constexpr const char* LoadEnd = "loadend";
             constexpr const char* Error = "error";
+            constexpr const char* Load = "load";
+            constexpr const char* Abort = "abort";
         }
+    }
+
+    const char* const XMLHttpRequest::EVENT_TYPE_NAMES[static_cast<size_t>(XMLHttpRequest::EventIndex::Count)] = {
+        EventType::ReadyStateChange,
+        EventType::Load,
+        EventType::Error,
+        EventType::LoadEnd,
+        EventType::Abort,
+    };
+
+    template<XMLHttpRequest::EventIndex Index>
+    Napi::Value XMLHttpRequest::GetEventHandler(const Napi::CallbackInfo&)
+    {
+        const auto it = m_onEventHandlerRefs.find(EVENT_TYPE_NAMES[static_cast<size_t>(Index)]);
+        if (it == m_onEventHandlerRefs.end())
+        {
+            return Env().Null();
+        }
+
+        return it->second.Value();
+    }
+
+    template<XMLHttpRequest::EventIndex Index>
+    void XMLHttpRequest::SetEventHandler(const Napi::CallbackInfo& info, const Napi::Value& value)
+    {
+        const char* eventType = EVENT_TYPE_NAMES[static_cast<size_t>(Index)];
+
+        // Assigning null/undefined clears the handler, matching the DOM behavior where
+        // `xhr.onload = null` detaches the previously assigned handler.
+        if (!value.IsFunction())
+        {
+            m_onEventHandlerRefs.erase(eventType);
+            return;
+        }
+
+        m_onEventHandlerRefs[eventType] = Napi::Persistent(value.As<Napi::Function>());
+        (void)info;
     }
 
     void XMLHttpRequest::Initialize(Napi::Env env)
@@ -88,6 +127,15 @@ namespace Babylon::Polyfills::Internal
                 // to tell a DNS failure from a refused connection or a missing local asset.
                 InstanceAccessor("errorCode", &XMLHttpRequest::GetErrorCode, nullptr),
                 InstanceAccessor("errorDetail", &XMLHttpRequest::GetErrorDetail, nullptr),
+                // DOM `on<event>` handler properties. Without these, `xhr.onreadystatechange = fn`
+                // silently sets an ordinary expando property that is never invoked, so code written
+                // against the standard XMLHttpRequest API waits forever for a callback that can
+                // never fire.
+                InstanceAccessor("onreadystatechange", &XMLHttpRequest::GetEventHandler<EventIndex::ReadyStateChange>, &XMLHttpRequest::SetEventHandler<EventIndex::ReadyStateChange>),
+                InstanceAccessor("onload", &XMLHttpRequest::GetEventHandler<EventIndex::Load>, &XMLHttpRequest::SetEventHandler<EventIndex::Load>),
+                InstanceAccessor("onerror", &XMLHttpRequest::GetEventHandler<EventIndex::Error>, &XMLHttpRequest::SetEventHandler<EventIndex::Error>),
+                InstanceAccessor("onloadend", &XMLHttpRequest::GetEventHandler<EventIndex::LoadEnd>, &XMLHttpRequest::SetEventHandler<EventIndex::LoadEnd>),
+                InstanceAccessor("onabort", &XMLHttpRequest::GetEventHandler<EventIndex::Abort>, &XMLHttpRequest::SetEventHandler<EventIndex::Abort>),
                 InstanceMethod("getAllResponseHeaders", &XMLHttpRequest::GetAllResponseHeaders),
                 InstanceMethod("getResponseHeader", &XMLHttpRequest::GetResponseHeader),
                 InstanceMethod("setRequestHeader", &XMLHttpRequest::SetRequestHeader),
@@ -322,11 +370,16 @@ namespace Babylon::Polyfills::Internal
                 {
                     RaiseEvent(EventType::Error);
                 }
+                else
+                {
+                    RaiseEvent(EventType::Load);
+                }
                 RaiseEvent(EventType::LoadEnd);
 
                 // Assume the XMLHttpRequest will only be used for a single request and clear the event handlers.
                 // Single use seems to be the standard pattern, and we need to release our strong refs to event handlers.
                 m_eventHandlerRefs.clear();
+                m_onEventHandlerRefs.clear();
             });
     }
 
@@ -348,6 +401,13 @@ namespace Babylon::Polyfills::Internal
             {
                 eventHandlerRef.Call({});
             }
+        }
+
+        // The DOM dispatches the `on<event>` handler alongside any addEventListener handlers.
+        const auto onIt = m_onEventHandlerRefs.find(eventType);
+        if (onIt != m_onEventHandlerRefs.end())
+        {
+            onIt->second.Call({});
         }
     }
 }
--- a/Polyfills/XMLHttpRequest/Source/XMLHttpRequest.h
+++ b/Polyfills/XMLHttpRequest/Source/XMLHttpRequest.h
@@ -69,6 +69,8 @@ namespace Babylon::Polyfills::Internal
         UrlLib::UrlRequest m_request{};
         JsRuntimeScheduler m_runtimeScheduler;
         ReadyState m_readyState{ReadyState::Unsent};
+        // Set by abort(); makes the in-flight continuation report 'abort' instead of 'error'.
+        bool m_aborted{false};
         std::unordered_map<std::string, std::vector<Napi::FunctionReference>> m_eventHandlerRefs;
         // The DOM `on<event>` handler properties (onreadystatechange, onload, ...). These are
         // kept separate from m_eventHandlerRefs because they have assignment semantics -- setting
--- a/Polyfills/XMLHttpRequest/Source/XMLHttpRequest.h
+++ b/Polyfills/XMLHttpRequest/Source/XMLHttpRequest.h
@@ -65,16 +65,24 @@ namespace Babylon::Polyfills::Internal
         void SetReadyState(ReadyState readyState);
         void RaiseEvent(const char* eventType);
 
+        // A registered event listener. `isEventHandler` marks the single entry owned by the
+        // matching `on<event>` property; every other entry came from addEventListener. Both
+        // kinds share one list per event type because that is what the DOM specifies: dispatch
+        // follows registration order, so `addEventListener("load", a)` then `xhr.onload = b`
+        // calls `a` then `b`, and reassigning `onload` keeps its original position rather than
+        // moving to the end ("If eventHandler's listener is not null, then return").
+        struct Listener
+        {
+            Napi::FunctionReference callback;
+            bool isEventHandler;
+        };
+
         std::string m_url{};
         UrlLib::UrlRequest m_request{};
         JsRuntimeScheduler m_runtimeScheduler;
         ReadyState m_readyState{ReadyState::Unsent};
         // Set by abort(); makes the in-flight continuation report 'abort' instead of 'error'.
         bool m_aborted{false};
-        std::unordered_map<std::string, std::vector<Napi::FunctionReference>> m_eventHandlerRefs;
-        // The DOM `on<event>` handler properties (onreadystatechange, onload, ...). These are
-        // kept separate from m_eventHandlerRefs because they have assignment semantics -- setting
-        // one replaces the previous handler -- whereas addEventListener accumulates.
-        std::unordered_map<std::string, Napi::FunctionReference> m_onEventHandlerRefs;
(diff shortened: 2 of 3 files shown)


bkaradzic-microsoft:
Thanks -- both comments addressed in c870d43.

**On the non-callable setter (`XMLHttpRequest.cpp:101`):** throwing a `TypeError` here would actually diverge from the DOM. `EventHandler` attributes are declared `[LegacyTreatNonObjectAsNull]` in WebIDL, so a non-callable assignment is coerced to null rather than rejected -- `xhr.onload = 0` leaves `xhr.onload === null` in every browser, silently. The current clear-on-non-function behavior matches that for primitives.

Strictly, the spec does *store* non-callable **objects** (they just never get invoked); we clear those too, because keeping a value we could never call would only defer the failure to dispatch time. I've documented that deliberate narrowing in a code comment and added a regression test (`should coerce a non-callable on<event> assignment to null`) pinning the no-throw behavior.

**On `onabort` (`XMLHttpRequest.cpp:138`):** good catch, this one was a real defect -- `Abort()` only called `m_request.Abort()`, so the completion continuation reported the cancellation as a transport `error` and `onabort` was dead API. `Abort()` now records the intent and the continuation raises `abort` + `loadend` instead of `error`, per the DOM. Covered by a new test asserting that aborting an in-flight request fires `abort` and never `error`/`load`.

While in here I also hardened `RaiseEvent` along the lines of `FileReader::Dispatch`: it now snapshots the handler list before dispatching (a handler calling `addEventListener`/`removeEventListener`, or reassigning an `on<event>` property, could reallocate the vector or rehash the map out from under the in-flight dispatch -- a use-after-free) and clears pending exceptions between handlers so a throwing handler neither aborts the remaining dispatch nor escapes into the native completion continuation.

Polyfills/XMLHttpRequest/Source/XMLHttpRequest.cpp
```diff
@@ -318,15 +372,26 @@ namespace Babylon::Polyfills::Internal
...
+                    // 'abort' + 'loadend' and never raises 'error'.
+                    RaiseEvent(EventType::Abort);
+                }
+                else if (failed)
                 {
                     RaiseEvent(EventType::Error);
                 }
+                else
+                {
+                    RaiseEvent(EventType::Load);
```
bghgary, line 426: `failed` is `statusCode < 200 || statusCode >= 300` (L372), so a 404 takes the error branch and `load` never fires. Per the XHR spec `error` is network-level only; a completed 404 dispatches `load`, with callers branching on `xhr.status` inside it.

The new 404 test asserts `loadFired === false`, locking that in. Worth confirming it is intended — the description's "per the spec" holds for transport errors, not HTTP status codes.

The condition predates this PR and is load-bearing for the UWP status-0 case, so this is not a request to simply invert it.

Polyfills/XMLHttpRequest/Source/XMLHttpRequest.h
```diff
@@ -52,6 +69,12 @@ namespace Babylon::Polyfills::Internal
         UrlLib::UrlRequest m_request{};
         JsRuntimeScheduler m_runtimeScheduler;
         ReadyState m_readyState{ReadyState::Unsent};
+        // Set by abort(); makes the in-flight continuation report 'abort' instead of 'error'.
+        bool m_aborted{false};
         std::unordered_map<std::string, std::vector<Napi::FunctionReference>> m_eventHandlerRefs;
+        // The DOM `on<event>` handler properties (onreadystatechange, onload, ...). These are
+        // kept separate from m_eventHandlerRefs because they have assignment semantics -- setting
+        // one replaces the previous handler -- whereas addEventListener accumulates.
+        std::unordered_map<std::string, Napi::FunctionReference> m_onEventHandlerRefs;
```
bghgary: `on<event>` handlers belong in the same listener list as `addEventListener`, not a parallel map dispatched ahead of it. Two divergences follow from the split:

- Order: browsers dispatch in registration order, so `addEventListener("load", a)` then `xhr.onload = b` gives `a, b`; here it gives `b, a`.
- `xhr.onload = f; xhr.addEventListener("load", f)` throws (`XMLHttpRequest.cpp` L269), where a browser registers both and calls `f` twice -- per DOM a duplicate add is a silent no-op, not an error.

HTML registers one internal listener on first set whose callback indirects through the stored value, so reassignment keeps its position ("If eventHandler's listener is not null, then return").

One `vector<Listener>` per type of `{ FunctionReference callback; bool isEventHandler; }` covers it: the setter replaces the flagged entry in place, appends when absent, erases when the value is not callable; the getter reads it back. Drops this map and the on-handler-first branch in `RaiseEvent`.

The duplicate throw predates this PR.

bghgary (changes requested):

Two inline.

Polyfills/XMLHttpRequest/Source/XMLHttpRequest.cpp
bkaradzic-microsoft: You're right, and the test is wrong. A 404 is a completed HTTP transaction — per spec `error` is for network-level failures only, so `load` should fire and callers branch on `xhr.status` inside it.

On the part you flagged as load-bearing: I audited every `m_statusCode` assignment in UrlLib, and the status-0 case is narrower than the current condition assumes.

```
UrlStatusCode { None = 0, Ok = 200 }
```

Every non-HTTP success path sets `Ok` explicitly — `UrlRequest_UWP.cpp:70,78` (local file read), `Win32.cpp:78,98`, `Android.cpp:90,130`, `Apple.mm:201`, `Unix.cpp:251`, `Windows_Shared.h:131,143`. `None` only ever appears as the initial value and in the reset at `UrlRequest_Base.h:371`.

So `statusCode == 0` means precisely "no HTTP response was obtained", which is the transport-failure sentinel the UWP fix relies on — it is never a successful local file read. That makes the condition separable:

```cpp
const bool failed = result.has_error() || statusCode == 0;
```

That keeps the missing-local-file-on-UWP behaviour intact (still `error` + `loadend`) while letting a completed 404 dispatch `load` + `loadend`.

I'll make that change and flip the 404 test to assert `loadFired === true`, `errorFired === false`, `status === 404`.

Polyfills/XMLHttpRequest/Source/XMLHttpRequest.h
bkaradzic-microsoft: Agreed — the parallel map is the wrong model, and I'll switch to the single-list design you described.

To confirm I've got the shape right:

```cpp
struct Listener
{
    Napi::FunctionReference callback;
    bool isEventHandler;   // registered via on<event>, not addEventListener
};
std::unordered_map<std::string, std::vector<Listener>> m_listeners;
```

- setter: find the `isEventHandler` entry — replace its callback in place if present, append if not, erase if the value isn't callable
- getter: read that entry back, `null` when absent
- `RaiseEvent`: one pass over the vector, so dispatch is registration order
- `m_onEventHandlerRefs` and the on-handler-first branch both go away

That fixes the ordering divergence: `addEventListener("load", a)` then `xhr.onload = b` now gives `a, b`, and reassigning `onload` keeps its slot rather than moving to the end — matching "If eventHandler's listener is not null, then return".

For the second divergence, I'll scope the duplicate check in `AddEventListener` to non-`isEventHandler` entries, so `xhr.onload = f; xhr.addEventListener("load", f)` registers both and calls `f` twice, as a browser does.

That leaves the pre-existing `addEventListener` duplicate throw. Per DOM a duplicate add is a silent no-op, so the throw is also wrong, but it's independent of this bug and nothing currently covers it — happy to drop it here for symmetry, or leave it for a separate PR. Let me know which you'd prefer.

I'll add a test asserting dispatch order across both registration styles.

Polyfills/XMLHttpRequest/Source/XMLHttpRequest.cpp
bkaradzic-microsoft: Done in 20b0c20.

```cpp
const bool failed = result.has_error() || statusCode == 0;
```

A completed 404 now dispatches `load` + `loadend`, and the missing-local-file-on-UWP case still reports `error` because it leaves the status at 0.

I re-ran the audit against the exact pinned UrlLib (`0c99133`) rather than my working copy, and it holds — `UrlStatusCode::None` (0) appears in only two places, the initial value and the `ResetForOpen` reset:

- `UrlRequest_Base.h:371` — `ResetForOpen`
- `UrlRequest_Base.h:147` — scheme resolver, assigns the resolver's own status

Every path that produces a response assigns an explicit code, including the non-HTTP ones that set `Ok` for a local file read: `UrlRequest_UWP.cpp:70,78`, `Win32.cpp:84,104`, `Android.cpp:90,130`, `Apple.mm:191,201`, `Unix.cpp:310,314`, `Windows_Shared.h:117,146,158`.

So `statusCode == 0` means precisely "no response was obtained", which is the transport-failure sentinel the UWP fix relies on.

Both 404 tests are flipped to assert `loadFired === true`, `errorFired === false`, `status === 404`. That includes the older `should fire 'error' event for a remote URL that returns HTTP 404` from #165 — it predates this PR, but it asserts the behaviour this changes, so I retitled it and kept its original point (that the completion events fire at all, with `readyState === 4`).

Polyfills/XMLHttpRequest/Source/XMLHttpRequest.h
bkaradzic-microsoft: Done in 20b0c20 — switched to the single-list design.

`m_onEventHandlerRefs` is gone. `m_listeners` is now one `vector<Listener>` per event type:

```cpp
struct Listener
{
    Napi::FunctionReference callback;
    bool isEventHandler;
};
```

- setter: finds the `isEventHandler` entry and replaces its callback **in place**, appends when absent, erases when the assigned value is not callable
- getter: reads that entry back, `null` when absent
- `RaiseEvent`: one pass over the vector, so dispatch is registration order
- `AddEventListener`: the duplicate check skips `isEventHandler` entries
- `RemoveEventListener`: the match also skips them, since the property is cleared by assigning `null`, not by `removeEventListener`

Four tests cover the behaviour, and each fails against the previous implementation:

| Test | Was | Now |
|---|---|---|
| `addEventListener("load", a)`, `onload = b`, `addEventListener("load", c)` | `b, a, c` | `a, b, c` |
| `onload = x`, `addEventListener("load", l)`, `onload = y` | `y, l` — but only incidentally, since on-handlers always ran first | `y, l`, because the entry keeps its slot |
| `onload = f`, `addEventListener("load", f)` | threw | `f` called twice |
| `onload = f`, `removeEventListener("load", f)` | n/a | `f` still called |

On the pre-existing `addEventListener` duplicate throw: I left it as-is here, since it is independent of this bug and nothing currently covers it. Happy to drop it in a follow-up if you would like DOM-correct silent no-op behaviour.

bghgary: [Responded by Copilot on behalf of @bghgary]

Remove it here. You're already in that code, and per DOM a duplicate add is a silent no-op — leaving the throw would park a second known-wrong behaviour next to the one you're fixing.

Your UrlLib audit holds, checked independently: all sixteen `m_statusCode` assignments set `Ok` explicitly on the non-HTTP success paths, and the only `None` is the reset in `UrlRequest_Base.h`. The blob-handler path returns early via `SetError` when unhandled, so a successful request can't land at 0 either.

bkaradzic-microsoft: Removed in 87c5c1b — the duplicate `addEventListener` now returns silently instead of throwing, with a comment citing the DOM "append listener" step.

The scan still skips `isEventHandler` entries, so the two-registration case is unchanged: `xhr.onload = f` followed by `xhr.addEventListener("load", f)` still calls `f` twice. Only two `addEventListener` calls with the same pair collapse to one.

Added a test pinning the new behaviour (doesn't throw, handler fires exactly once); the existing "called twice" test guards the other direction. Suite is 227 passing on Windows.

And thanks for double-checking the `UrlLib` status-code paths — good to have that confirmed independently rather than resting on my read alone.

One note: the identical throw also exists in `Polyfills/AbortController/Source/AbortSignal.cpp:129`. I left it alone since it's outside this PR's surface, but happy to fix it here or file a follow-up, whichever you prefer.

Polyfills/XMLHttpRequest/Source/XMLHttpRequest.cpp
```diff
@@ -59,6 +60,68 @@ namespace Babylon::Polyfills::Internal
...
+        auto& listeners = m_listeners[EVENT_TYPE_NAMES[static_cast<size_t>(Index)]];
+        const auto it = std::find_if(listeners.begin(), listeners.end(), [](const Listener& listener) {
+            return listener.isEventHandler;
+        });
+
+        // `EventHandler` attributes are declared [LegacyTreatNonObjectAsNull] in WebIDL, so a
+        // non-callable assignment is coerced to null rather than throwing: `xhr.onload = 0`
+        // leaves `xhr.onload === null`. We extend that to non-callable objects too -- storing a
+        // value we could never invoke would only defer the failure to dispatch time.
+        if (!value.IsFunction())
```
bghgary: `[LegacyTreatNonObjectAsNull]` only turns non-object values into null; non-callable objects are retained as no-op handlers. Preserve those object assignments instead of clearing them.

Polyfills/XMLHttpRequest/Source/XMLHttpRequest.cpp
```diff
@@ -254,6 +335,10 @@ namespace Babylon::Polyfills::Internal
 
     void XMLHttpRequest::Abort(const Napi::CallbackInfo&)
     {
+        // Record the caller's intent so the in-flight continuation reports this as an abort
+        // rather than a transport error. If no request is in flight this is inert, matching the
+        // DOM, where abort() on an unsent request produces no observable events.
+        m_aborted = true;
```
bghgary: `abort()` permanently poisons both layers: `m_aborted` never resets, and pinned UrlLib already has a cancellation source that stays canceled because `ResetForOpen()` does not replace it. Follow the XHR state rules: before an active send do nothing; from `DONE` transition to `UNSENT` without `readystatechange`; for an active send, cancel and dispatch `readystatechange`, `abort`, and `loadend` synchronously, suppress its later continuation, and reset or recreate both states before reuse.

Polyfills/XMLHttpRequest/Source/XMLHttpRequest.cpp
```diff
@@ -321,18 +406,36 @@ namespace Babylon::Polyfills::Internal
...
+                else
+                {
+                    RaiseEvent(EventType::Load);
+                }
                 RaiseEvent(EventType::LoadEnd);
 
                 // Assume the XMLHttpRequest will only be used for a single request and clear the event handlers.
                 // Single use seems to be the standard pattern, and we need to release our strong refs to event handlers.
-                m_eventHandlerRefs.clear();
+                m_listeners.clear();
```
bghgary: Preserve the new `on<event>` entries after completion. Clearing the unified map makes `xhr.onload` read back as `null` and loses the property handler before a reused XHR's next request.

Polyfills/XMLHttpRequest/Source/XMLHttpRequest.cpp
```diff
@@ -346,13 +449,39 @@ namespace Babylon::Polyfills::Internal
...
-        const auto it = m_eventHandlerRefs.find(eventType);
-        if (it != m_eventHandlerRefs.end())
+
+        Napi::Env env = Env();
+
+        // Snapshot the handlers before dispatching. A handler may call addEventListener,
+        // removeEventListener, or reassign an on<event> property while it runs, which would
+        // otherwise reallocate the vector or rehash the map out from under this dispatch.
+        // (Mirrors FileReader::Dispatch.)
+        std::vector<Napi::Function> handlers{};
```
bghgary: Snapshot stable listener records rather than callback values, then check each record's current state before invocation. A listener removed before a later callback in the same dispatch must not run, and an `on<event>` property reassigned before its turn must invoke the new value.

Polyfills/XMLHttpRequest/Source/XMLHttpRequest.cpp
```diff
@@ -346,13 +449,39 @@ namespace Babylon::Polyfills::Internal
...
+                    handlers.push_back(listener.callback.Value());
+                }
+            }
+        }
+
+        for (const auto& handler : handlers)
         {
-            const auto& eventHandlerRefs = it->second;
-            for (const auto& eventHandlerRef : eventHandlerRefs)
+            handler.Call({});
```
bghgary: Invoke each callback with the XHR wrapper as `this` and the appropriate `Event` or `ProgressEvent`, with the correct `type`, `target`, and `currentTarget`. The new handler properties otherwise break normal callbacks that read `this.status` or `event.target`.

Polyfills/XMLHttpRequest/Source/XMLHttpRequest.cpp
```diff
@@ -88,6 +151,15 @@ namespace Babylon::Polyfills::Internal
                 // to tell a DNS failure from a refused connection or a missing local asset.
                 InstanceAccessor("errorCode", &XMLHttpRequest::GetErrorCode, nullptr),
                 InstanceAccessor("errorDetail", &XMLHttpRequest::GetErrorDetail, nullptr),
+                // DOM `on<event>` handler properties. Without these, `xhr.onreadystatechange = fn`
+                // silently sets an ordinary expando property that is never invoked, so code written
+                // against the standard XMLHttpRequest API waits forever for a callback that can
+                // never fire.
+                InstanceAccessor("onreadystatechange", &XMLHttpRequest::GetEventHandler<EventIndex::ReadyStateChange>, &XMLHttpRequest::SetEventHandler<EventIndex::ReadyStateChange>),
```
bghgary, line 237: Update `Polyfills/XMLHttpRequest/Readme.md` with these handler properties and event semantics. It still says `onload`-style properties are unsupported and that non-2xx responses raise `error`.

Polyfills/XMLHttpRequest/Source/XMLHttpRequest.cpp
```diff
@@ -346,13 +449,39 @@ namespace Babylon::Polyfills::Internal
...
-            const auto& eventHandlerRefs = it->second;
-            for (const auto& eventHandlerRef : eventHandlerRefs)
+            handler.Call({});
+
+            // A throwing handler must not abort the remaining dispatch, and the exception must
+            // not escape into the native completion continuation that called us.
+            if (env.IsExceptionPending())
             {
-                eventHandlerRef.Call({});
+                env.GetAndClearPendingException();
```
bghgary: Don't discard exceptions thrown by handlers. Continue dispatching to later listeners, but report every thrown exception through the runtime's unhandled-exception path.

matthargett:
Heads-up on a rooting hazard that this PR mirrors from `FileReader::Dispatch` ("Mirrors FileReader::Dispatch"): the snapshot is a `std::vector<Napi::Function>` of raw values held across the listener calls. On JavaScriptCore nothing roots a `napi_value` that lives only on the C++ heap (the backend's handle scopes are stubs and it scans just the C stack), so if a listener removes a *later* listener during dispatch, that later `Napi::Function` can be collected before the loop reaches it — the same class of bug as the `CompressionStream` output-chunk corruption fixed on #211, which reproduces deterministically under `JSC_collectContinuously=1`. Snapshotting the `FunctionReference`s (or a JS array) instead of the values avoids it. I was going to send a small PR for `FileReader`; since this PR adds a second copy of the pattern, either I fold both into one follow-up after this lands, or you take the `FunctionReference` snapshot here and I do `FileReader` only — your call.

matthargett:
Follow-up: the FileReader side of this is now its own PR, #248 (snapshot `Napi::FunctionReference`s via `Napi::Persistent` instead of bare values, released when dispatch returns). If you mirror that shape in `XMLHttpRequest::Dispatch` here, both stay consistent; happy to rebase #248 on top of this if it lands first.

Polyfills/XMLHttpRequest/Source/XMLHttpRequest.cpp
```diff
@@ -59,9 +59,48 @@ namespace Babylon::Polyfills::Internal
...
+        // Assigning null/undefined clears the handler, matching the DOM behavior where
+        // `xhr.onload = null` detaches the previously assigned handler.
+        if (!value.IsFunction())
+        {
+            m_onEventHandlerRefs.erase(eventType);
+            return;
+        }
+
+        m_onEventHandlerRefs[eventType] = Napi::Persistent(value.As<Napi::Function>());
+        (void)info;
```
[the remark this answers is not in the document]
bkaradzic-microsoft: Addressed in `8cf3958`: primitive values still clear the handler per `[LegacyTreatNonObjectAsNull]`, while non-callable objects now throw `TypeError` during callback conversion and leave the existing handler unchanged. Added focused coverage for that behavior.

Polyfills/XMLHttpRequest/Source/XMLHttpRequest.cpp
```diff
@@ -88,6 +127,15 @@ namespace Babylon::Polyfills::Internal
...
+                // DOM `on<event>` handler properties. Without these, `xhr.onreadystatechange = fn`
+                // silently sets an ordinary expando property that is never invoked, so code written
+                // against the standard XMLHttpRequest API waits forever for a callback that can
+                // never fire.
+                InstanceAccessor("onreadystatechange", &XMLHttpRequest::GetEventHandler<EventIndex::ReadyStateChange>, &XMLHttpRequest::SetEventHandler<EventIndex::ReadyStateChange>),
+                InstanceAccessor("onload", &XMLHttpRequest::GetEventHandler<EventIndex::Load>, &XMLHttpRequest::SetEventHandler<EventIndex::Load>),
+                InstanceAccessor("onerror", &XMLHttpRequest::GetEventHandler<EventIndex::Error>, &XMLHttpRequest::SetEventHandler<EventIndex::Error>),
+                InstanceAccessor("onloadend", &XMLHttpRequest::GetEventHandler<EventIndex::LoadEnd>, &XMLHttpRequest::SetEventHandler<EventIndex::LoadEnd>),
+                InstanceAccessor("onabort", &XMLHttpRequest::GetEventHandler<EventIndex::Abort>, &XMLHttpRequest::SetEventHandler<EventIndex::Abort>),
```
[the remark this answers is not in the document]
bkaradzic-microsoft: Addressed on the current branch: `abort()` records `m_aborted`, and the request continuation dispatches `abort` followed by `loadend` instead of `error` when cancellation settles. The unit suite includes a focused aborted-request regression.

Polyfills/XMLHttpRequest/Source/XMLHttpRequest.cpp
```diff
@@ -59,6 +60,71 @@ namespace Babylon::Polyfills::Internal
...
+            if (it != listeners.end())
+            {
+                listeners.erase(it);
+            }
+
+            return;
+        }
+        if (!value.IsFunction())
+        {
+            throw Napi::TypeError::New(Env(), "Event handler must be callable");
```
bghgary: `[LegacyTreatNonObjectAsNull]` converts non-object values to null, but a non-callable object remains the getter value and performs no operation when dispatched. Don't throw here or pin that behavior in the new test; preserve the assigned object's identity.

bghgary (changes requested):

Concerns inline.

Polyfills/XMLHttpRequest/Source/XMLHttpRequest.cpp
bkaradzic-microsoft: Addressed in `91750df`: abort before send is inert; abort after DONE resets to UNSENT silently; active abort now cancels and synchronously dispatches `readystatechange`, `abort`, and `loadend`, then resets to UNSENT. Each send has a generation guard, and `open()` creates a fresh UrlRequest so the canceled UrlLib source cannot poison reuse.

Polyfills/XMLHttpRequest/Source/XMLHttpRequest.cpp
bkaradzic-microsoft: Addressed in `91750df`: completion no longer clears the listener map. Added a regression that reads back the same `onload` handler after completion and reuses the XHR for a second successful request.

Polyfills/XMLHttpRequest/Source/XMLHttpRequest.cpp
bkaradzic-microsoft: Addressed in `91750df`: dispatch snapshots shared listener records, checks each record's active state immediately before invocation, and reads the record's current callback. Tests cover removal and `onload` reassignment by an earlier listener.

Polyfills/XMLHttpRequest/Source/XMLHttpRequest.cpp
bkaradzic-microsoft: Addressed in `91750df`: callbacks are invoked with the XHR wrapper as `this` and receive an event with `type`, `target`, and `currentTarget`; progress-style terminal events also expose `lengthComputable`, `loaded`, and `total`. Added regression coverage.

Polyfills/XMLHttpRequest/Source/XMLHttpRequest.cpp
bkaradzic-microsoft: Addressed in `91750df`: the README now documents the `on<event>` properties, supported event sequence, abort, and that completed non-2xx responses dispatch `load` rather than `error`.

Polyfills/XMLHttpRequest/Source/XMLHttpRequest.cpp
bkaradzic-microsoft: Addressed in `91750df`: dispatch continues after a throwing listener, captures each exception independently, and queues each one back through JsRuntime so AppRuntime's unhandled-exception handler observes it instead of silently discarding it.

Polyfills/XMLHttpRequest/Source/XMLHttpRequest.cpp
bkaradzic-microsoft: Addressed in `91750df`, correcting my earlier interpretation: primitive assignments clear the handler, while non-callable objects are retained by identity for the getter and skipped during dispatch. The TypeError regression was replaced with the object-identity behavior.

Polyfills/XMLHttpRequest/Source/XMLHttpRequest.cpp
bghgary: This still overwrites a request started reentrantly by a synchronous abort callback. If `readystatechange`, `abort`, or `loadend` calls `open()` and `send()`, `Abort()` unconditionally sets `m_readyState = Unsent` after the callback sequence returns. Guard that final reset with the send generation so the new request keeps its state.

Polyfills/XMLHttpRequest/Source/XMLHttpRequest.cpp
bghgary: This still passes a plain object rather than an `Event` for `readystatechange` or a `ProgressEvent` for terminal events. The fields and receiver are fixed, but `instanceof`, the event prototypes, and standard event methods are still absent.

bghgary (commented):

Two concerns remain.

Polyfills/XMLHttpRequest/Source/XMLHttpRequest.cpp
bkaradzic-microsoft: Fixed in e0e79fc. Abort now captures the invalidated send generation and checks it after readystatechange and abort before dispatching more old-request events; after loadend it resets to UNSENT only if no replacement request was started. The async completion path also skips an old loadend if onload/onerror opens a new request. Added regressions that open and send from each abort callback and from onload.

Polyfills/XMLHttpRequest/Source/XMLHttpRequest.cpp
bkaradzic-microsoft: Fixed in e0e79fc. readystatechange now dispatches an Event, while load/error/abort/loadend dispatch ProgressEvent instances. The event factory retains host constructors where available and installs missing ones; it supplies Event prototype methods, progress fields, target/currentTarget and dispatch-phase state, and stopImmediatePropagation skips later listeners. Added regression coverage for instanceof, methods, receiver and event lifecycle.

Polyfills/XMLHttpRequest/Source/XMLHttpRequest.cpp
```diff
@@ -346,15 +574,70 @@ namespace Babylon::Polyfills::Internal
...
+                unhandledErrors.push_back(std::make_shared<Napi::ObjectReference>(Napi::Persistent(error.Value())));
+                continue;
+            }
+
+            if (env.IsExceptionPending())
+            {
+                auto error = env.GetAndClearPendingException();
+                unhandledErrors.push_back(std::make_shared<Napi::ObjectReference>(Napi::Persistent(error.Value())));
+            }
+            if (dispatch.Get("isStopped").As<Napi::Function>().Call(dispatch, {}).ToBoolean().Value())
```
bghgary: If a listener calls `event.stopImmediatePropagation()` and then throws, the `catch` continues before this flag check, so later listeners still run. Please check the stopped flag after both successful and throwing callbacks, and cover that case.

Polyfills/XMLHttpRequest/Source/XMLHttpRequest.cpp
```diff
@@ -98,6 +249,12 @@ namespace Babylon::Polyfills::Internal
                 InstanceMethod("send", &XMLHttpRequest::Send),
             });
 
+        auto eventFactory = env.RunScript(EVENT_FACTORY_SOURCE, "XMLHttpRequestEvents.js");
```
bghgary: Please update the PR description to match the final implementation: it still shows `Napi::FunctionReference` listeners, says non-callable assignments become `null`, and lists the old test count despite the later changes.

bghgary: This new `Env::RunScript` call does not compile under Node-API-JSI (C2039 in UWP/Win32), and Android Hermes crashes in `napi_run_script` from this initializer (Win32 Hermes also exits with an access violation). Please make event initialization work on these supported backends before merging.

Tests/UnitTests/Scripts/tests.ts
```diff
@@ -160,31 +160,518 @@ describe("XMLHTTPRequest", function () {
...
+                        return;
+                    }
+                    restarted = true;
+                    xhr.open("GET", "app:///Scripts/symlink_target.js");
+                    xhr.send();
+                });
+                xhr.addEventListener("abort", () => { aborts++; });
+                xhr.addEventListener("load", () => { loads++; });
+                xhr.addEventListener("loadend", () => {
+                    if (xhr.status === 200 && restarted) {
```
bghgary: In the `loadend` restart case, the first listener starts the replacement request and this second listener reads `xhr.status` while its worker may still update it. Ubuntu ThreadSanitizer reports this exact GetStatus/UrlRequest race. Avoid reading the in-flight request status during the old event; check it after completion instead.

Polyfills/XMLHttpRequest/Source/XMLHttpRequest.cpp
```diff
@@ -98,6 +249,12 @@ namespace Babylon::Polyfills::Internal
                 InstanceMethod("send", &XMLHttpRequest::Send),
             });
 
+        auto eventFactory = Napi::Eval(env, EVENT_FACTORY_SOURCE, "XMLHttpRequestEvents.js");
```
bghgary: Please update the PR description: its `Listener` snippet and 12-test count no longer match this branch, and it omits the later event-instance and abort reentrancy changes.

Polyfills/XMLHttpRequest/Source/XMLHttpRequest.cpp
```diff
@@ -213,48 +366,90 @@ namespace Babylon::Polyfills::Internal
...
     }
 
-    void XMLHttpRequest::Abort(const Napi::CallbackInfo&)
+    void XMLHttpRequest::Abort(const Napi::CallbackInfo& info)
     {
-        m_request.Abort();
+        if (m_readyState == ReadyState::Done)
+        {
+            m_readyState = ReadyState::Unsent;
+            return;
```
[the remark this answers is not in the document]
bkaradzic-microsoft: Fixed in 07f1dab. Aborting a completed request now replaces its UrlRequest (preserving responseType), clears the status snapshots, and returns to UNSENT without retaining the previous body, response URL, headers, or transport diagnostics. Buffer responses are null outside DONE. Added a regression that completes a request, aborts it, checks the cleared fields, and checks arraybuffer responseType preservation. Local JavaScriptCore and QuickJS suites pass (247 JavaScript and 24 native tests each); CI for this push is pending.

Polyfills/XMLHttpRequest/Source/XMLHttpRequest.cpp
```diff
@@ -59,6 +60,147 @@ namespace Babylon::Polyfills::Internal
...
+    void XMLHttpRequest::SetEventHandler(const Napi::CallbackInfo&, const Napi::Value& value)
+    {
+        auto& listeners = m_listeners[EVENT_TYPE_NAMES[static_cast<size_t>(Index)]];
+        const auto it = std::find_if(listeners.begin(), listeners.end(), [](const std::shared_ptr<Listener>& listener) {
+            return listener->isEventHandler;
+        });
+
+        // Event handler attributes treat primitive values as null. Object values are retained
+        // verbatim for the getter; non-callable objects are simply skipped during dispatch.
+        if (!value.IsObject())
```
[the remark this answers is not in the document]
bkaradzic-microsoft: Fixed in 07f1dab. The on<event> setter now throws TypeError for a non-callable object before changing the listener list, while primitive values still clear the handler per LegacyTreatNonObjectAsNull. Updated the regression to verify a failed object assignment preserves an existing handler and leaves a new one null. Local JavaScriptCore and QuickJS suites pass (247 JavaScript and 24 native tests each); CI is pending.

Polyfills/XMLHttpRequest/Source/XMLHttpRequest.cpp
```diff
@@ -310,50 +514,131 @@ namespace Babylon::Polyfills::Internal
...
+
+            try
+            {
+                callback.As<Napi::Function>().Call(jsThis, {event});
+            }
+            catch (const Napi::Error& error)
+            {
+                unhandledErrors.push_back(std::make_shared<Napi::Error>(error));
+                continue;
+            }
```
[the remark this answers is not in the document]
bkaradzic-microsoft: Fixed in 07f1dab. The catch path now records the listener error but still checks isStopped before proceeding, so stopImmediatePropagation remains effective even when the callback throws. Added a native regression that checks the later listener is skipped and the error still reaches the host unhandled-exception handler. Local JavaScriptCore and QuickJS suites pass (247 JavaScript and 24 native tests each); CI is pending.

Polyfills/XMLHttpRequest/Source/XMLHttpRequest.h
```diff
@@ -39,19 +41,55 @@ namespace Babylon::Polyfills::Internal
...
+
+        // A registered event listener. `isEventHandler` marks the single entry owned by the
+        // matching `on<event>` property; every other entry came from addEventListener. Both
+        // kinds share one list per event type because that is what the DOM specifies: dispatch
+        // follows registration order, so `addEventListener("load", a)` then `xhr.onload = b`
+        // calls `a` then `b`, and reassigning `onload` keeps its original position rather than
+        // moving to the end ("If eventHandler's listener is not null, then return").
+        struct Listener
+        {
+            Napi::ObjectReference callback;
```
bghgary: Please update the PR description's obsolete `Listener` snippet and 12-test count, and describe the `error`-to-`abort` change as behavioral rather than purely additive.

Polyfills/XMLHttpRequest/Source/XMLHttpRequest.cpp
```diff
@@ -263,7 +473,16 @@ namespace Babylon::Polyfills::Internal
...
         try
         {
-            m_request.Open(MethodType::StringToEnum(info[0].As<Napi::String>().Utf8Value()), m_url);
+            if (m_sendActive)
+            {
+                m_request->Abort();
+                m_sendActive = false;
+            }
+            ++m_sendId;
+            m_request = std::make_shared<UrlLib::UrlRequest>();
```
bghgary: Reopening replaces `m_request`, resetting a previously selected `responseType` to `"text"`. Preserve `responseType` across request replacement and cover an `"arraybuffer"` request followed by reuse.

Polyfills/XMLHttpRequest/Source/XMLHttpRequest.cpp
```diff
@@ -263,7 +473,16 @@ namespace Babylon::Polyfills::Internal
...
+            m_request->Open(MethodType::StringToEnum(info[0].As<Napi::String>().Utf8Value()), m_url);
```
bghgary: If `open()` throws after a completed response, the new request is empty while `status` and `statusText` still describe the old response. Keep the request and its snapshots consistent when opening fails, and cover that failure path.

Polyfills/XMLHttpRequest/Source/XMLHttpRequest.cpp
```diff
@@ -213,48 +375,96 @@ namespace Babylon::Polyfills::Internal
...
+        }
+
+        m_sendActive = false;
+        const auto abortedSendId = ++m_sendId;
+        m_statusCode = 0;
+        m_statusText.clear();
+        m_request->Abort();
+
+        auto jsThis = info.This().As<Napi::Object>();
+        SetReadyState(ReadyState::Done, jsThis);
```
bghgary: Please make `response` and `responseText` safe to read in synchronous `onabort`: `UrlRequest::Abort()` only signals cancellation, so these getters can read a buffer its worker is still writing. Avoid reading the in-flight buffer during abort and cover this under ThreadSanitizer.

Polyfills/XMLHttpRequest/Source/XMLHttpRequest.cpp
```diff
@@ -263,7 +473,16 @@ namespace Babylon::Polyfills::Internal
 
         try
         {
-            m_request.Open(MethodType::StringToEnum(info[0].As<Napi::String>().Utf8Value()), m_url);
+            if (m_sendActive)
+            {
+                m_request->Abort();
+                m_sendActive = false;
+            }
+            ++m_sendId;
```
bghgary: Validate the new `open()` arguments before canceling an active request and advancing its send ID. `open("INVALID", url)` currently throws after suppressing that request's completion, so its listeners never settle. Please cover this case.

Polyfills/XMLHttpRequest/Source/XMLHttpRequest.cpp
```diff
@@ -310,50 +529,132 @@ namespace Babylon::Polyfills::Internal
...
+                    return;
+                }
                 if (failed)
                 {
-                    RaiseEvent(EventType::Error);
+                    RaiseEvent(EventType::Error, jsThis);
+                }
+                else
+                {
+                    RaiseEvent(EventType::Load, jsThis);
```
bghgary: On Android, UrlLib sets status before reading the body and can swallow a Java read exception without clearing it. A truncated 200 response therefore reaches this success branch and fires the new `load` event. Please ensure UrlLib reports body-read failure before dispatching `load`.

Polyfills/XMLHttpRequest/Source/XMLHttpRequest.cpp
```diff
@@ -213,48 +375,96 @@ namespace Babylon::Polyfills::Internal
...
+                // Per DOM, re-adding an identical (type, callback, capture) triple is a silent
+                // no-op rather than an error: "If eventTarget's event listener list does not
+                // contain an event listener whose type is listener's type [...] then append
+                // listener". The listener stays registered once and is dispatched once.
+                return;
             }
         }
 
-        m_eventHandlerRefs[eventType].push_back(Napi::Persistent(eventHandler));
+        listeners.push_back(std::make_shared<Listener>(Listener{Napi::Persistent(eventHandler.As<Napi::Object>()), false}));
```
bghgary: A handler that captures its own XHR remains strongly rooted through this persistent callback after the page drops the XHR; completion no longer releases these references, so neither the wrapper nor its native owner can be collected. Please preserve reusable listeners without permanently retaining unreachable requests, and cover a self-capturing handler.

Polyfills/XMLHttpRequest/Source/XMLHttpRequest.cpp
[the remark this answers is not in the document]
bghgary: No change needed to the original object behavior: Web IDL's `[LegacyTreatNonObjectAsNull]` retains non-callable objects as inert callbacks and returns the same object from the getter; only non-objects become `null`. The new `TypeError` branch and test reverse that requirement. Please restore the object-identity behavior; the JSI exception-bridge fix does not address the specification mismatch.

PR #223, opened 2026-08-13
Fix use-after-free: keep escaped handles alive when their escapable scope closes

bkaradzic-microsoft:
[Updated by Copilot on behalf of @bghgary]

`napi_escape_handle` inserted the escaped handle at the scope start index so it would live in the parent scope, but `napi_close_escapable_handle_scope` recomputed `scope_start` from the token and called `resize(scope_start)`, freeing the very handle the close was supposed to preserve. Every caller of `napi_escape_handle` got a dangling `napi_value` back.

This is reachable from ordinary code, not just direct N-API use. `Napi::ObjectReference::Get` uses an `EscapableHandleScope` and `Napi::Error::Message()` / `what()` are built on it, so reading the message of a native error on QuickJS was a heap-use-after-free. `Napi::FunctionReference::Call` and `MakeCallback` escape as well, which puts every WebSocket, `setTimeout`, `XMLHttpRequest` and `AbortSignal` callback on this path: instrumenting `napi_escape_handle` counted ~201 escapes in a single `JavaScript.All` run with no escape-specific test in scope.

**How it was found.** BabylonNative [#1835](https://github.com/BabylonJS/BabylonNative/pull/1835) adds tests that make a native module throw. `ExternalCallback::Callback` calls `e.what()` when there is no pending QuickJS exception, walking straight into the freed handle; its `Ubuntu_Clang_QuickJS` job segfaulted while every other engine and platform passed.

```
#0 ToJSValue                          js_native_api_quickjs.cc:302
#3 Napi::Error::Message
#4 Napi::Error::what
#5 ExternalCallback::Callback          js_native_api_quickjs.cc:164
freed by:
#1 napi_close_escapable_handle_scope   js_native_api_quickjs.cc:1939
#2 Napi::ObjectReference::Get
```

## The change

Each open escapable scope gets a record on the env, keyed by a monotonic counter that is handed out as the opaque token. The escaped handle lives in that record until `napi_close_escapable_handle_scope` pushes it onto the handle stack, once the scope's own handles are gone; it lands at `scope_start`, in the parent scope, so it outlives the close.

The token is a counter rather than a position because two escapable scopes opened with no handle allocated between them occupy the same position. Keyed on that, their escaped handles collide and the second scope to escape is refused with `napi_escape_called_twice` having never escaped.

The handle stack is never modified in the middle, which matters: inserting at `scope_start` shifts every entry above it and invalidates the recorded start of any nested scope still open, reintroducing the same dangling value by a different route.

A close whose recorded start is past the end of the stack now reports `napi_handle_scope_mismatch` rather than resizing, which previously grew the stack with null entries for the next close to dereference. Env teardown frees handles still held for scopes that were never closed.

`napi_open_handle_scope` keeps its position-derived token: a position is all a regular scope needs, and its comment now says not to key per-scope state on it, which is the mistake the escapable version made.

## Chakra and JavaScriptCore

Both returned the escapee without tracking scopes, so neither could report `napi_escape_called_twice`. Both now track open escapable scopes; values there are rooted independently of any scope, so this is the error contract only. That removes the need for `JSRUNTIMEHOST_NAPI_ESCAPE_HANDLE_IS_PASSTHROUGH`, so `SecondEscapeIsRejected` runs on every backend rather than being compiled out on two of them.

## Testing

Four tests in `Tests/UnitTests/Shared/Shared.cpp`:

- `EscapedHandleOutlivesItsScope` — reproduces the original `heap-use-after-free` under ASan without the fix.
- `NestedEscapableScopesBothEscape` — fails on every run against the pre-fix implementation.
- `SecondEscapeIsRejected` — the `napi_escape_called_twice` contract.
- `AdjacentEscapableScopesEscapeIndependently` — two scopes with no handle allocated between them; confirmed to fail against the position-derived token and pass with the counter.

Each test closes its escapable scopes on every exit path. Leaving one open made the enclosing `Napi::HandleScope` fail to close, and `Napi::Error::Fatal` throws from a destructor that is implicitly `noexcept`, so a failing assertion terminated the process instead of reporting `FAILED`.

Verified locally at this head on Windows Release: QuickJS 10/10 and Chakra 10/10. V8, JavaScriptCore and Hermes are covered by CI.

The BabylonNative #1835 end-to-end run (clang + QuickJS + RelWithDebInfo, changing only this dependency: `master` gives exit 139, 1, 139; this branch gives exit 0 × 5, clean 16/16) was made against `6238b5ab`, before the scope-identity change.

### DIFF
--- a/Core/Node-API/Source/js_native_api_chakra.cc
+++ b/Core/Node-API/Source/js_native_api_chakra.cc
@@ -1929,8 +1929,9 @@ napi_status napi_open_escapable_handle_scope(
   napi_escapable_handle_scope* result) {
   CHECK_ENV(env);
   CHECK_ARG(env, result);
-  *result = reinterpret_cast<napi_escapable_handle_scope>(
-    ++env->next_escapable_scope_token);
+  const size_t token = ++env->next_escapable_scope_token;
+  env->open_escapable_scopes.emplace(token, false);
+  *result = reinterpret_cast<napi_escapable_handle_scope>(token);
   return napi_ok;
 }
 
@@ -1940,7 +1941,11 @@ napi_status napi_close_escapable_handle_scope(
   napi_escapable_handle_scope scope) {
   CHECK_ENV(env);
   CHECK_ARG(env, scope);
-  env->escaped_scopes.erase(reinterpret_cast<size_t>(scope));
+  const auto it = env->open_escapable_scopes.find(reinterpret_cast<size_t>(scope));
+  if (it == env->open_escapable_scopes.end()) {
+    return napi_set_last_error(env, napi_invalid_arg);
+  }
+  env->open_escapable_scopes.erase(it);
   return napi_ok;
 }
 
@@ -1955,9 +1960,14 @@ napi_status napi_escape_handle(napi_env env,
   CHECK_ARG(env, scope);
   CHECK_ARG(env, escapee);
   CHECK_ARG(env, result);
-  if (!env->escaped_scopes.insert(reinterpret_cast<size_t>(scope)).second) {
+  const auto it = env->open_escapable_scopes.find(reinterpret_cast<size_t>(scope));
+  if (it == env->open_escapable_scopes.end()) {
+    return napi_set_last_error(env, napi_invalid_arg);
+  }
+  if (it->second) {
     return napi_set_last_error(env, napi_escape_called_twice);
   }
+  it->second = true;
   *result = escapee;
   return napi_ok;
 }
--- a/Core/Node-API/Source/js_native_api_chakra.cc
+++ b/Core/Node-API/Source/js_native_api_chakra.cc
@@ -1929,7 +1929,8 @@ napi_status napi_open_escapable_handle_scope(
   napi_escapable_handle_scope* result) {
   CHECK_ENV(env);
   CHECK_ARG(env, result);
-  *result = reinterpret_cast<napi_escapable_handle_scope>(1);
+  *result = reinterpret_cast<napi_escapable_handle_scope>(
+    ++env->next_escapable_scope_token);
   return napi_ok;
 }
 
@@ -1939,11 +1940,13 @@ napi_status napi_close_escapable_handle_scope(
   napi_escapable_handle_scope scope) {
   CHECK_ENV(env);
   CHECK_ARG(env, scope);
+  env->escaped_scopes.erase(reinterpret_cast<size_t>(scope));
   return napi_ok;
 }
 
-// Stub implementation of handle scope apis for JSRT.
-// This one will return escapee value as this is called from leveldown db.
+// JSRT roots values independently of any scope, so the escapee is returned as
+// is. The scope is still tracked so a second escape is rejected as Node-API
+// requires.
 napi_status napi_escape_handle(napi_env env,
                                napi_escapable_handle_scope scope,
                                napi_value escapee,
@@ -1952,6 +1955,9 @@ napi_status napi_escape_handle(napi_env env,
   CHECK_ARG(env, scope);
   CHECK_ARG(env, escapee);
   CHECK_ARG(env, result);
+  if (!env->escaped_scopes.insert(reinterpret_cast<size_t>(scope)).second) {
+    return napi_set_last_error(env, napi_escape_called_twice);
+  }
   *result = escapee;
   return napi_ok;
 }
--- a/Core/Node-API/Source/js_native_api_quickjs.cc
+++ b/Core/Node-API/Source/js_native_api_quickjs.cc
@@ -1917,9 +1917,13 @@ napi_status napi_open_escapable_handle_scope(napi_env env, napi_escapable_handle
   CHECK_ENV(env);
   CHECK_ARG(env, result);
   
-  // Same as regular handle scope for QuickJS
   env->current_scope_start = env->handle_scope_stack.size();
-  *result = reinterpret_cast<napi_escapable_handle_scope>(env->current_scope_start + 1);
+  
+  // The token is a counter, not a position: scopes opened with no handle allocated
+  // between them share a position and would otherwise be indistinguishable.
+  const size_t token = ++env->next_escapable_scope_token;
+  env->escapable_scopes.emplace(token, napi_env__::EscapableScope{env->current_scope_start, nullptr});
+  *result = reinterpret_cast<napi_escapable_handle_scope>(token);
   
   napi_clear_last_error(env);
   return napi_ok;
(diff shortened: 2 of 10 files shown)


bkaradzic-microsoft:
Good catch, thanks. Reworked the test to report through the promise and return early, matching the convention the other Dispatch-based tests here use. As you note it also removes a real hazard: an assertion failure in the setup calls would previously have left the promise unset and hung the waiter rather than failing the test.

Also dropped the double-escape assertion in a separate commit. The JavaScriptCore shim does not implement the `napi_escape_called_twice` check, so it failed \Ubuntu_clang\. That is a separate gap from the use-after-free fixed here, so the test now covers only the portable contract. The QuickJS implementation still returns `napi_escape_called_twice`, which it needs to, since a second escape would insert a second handle while close only preserves one.

bkaradzic-microsoft:
The nested-scope report was correct, and it was a real bug rather than a theoretical one. Thanks — I've reworked the fix.

**What was wrong.** `napi_escape_handle` inserted the escaped handle into `handle_scope_stack` at `scope_start`. Scope tokens are derived from the stack size at open, so that insert shifts every entry above it and silently invalidates the recorded start of any nested scope that is still open. Closing that inner scope then keeps the wrong slot and frees the escaped handle — which reproduces exactly the dangling `napi_value` this PR set out to fix.

**How it's fixed.** Rather than patching up the indices, escaped handles are now held aside in a map on the env keyed by scope start, and pushed onto the stack in `napi_close_escapable_handle_scope` once the scope's own handles are gone. They land at `scope_start`, in the parent scope, so they still outlive the close — but the stack is never modified in the middle, so no index is ever invalidated and the whole class of bug goes away. The result is a bit smaller than what it replaces. Env teardown now also frees handles still held aside for scopes that were never closed, which the stack-based version got for free.

`current_scope_start` was being adjusted to compensate for the shift. It turns out to be written in several places and never read anywhere in the repo, so that bookkeeping went away with the insert.

**Tests.** Both suggestions are in:

- `NestedEscapableScopesBothEscape` — escapes from an inner and an outer scope and reads both back. I verified it's a real guard: against the previous implementation it fails on every run; against this one it passes.
- `SecondEscapeIsRejected` — covers the `napi_escape_called_twice` contract.

QuickJS and V8 are both 9/9 locally.

One pre-existing limitation I'll note but haven't changed here: two scopes opened at the same stack size get identical tokens and are indistinguishable. Fixing that needs stable scope IDs instead of index-derived tokens, which is a bigger change than this PR should carry.

bkaradzic-microsoft:
Some real-world evidence for this fix, which I ran into by accident.

BabylonNative [#1835](https://github.com/BabylonJS/BabylonNative/pull/1835) has been failing `Ubuntu_Clang_QuickJS` consistently. It turns out to be this exact bug, hit through a path nobody writes by hand:

```
#0  js_force_tostring                       quickjs.c:4813
#3  napi_get_value_string_utf8              js_native_api_quickjs.cc:696
#6  Napi::Error::Message                    napi-inl.h:3087
#7  Napi::Error::what
#8  ExternalCallback::Callback              js_native_api_quickjs.cc:164
```

`Napi::Error::what()` reads the error's `message`, and `ObjectReference::Get` fetches properties through an escapable handle scope:

```cpp
inline MaybeOrValue<Napi::Value> ObjectReference::Get(const char* utf8name) const {
  EscapableHandleScope scope(_env);
  ...
  return scope.Escape(result);
}
```

So any native code that throws a `Napi::Error` which reaches the shim's catch-all lands on the escaped handle after its scope has closed. At the crash the JSValue reads `tag = -7` (string) with `ptr = 0x7ff8dec9a216` — not pointer-aligned, i.e. reused memory.

That makes this reachable from ordinary `Napi::Error` use rather than only from addons that explicitly open escapable scopes, which I think raises the priority a bit.

**A/B on the same tree**, CI flags (clang, QuickJS, RelWithDebInfo, no sanitizers), only this dependency swapped:

| JsRuntimeHost | Result |
| --- | --- |
| master | exit 139, 1, 139 — segfault |
| this branch | exit 0 × 5 — clean, 16/16 |

Worth noting the crash is intermittent in the way use-after-free usually is (139/1/139), so it reads as flaky CI rather than as a clear bug.

Core/Node-API/Source/js_native_api_quickjs.cc
```diff
@@ -1929,14 +1933,28 @@ napi_status napi_close_escapable_handle_scope(napi_env env, napi_escapable_handl
...
+    return napi_set_last_error(env, napi_invalid_arg);
+  }
+  
+  const size_t scope_start = it->second.scope_start;
   
   for (size_t i = scope_start; i < env->handle_scope_stack.size(); i++) {
     JS_FreeValue(env->context, *env->handle_scope_stack[i]);
   }
   
   env->handle_scope_stack.resize(scope_start);
```
bghgary: `resize` grows the stack with null `unique_ptr`s when `scope_start` is past the end, and the next close dereferences them. Reaching it needs scopes closed out of order, which Node-API forbids, and `napi_close_handle_scope` has the same shape.

Core/Node-API/Source/js_native_api_chakra.cc
```diff
@@ -1952,6 +1955,9 @@ napi_status napi_escape_handle(napi_env env,
   CHECK_ARG(env, scope);
   CHECK_ARG(env, escapee);
   CHECK_ARG(env, result);
+  if (!env->escaped_scopes.insert(reinterpret_cast<size_t>(scope)).second) {
```
bghgary: This inserts on any token value, so an escape against a stale or never-opened scope reports success and leaves an entry that only a matching close removes. QuickJS returns `napi_invalid_arg` for a token it does not know. The JavaScriptCore copy is the same.

bghgary:
[Commented by Copilot on behalf of @bghgary]

I pushed four commits here rather than leaving review comments. Two of them change decisions you had made, so calling those out explicitly:

- The escaped-handle map was keyed on the scope start index. Two escapable scopes opened with no handle allocated between them share that index, so their entries collided and the second scope to escape was refused with `napi_escape_called_twice` having never escaped. Scopes now get a record keyed by a counter, and `AdjacentEscapableScopesEscapeIndependently` covers it.
- Chakra and JavaScriptCore now track open escapable scopes, which deletes `JSRUNTIMEHOST_NAPI_ESCAPE_HANDLE_IS_PASSTHROUGH` and lets `SecondEscapeIsRejected` run on every backend instead of being compiled out on two.

The description is updated to match. Say if you disagree with either, particularly the second, since it widens the change past QuickJS.

bghgary (approved):

Pushed to this branch rather than leaving these as asks.

The scope token came from the handle-stack position, so two escapable scopes opened with no handle between them shared a key and the second to escape was refused with `napi_escape_called_twice` having never escaped; scopes now get a record keyed by a counter. The tests closed their scopes only on success, turning a failing assertion into process termination with no `FAILED` line; they now close on every path. Chakra and JavaScriptCore track open scopes, which deletes `JSRUNTIMEHOST_NAPI_ESCAPE_HANDLE_IS_PASSTHROUGH`, lets `SecondEscapeIsRejected` run on every backend, and bounds state that previously only a matching close could remove. A non-LIFO close no longer resizes the QuickJS stack past its end.

QuickJS 10/10, Chakra 10/10 on Windows Release; the new test fails against the pre-fix code. JavaScriptCore does not build here, so its change rests on CI.

`napi_open_handle_scope` keeps its position-derived token. A position is all a regular scope needs, because nothing stores per-scope state against it: the token only says where the close truncates, and two closes truncating to the same index is a no-op. Giving it a counter would mean recovering the position from a lookup, so a token this code has never seen becomes an error return, and `HandleScope`'s destructor turns any non-`napi_ok` into `Error::Fatal` — a new abort path on a scope opened per `Dispatch`, for no behavioural change. Its comment now records that the token is a position and must not be keyed on, which is the mistake the escapable version made.

PR #224, opened 2026-08-20
Add JsConsoleLogger so native code can log to the JS console

bkaradzic-microsoft:
Groundwork for a BabylonNative follow-up to BabylonJS/BabylonNative#1824, where this was agreed in review.

## Why

Native diagnostics written to `stdout`/`stderr` are invisible on the platforms where they matter most: on Android and iOS there is no attached terminal, so the message reaches nobody. Routing them through the JS console instead puts native messages wherever the host has already directed the script's own.

BabylonNative has had exactly this utility for a while, but it lives in `Plugins/NativeEngine/Source/`, which is private to that plugin, so nothing else can use it. The Canvas polyfill writes its font and diagnostic warnings to `stderr` for that reason alone.

## What

Moved it to `Core/JsRuntime`, where it needs only `napi` and any consumer can reach it. The BabylonNative side (bumping the pin, deleting its private copy, and routing the Canvas warnings through it) follows separately.

Behavior is unchanged, with one exception: `LogMethod` was a namespace-scope function with external linkage that was declared in no header, so it was an unnecessary exported symbol and a potential ODR collision. It is now in an anonymous namespace.

## Tests

Added two, since the type had none:

- `JsConsoleLogger.RoutesToConsole` — asserts all three methods reach the Console polyfill callback with the right `LogLevel` and message, in order.
- `JsConsoleLogger.NoConsoleIsNotFatal` — the documented no-op path, covering both no `console` at all and a `console` whose `warn` is not a function.

Full suite locally (Win32/Chakra): 12/12 gtest suites (was 10), 216 JS assertions passing.

### DIFF
--- a/Core/JsRuntime/Include/Babylon/JsConsoleLogger.h
+++ b/Core/JsRuntime/Include/Babylon/JsConsoleLogger.h
@@ -1,0 +1,25 @@
+#pragma once
+
+#include <napi/env.h>
+
+namespace Babylon
+{
+    /**
+     * Utility struct to log messages to the JavaScript console.
+     *
+     * Native code that writes diagnostics to stdout/stderr is invisible on the platforms
+     * where diagnostics matter most: on Android and iOS there is no attached terminal, so
+     * the message reaches nobody. Routing through the JS console instead puts native
+     * messages in the same place as the script's own, wherever the host has directed it.
+     *
+     * Each call is a no-op if no console object, or no such method on it, is present.
+     */
+    struct JsConsoleLogger final
+    {
+        JsConsoleLogger() = delete;
+
+        static void LogInfo(Napi::Env env, const char* message);
+        static void LogWarn(Napi::Env env, const char* message);
+        static void LogError(Napi::Env env, const char* message);
+    };
+}
--- a/Core/JsRuntime/Source/JsConsoleLogger.cpp
+++ b/Core/JsRuntime/Source/JsConsoleLogger.cpp
@@ -1,0 +1,39 @@
+#include <Babylon/JsConsoleLogger.h>
+
+namespace Babylon
+{
+    namespace
+    {
+        void LogMethod(Napi::Env env, const char* methodName, const char* message)
+        {
+            auto console = env.Global().Get("console");
+
+            if (console.IsObject())
+            {
+                auto consoleLog{console.ToObject().Get(methodName)};
+
+                if (consoleLog.IsFunction())
+                {
+                    auto consoleLogFunction = consoleLog.As<Napi::Function>();
+                    auto messageStr = Napi::String::New(env, message);
+                    consoleLogFunction.Call(console, {messageStr});
+                }
+            }
+        }
+    }
+
+    void JsConsoleLogger::LogInfo(Napi::Env env, const char* message)
+    {
+        LogMethod(env, "log", message);
+    }
+
+    void JsConsoleLogger::LogWarn(Napi::Env env, const char* message)
+    {
+        LogMethod(env, "warn", message);
+    }
+
+    void JsConsoleLogger::LogError(Napi::Env env, const char* message)
+    {
+        LogMethod(env, "error", message);
+    }
+}
--- a/Core/JsRuntime/Source/JsConsoleLogger.cpp
+++ b/Core/JsRuntime/Source/JsConsoleLogger.cpp
@@ -6,18 +6,35 @@ namespace Babylon
     {
         void LogMethod(Napi::Env env, const char* methodName, const char* message)
         {
-            auto console = env.Global().Get("console");
-
-            if (console.IsObject())
+            try
             {
-                auto consoleLog{console.ToObject().Get(methodName)};
+                auto console = env.Global().Get("console");
 
-                if (consoleLog.IsFunction())
+                if (console.IsObject())
                 {
-                    auto consoleLogFunction = consoleLog.As<Napi::Function>();
-                    auto messageStr = Napi::String::New(env, message);
-                    consoleLogFunction.Call(console, {messageStr});
+                    auto consoleLog{console.ToObject().Get(methodName)};
+
+                    if (consoleLog.IsFunction())
+                    {
+                        auto consoleLogFunction = consoleLog.As<Napi::Function>();
+                        auto messageStr = Napi::String::New(env, message);
+                        consoleLogFunction.Call(console, {messageStr});
+                    }
                 }
+            }
+            catch (...)
+            {
+            }
+
+            // Every step above can fail on script the host does not control: `console` and the
+            // method can be accessors that throw, and the call itself is arbitrary user code.
+            // N-API also leaves a pending exception on `env` independently of throwing a C++
+            // exception, so swallowing the C++ side is not enough. Returning with one pending
+            // would surface the failure at some unrelated later point in the caller, which is
+            // never an acceptable outcome for a diagnostic helper.
+            if (env.IsExceptionPending())
+            {
+                (void)env.GetAndClearPendingException();
             }
         }
     }
--- a/Tests/UnitTests/Shared/Shared.cpp
+++ b/Tests/UnitTests/Shared/Shared.cpp
@@ -220,12 +220,21 @@ TEST(JsConsoleLogger, ThrowingConsoleLeavesNoPendingException)
     runtime.Dispatch([&result](Napi::Env env) mutable {
         std::string failures{};
 
+        // Everything here is built out of plain JS rather than Node-API host functions.
         // Napi::Object::DefineProperty and Napi::PropertyDescriptor do not exist in the JSI
-        // Node-API port, so go through JS's own Object.defineProperty, which every engine has.
+        // port at all, and throwing out of a Napi::Function callback aborts the process there
+        // ("Fatal error in v8::ToLocalChecked"), so the throwers have to be real JS functions.
         // Redefining is also why plain Set will not do: once `console` is an accessor without
         // a setter, assigning to it is either silently dropped or a throw.
+        auto functionCtor = env.Global().Get("Function").As<Napi::Function>();
         auto objectCtor = env.Global().Get("Object").As<Napi::Object>();
         auto defineProperty = objectCtor.Get("defineProperty").As<Napi::Function>();
+
+        auto makeThrower = [&env, &functionCtor](const char* what) {
+            return functionCtor
+                .New({Napi::String::New(env, std::string{"throw new Error('"} + what + "');")})
+                .As<Napi::Function>();
+        };
         auto defineConsole = [&env, &defineProperty](Napi::Object descriptor) {
             descriptor.Set("configurable", Napi::Boolean::New(env, true));
             defineProperty.Call({env.Global(), Napi::String::New(env, "console"), descriptor});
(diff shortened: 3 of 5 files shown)


Tests/UnitTests/Shared/Shared.cpp
```diff
@@ -139,6 +142,70 @@ TEST(Console, Log)
...
+    });
+
+    Babylon::ScriptLoader loader{runtime};
+    loader.Dispatch([&done](auto) {
+        done.set_value();
+    });
+    done.get_future().get();
+
+    SUCCEED();
+}
```
bghgary: `NoConsoleIsNotFatal` passes whether or not these calls throw, so it cannot catch the regression it is named for. `AppRuntime::Dispatch` catches `Napi::Error` and hands it to `UnhandledExceptionHandler` (`Core/AppRuntime/Source/AppRuntime.cpp:119-131`), the default handler only prints, and the only assertion here is `SUCCEED()`.

I removed the `IsObject()` and `IsFunction()` guards from `LogMethod` and rebuilt (Win32 x64 Chakra): this test still passed. With the assertion below it fails on `Which is: 1`, and still passes against the code as authored.

One limit worth knowing about the fix: all four calls share a single `Dispatch` lambda, so the first throw skips the remaining three and the count only ever reaches 1.

```suggestion
TEST(JsConsoleLogger, NoConsoleIsNotFatal)
{
    // Every method is documented as a no-op when there is no console object, or no such
    // method on it. Nothing here installs the Console polyfill.
    int unhandled{0};

    Babylon::AppRuntime::Options options{};
    options.UnhandledExceptionHandler = [&unhandled](const Napi::Error&) { ++unhandled; };

    Babylon::AppRuntime runtime{options};

    std::promise<void> done;

    runtime.Dispatch([](Napi::Env env) mutable {
        Babylon::JsConsoleLogger::LogInfo(env, "dropped");
        Babylon::JsConsoleLogger::LogWarn(env, "dropped");
        Babylon::JsConsoleLogger::LogError(env, "dropped");

        // A console whose methods are not functions must be tolerated too.
        auto console = Napi::Object::New(env);
        console.Set("warn", Napi::Number::New(env, 42));
        env.Global().Set("console", console);
        Babylon::JsConsoleLogger::LogWarn(env, "dropped");
    });

    Babylon::ScriptLoader loader{runtime};
    loader.Dispatch([&done](auto) {
        done.set_value();
    });
    done.get_future().get();

    EXPECT_EQ(unhandled, 0);
}
```

Core/JsRuntime/Source/JsConsoleLogger.cpp
```diff
@@ -0,0 +1,39 @@
...
+                auto consoleLog{console.ToObject().Get(methodName)};
+
+                if (consoleLog.IsFunction())
+                {
+                    auto consoleLogFunction = consoleLog.As<Napi::Function>();
+                    auto messageStr = Napi::String::New(env, message);
+                    consoleLogFunction.Call(console, {messageStr});
+                }
+            }
+        }
```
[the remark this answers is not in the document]
bghgary: This does not apply in this build configuration. `napi.h:59-66` defines `NAPI_CPP_EXCEPTIONS` whenever `_CPPUNWIND`/`__EXCEPTIONS` is set and neither macro is pre-defined; nothing in the repo defines `NAPI_DISABLE_CPP_EXCEPTIONS`, and the generated `JsRuntime.vcxproj` builds with `ExceptionHandling=Sync` (`/EHsc`). Under that, `NAPI_THROW_IF_FAILED` is `throw Napi::Error::New(env)` (`napi.h:95-96`), and `Error::New` calls `napi_get_and_clear_last_exception` (`napi-inl.h:2896-2900`) — the pending exception is cleared and moved into the thrown object, so `env` is not left carrying one.

`CaptureCurrentJsStack` needs its explicit clear because it wraps its N-API calls in `catch (...)` and swallows the C++ exception. `LogMethod` has no catch, so nothing is swallowed.

What does happen is that a C++ `Napi::Error` propagates to the caller. Verified by removing the `IsObject()`/`IsFunction()` guards and observing it arrive at `AppRuntime::Dispatch`'s `catch (const Napi::Error&)` — which a silently-pending exception could not have reached.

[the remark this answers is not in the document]
bkaradzic-microsoft: Agreed, and the precedent you point at is the right one -- fixed in e0ed6a1.

I wrapped the whole lookup-and-call in `try`/`catch (...)` *and* cleared any exception left pending on `env`. Both are needed, and the test I added shows why. In this build node-addon-api converts the failure into a C++ throw and clears the pending exception as part of that conversion, so `IsExceptionPending()` alone observes nothing -- the escaping C++ exception is the real failure mode. In a build without C++ exceptions there is no conversion and the exception stays pending instead, which is the case you described.

Worth recording that my first attempt at the test installed the throwing accessor from JS via `loader.Eval` and passed against the *unfixed* code, so it was proving nothing. Installing the accessor from C++ inside the same dispatch makes it deterministic, and reverting the guard now fails it with `throwing console getter escaped as a C++ exception; throwing console.warn escaped as a C++ exception;`, covering both the accessor and the call.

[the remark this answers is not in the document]
bkaradzic-microsoft: Follow-up on this thread, since getting the test to run everywhere turned up something worth recording.

`Win32_x64_JSI` failed three times on the way here, each for a different reason:

1. `Napi::Object::DefineProperty` and `Napi::PropertyDescriptor` do not exist in the JSI Node-API port, so it would not compile. Went through JS's own `Object.defineProperty` instead.
2. Throwing out of a `Napi::Function` callback aborts the process under JSI (`Fatal error in v8::ToLocalChecked: Empty MaybeLocal`), so the throwers could not be host functions. Built them as real JS functions via `new Function(...)`.
3. Even then it aborts -- the shim raises the same fatal error for **any** JS exception raised through a native property access, so the process dies inside `env.Global().Get("console")` before `LogMethod` gets control.

(3) is not something this PR can defend against: there is no return path from the abort. So the test now carries a `JSRUNTIMEHOST_NAPI_ENGINE_JSI` opt-out matching the two that already exist in this file, and it covers Chakra, V8, JavaScriptCore and QuickJS. The guard in `LogMethod` stays unconditional.

Worth being explicit that this means a throwing `console` is fatal under JSI regardless of what any caller does. Happy to file that separately if it is not already known.

Core/JsRuntime/Include/Babylon/JsConsoleLogger.h
```diff
@@ -0,0 +1,25 @@
...
+{
+    /**
+     * Utility struct to log messages to the JavaScript console.
+     *
+     * Native code that writes diagnostics to stdout/stderr is invisible on the platforms
+     * where diagnostics matter most: on Android and iOS there is no attached terminal, so
+     * the message reaches nobody. Routing through the JS console instead puts native
+     * messages in the same place as the script's own, wherever the host has directed it.
+     *
+     * Each call is a no-op if no console object, or no such method on it, is present.
```
bghgary: `e0ed6a1` widened the guarantee — a throwing `console` accessor and a throwing method are both swallowed, and no exception is left pending on `env` — but this line still describes only the missing-`console` case. A consumer reading the header cannot tell that the throwing case is handled, and this is the public contract of the thing the PR exists to make reachable.

The JSI caveat you found belongs in whatever you decide is the right place for it, since the exception is on the same path a caller would be reasoning about here.

```suggestion
     * Each call is a no-op if no console object, or no such method on it, is present.
     * A `console` accessor or method that throws is also swallowed, and no exception is
     * left pending on `env`.
```

Core/JsRuntime/Source/JsConsoleLogger.cpp
```diff
@@ -0,0 +1,56 @@
+#include <Babylon/JsConsoleLogger.h>
+
+namespace Babylon
+{
+    namespace
+    {
+        void LogMethod(Napi::Env env, const char* methodName, const char* message)
+        {
+            try
```
bghgary: The clear below takes any pending exception, including one the caller set before calling in. `JsRuntime::Dispatch` depends on that surviving — it rethrows it at `Core/JsRuntime/Source/JsRuntime.cpp:56-58` to bubble the error to the dispatcher — so a caller that does `ThrowAsJavaScriptException()` and then logs a diagnostic loses the error silently.

Measured on Win32/Chakra with a pending exception set before `LogError`: `pendingAfter=no; logged=0`. The message does not get through either, since the first N-API call fails against the pending exception — so the caller loses its error and gets no log. On the pre-`e0ed6a1` code the exception escaped instead, so this arrives with the fix rather than from BabylonNative.

Returning early costs nothing, because the message is already dropped in this state, and `ThrowingConsoleLeavesNoPendingException` still passes with it applied.

```suggestion
        void LogMethod(Napi::Env env, const char* methodName, const char* message)
        {
            // An exception already pending on entry is the caller's, and every N-API call
            // below would fail against it anyway. Leave it for them to propagate.
            if (env.IsExceptionPending())
            {
                return;
            }

            try
```

Core/JsRuntime/Include/Babylon/JsConsoleLogger.h
```diff
@@ -0,0 +1,25 @@
...
+     */
+    struct JsConsoleLogger final
+    {
+        JsConsoleLogger() = delete;
+
+        static void LogInfo(Napi::Env env, const char* message);
+        static void LogWarn(Napi::Env env, const char* message);
+        static void LogError(Napi::Env env, const char* message);
```
bghgary: These cross a library boundary without `BABYLON_API`, which is `__cdecl` on Windows, so a consumer built with a non-default calling convention gets a mismatch. `JsRuntime.h` in this same directory and target applies it to every entry point, including the `static` member `NativeObject::GetFromJavaScript`. This was a private plugin header in BabylonNative, where it never had to.

Needs `#include <Babylon/Api.h>` alongside `<napi/env.h>`.

```suggestion
        static void BABYLON_API LogInfo(Napi::Env env, const char* message);
        static void BABYLON_API LogWarn(Napi::Env env, const char* message);
        static void BABYLON_API LogError(Napi::Env env, const char* message);
```

CedricGuillemet:
Some code like 3rd party libraries will log to stdout/stderr and we won't be able to see that. So instead of having our own log function, I think it's better to route stdout/err to platform specific output like this : https://github.com/BabylonJS/AndroidExtensions/blob/main/Source/StdoutLogger.cpp

bkaradzic-microsoft:
Agreed. Redirecting stdout/stderr at the platform host is broader and also catches third-party native output; this helper only helps call sites we explicitly migrate.

JsRuntimeHost already uses `AndroidExtensions::StdoutLogger` in its Android test host, so duplicating that behavior in `Core/JsRuntime` is the wrong layer. I'll close this PR. The BabylonNative follow-up should keep native diagnostics on stderr and enable stream forwarding at the application boundary: reuse `AndroidExtensions` on Android and add the equivalent platform sink on Apple.

bkaradzic-microsoft:
Reworked this around Cedric's stream-forwarding suggestion and replaced the old `JsConsoleLogger` design in #233.

This is not Android/Apple-only: Windows preserves the existing destination and mirrors to `OutputDebugString`; Android mirrors to logcat; Apple platforms mirror to unified logging; Linux and other Unix hosts intentionally keep their already-routable standard streams unchanged.

The router now lives in `Core/Foundation`, is opt-in at the application boundary, and captures third-party `printf`/`fprintf` output in addition to BabylonNative diagnostics.

GitHub would not allow this PR to be reopened after its branch was replaced, so the cross-platform implementation is now in #233.

PR #229, opened 2026-08-21
Implement napi type tags and reject unwrapping objects that were never wrapped

### LINKED ISSUE #226: napi_unwrap does not reject objects that were never wrapped (V8 port faults, QuickJS port confuses types)
### Summary

`napi_unwrap` must fail when it is handed an object that was never wrapped. The V8 and QuickJS ports both violate that, so `Napi::ObjectWrap<T>::Unwrap` cannot safely be called on an object whose type has not already been established. On V8 it is an outright memory-safety hole.

This affects every `Unwrap` call site in a consumer, not one polyfill. I hit it in BabylonNative's Canvas polyfill (BabylonJS/BabylonNative#1844), where `ctx.fill(Object.create(Path2D.prototype))` was an access violation.

### V8 port — dereferences internal field 0 unchecked

`Core/Node-API/Source/js_native_api_v8.cc`, `inline napi_status Unwrap(...)` (~line 345). A `[BABYLON-NATIVE-ADDITION]` marked *"Increase perf by using internal field instead of private property"* replaced the private-property lookup, **including its validity check**:

```cpp
// upstream
auto val = obj->GetPrivate(context, NAPI_PRIVATE_KEY(context, wrapper)).ToLocalChecked();
RETURN_STATUS_IF_FALSE(env, val->IsExternal(), napi_invalid_arg);
```

with a bare `obj->GetAlignedPointerFromInternalField(0)`, whose result is then dereferenced (`reference->Data()`).

For any object that is not a wrapped instance, internal field 0 is not a `Reference*`. The read returns garbage and the dereference faults. Reproduction:

```js
const impostor = Object.create(Path2D.prototype);
ctx.fill(impostor);   // 0xC0000005
```

I confirmed this in a local V8 build: deterministic access violation, and it disappears when the unwrap is replaced with a checked lookup.

### QuickJS port — walks the prototype chain

`Core/Node-API/Source/js_native_api_quickjs.cc`, `napi_unwrap` (~line 2495). After the fast path on `js_wrap_class_id` there is a *"Fallback: search the prototype chain for a legacy wrapper object"*.

That returns **some other object's** native pointer — a type confusion rather than a crash. An object created with `Object.create(RealType.prototype)` unwraps to whatever instance is reachable on the chain. It does at least return `napi_invalid_arg` when nothing is found.

### JSI port

No C API, and `ObjectWrap<T>::Unwrap` returns `nullptr` for a non-wrapped object (`napi-inl.h:2268`) rather than throwing. Safe, but inconsistent with the other two.

### Why this is hard to work around downstream

While fixing the Canvas case I found no portable way to do a type check:

- `napi_type_tag_object` / `napi_check_object_type_tag` exist only in `js_native_api_v8.cc`.
- `Napi::Object::DefineProperty` / `PropertyDescriptor` are absent from the JSI port, so a non-enumerable brand cannot be installed.
- `GetInstanceData` / `SetInstanceData` / `AddCleanupHook` are absent from the JSI port, so there is nowhere to keep per-`Env` C++ state.
- `Napi::ObjectWrap<T>::Value()` **throws** on the QuickJS port, which rules out an identity check against a candidate instance.

I ended up branding each instance with a `Napi::External<T>` and validating the pointer against a registry of live addresses. That works, but every consumer having to invent this is a strong argument for fixing the ports.

### Suggested fix

Restore the validity check in the V8 port (keep the internal-field fast path, but verify the field actually holds the wrapper before dereferencing), and drop the prototype-chain fallback in the QuickJS port so a non-wrapped object returns `napi_invalid_arg`.

Related: #225 (the `napi_throw` family reports failure on success), found while chasing the same PR.

bghgary:

## Context

`napi_unwrap` returns `void*` and does no type check, so it cannot answer "is this JS object one of my type". BabylonNative#1844 hand-rolls a ~100-line registry to get that answer, because the Node-API primitive for it — `napi_type_tag_object` / `napi_check_object_type_tag` — is unreachable here: `NAPI_VERSION` is pinned to 5 and the whole type-tag surface sits behind `NAPI_VERSION >= 8`. Even the V8 body was dead code, and would not have compiled, since the `NAPI_PRIVATE_KEY` it calls is commented out.

## Worth a look

- **`NAPI_VERSION` stays at 5; only the type-tag surface is ungated.** `napi_get_version` returns that macro, so raising it to 8 would have all four ports claim v8 while Chakra and JavaScriptCore implement none of v6/v7 (BigInt, `napi_get_all_property_names`, instance data, ArrayBuffer detaching). `env_hermes.cc` also depends on the 5.
- **QuickJS, Chakra and JavaScriptCore keep tags in a `WeakMap` held only on `napi_env__`.** They have no per-object native slot for an arbitrary object, and a hidden own property is not private — `Object.getOwnPropertySymbols` hands script the key even under a symbol, and the tag can then be read off a real instance and replayed onto a spoofed one. V8 keeps using `v8::Private`.
- **QuickJS `napi_wrap` now rejects an object that no napi class constructor created.** Wrapping one used to splice a wrapper into its prototype chain, which is what unwrap's chain walk existed to find. That is the same constraint V8's internal field already imposes.
- **JavaScriptCore had the same prototype-chain defect** as QuickJS, one layer down in `NativeInfo::Query`, and it is not in #226. It applies to references too: two distinct objects reported the same object id.

## Verification

Built and ran the full suite on all six engine configurations: V8, Chakra, QuickJS, JSI and Hermes on Windows x64, and JavaScriptCore on Ubuntu 24.04. The new `NodeApi.TypeTags` test was also run against each guard reverted in turn — V8 dies with an access violation, QuickJS and JavaScriptCore hand back the wrong object's pointer.

Fixes #226.

### DIFF
--- a/Core/Node-API/Include/Shared/napi/js_native_api.h
+++ b/Core/Node-API/Include/Shared/napi/js_native_api.h
@@ -613,8 +613,9 @@ NAPI_EXTERN napi_status NAPI_CDECL
 napi_is_detached_arraybuffer(napi_env env, napi_value value, bool* result);
 #endif  // NAPI_VERSION >= 7
 
-#if NAPI_VERSION >= 8
 // Type tagging
+// [BABYLON-NATIVE-ADDITION]: exposed unconditionally, see napi_type_tag in
+// js_native_api_types.h.
 NAPI_EXTERN napi_status NAPI_CDECL napi_type_tag_object(
     napi_env env, napi_value value, const napi_type_tag* type_tag);
 
@@ -623,6 +624,8 @@ napi_check_object_type_tag(napi_env env,
                            napi_value value,
                            const napi_type_tag* type_tag,
                            bool* result);
+
+#if NAPI_VERSION >= 8
 NAPI_EXTERN napi_status NAPI_CDECL napi_object_freeze(napi_env env,
                                                       napi_value object);
 NAPI_EXTERN napi_status NAPI_CDECL napi_object_seal(napi_env env,
--- a/Core/Node-API/Include/Shared/napi/js_native_api_types.h
+++ b/Core/Node-API/Include/Shared/napi/js_native_api_types.h
@@ -163,11 +163,13 @@ typedef enum {
 } napi_key_conversion;
 #endif // NAPI_VERSION >= 6
 
-#if NAPI_VERSION >= 8
+// [BABYLON-NATIVE-ADDITION]: type tags are implemented by every engine here, so
+// they are exposed unconditionally rather than at NAPI_VERSION >= 8. The tree
+// stays at NAPI_VERSION 5 because napi_get_version reports that macro and the
+// rest of v6/v7/v8 is not implemented on all engines.
 typedef struct {
   uint64_t lower;
   uint64_t upper;
 } napi_type_tag;
-#endif // NAPI_VERSION >= 8
 
 #endif  // SRC_JS_NATIVE_API_TYPES_H_
--- a/Core/Node-API/Include/Shared/napi/napi-inl.h
+++ b/Core/Node-API/Include/Shared/napi/napi-inl.h
@@ -1343,8 +1343,8 @@ inline TypeTaggable::TypeTaggable() : Value() {}
 inline TypeTaggable::TypeTaggable(napi_env _env, napi_value _value)
     : Value(_env, _value) {}
 
-#if NAPI_VERSION >= 8
-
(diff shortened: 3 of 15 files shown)


matthargett:
Please can you look at #189? after merging it we can it go onto proper NAPI v8 support instead of a non-standard offshoot that will keep BabylonNative users locked out of the modern NodeJS addon ecosystem.

PR #232, opened 2026-08-26
Pump V8's foreground task queue so async WebAssembly settles

bkaradzic-microsoft:
V8 finishes asynchronous WebAssembly compilation on a background thread and then posts a **foreground** task to settle the promise on the isolate thread. `AppRuntime_JSI` already ran those via `TaskRunnerAdapter`; the direct-V8 path dropped them, so asynchronous WebAssembly promises could hang. Synchronous `new WebAssembly.Module` was unaffected.

Foreground posts now go through `V8ForegroundTaskRunner`, which dispatches `v8::Task::Run` through AppRuntime. Delayed posts use `DelayedTaskScheduler`, extracted into Foundation from the timer queue behind `setTimeout` / `setInterval`. AppRuntime registers its scheduler with the environment for the polyfill to borrow; standalone `JsRuntime` instances retain their original API and do not acquire timer ownership. `TimeoutDispatcher` owns a fallback scheduler when no host scheduler is registered. Interval rearming remains on the JavaScript thread after the callback.

Delayed V8 posts round their absolute requested time upward to microsecond precision, and completed pending records are removed even when completion precedes `Schedule` returning. Shared callback state and explicit host unregistration guard teardown independently of retained task-runner references. `V8Platform` and the foreground runner are in dedicated files; the wake-and-pump bridge and `PumpMessageLoop` are gone.

Added asynchronous WebAssembly promise-settlement tests plus native timing, pending-record, cancellation, scheduler-ownership, and shutdown regressions.

Validation for 7192ac0 (Windows Release): V8 11.9 passed 17 selected native tests, including the JavaScript suite with 228 passing; Chakra passed 9 selected native tests, including 225 JavaScript tests with the 3 V8-only tests skipped. The 16 native timer/lifetime tests also passed 25 repeated iterations. The V8 implementation files syntax-compile against the actual Android V8 11.0 headers; Android device execution remains for CI.


Core/AppRuntime/Source/AppRuntime_V8.cpp
```diff
@@ -72,6 +245,20 @@ namespace Babylon
...
+        // pending wake-up; it is cleared once that wake-up has been serviced.
+        auto wakePending = std::make_shared<std::atomic_bool>(false);
+        Module::Instance().Platform().SetWake(isolate, [this, wakePending]() {
+            if (wakePending->exchange(true))
+            {
+                return;
+            }
+
+            Dispatch([wakePending](Napi::Env) { wakePending->store(false); });
+        });
```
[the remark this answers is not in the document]
bkaradzic-microsoft: Good catch on the ordering — the flag really is cleared before `Dispatch` pumps. But I think the suggested fix inverts a deliberate invariant, so I've kept the ordering and documented it instead (b7fc69e).

Clearing *after* pumping loses wake-ups:

1. `PumpMessageLoop` drains the queue and returns false.
2. A background thread posts a foreground task `T`. `Wake` sees `wakePending == true`, so it returns without dispatching.
3. The callback stores `false`.
4. `T` is sitting in the queue with nothing scheduled to pump it.

That is exactly the hang this PR fixes, just narrowed to a race window instead of happening every time.

Clearing *before* pumping is the standard edge-triggered pattern and cannot lose a task. A post either happens before the clear — in which case the pump that immediately follows drains it — or after the clear, in which case it schedules a fresh wake.

The redundant churn is real but bounded: the `exchange` still collapses a burst, so the worst case is one no-op dispatch plus one empty pump per drain cycle. That is a cheap price for not being able to drop a task. If it ever showed up in a profile, the safe fix would be a generation counter (re-check for new posts after pumping and re-arm), not swapping the order.

bkaradzic-microsoft:
CI found two things my local V8 build could not, both fixed in fda5f78.

**1. Android build break.** Android builds against V8 **11.0**, desktop against **11.9**. `DispatchingPlatform` forwarded the entire `v8::Platform` interface, and five of those members simply do not exist in 11.0: `GetThreadIsolatedAllocator`, `CreateBlockingScope`, `CurrentClockTimeMilliseconds`, `CurrentClockTimeMillisecondsHighResolution`, and the `TaskPriority` overload of `GetForegroundTaskRunner`. Overriding a method the base class does not declare is a hard error, hence `no type named 'ThreadIsolatedAllocator' in namespace 'v8'` and friends.

Rather than guess version thresholds I pulled the Android package and read its `v8-version.h` (11.0.226.16) and `v8-platform.h` directly, so the gating matches the header rather than my memory of when each member landed. I also checked all nine of 11.0's pure virtuals are still overridden -- that is the part that would silently break the build again. Where a member is gated out we fall back to `v8::Platform`'s own default, which is benign for each: a null allocator, a null blocking scope, and clock values derived from `CurrentClockTimeMillis`.

**2. The WebAssembly tests ran on every engine.** The fix is V8-only, so JSC/Chakra/Hermes/QuickJS did not fail -- they *hung*, three times 30s apiece. Those engines' runtimes have the same class of gap; fixing them is a separate change.

I added a `hostEngine` global mirroring the existing `hostPlatform` and scoped the suite with a `beforeEach` skip. Verified both directions rather than just the happy path:

| engine | result |
|---|---|
| V8 | 220 passing, tests execute |
| Chakra | 217 passing, **3 pending**, 11.7s (was ~90s of timeouts) |

Worth noting the earlier green runs on this PR were queue artifacts -- this was the first run that actually built Android, so the breakage was there from the start rather than introduced by the last push.

bkaradzic-microsoft:
Follow-up: that push missed a second definition site, now folded in (the commit above is amended to `4a5d722`).

Android does not build the `UnitTests` target -- it builds a separate `UnitTestsJNI` library from its own `CMakeLists.txt`, which compiles the same `Shared.cpp`. So adding `JSRUNTIMEHOST_ENGINE` only to `UnitTests` left the JNI target without it: `error: use of undeclared identifier 'JSRUNTIMEHOST_ENGINE'`, which broke all three Android legs including JSC and QuickJS.

One thing worth flagging while I was in there: `JSRUNTIMEHOST_PLATFORM` is passed to the JNI target as `-DJSRUNTIMEHOST_PLATFORM=\"\"` -- empty. It is set as an ordinary variable inside the root project's scope, and `add_subdirectory` does not propagate that back up to the JNI `CMakeLists.txt`, so `hostPlatform` is `""` on Android today. `NAPI_JAVASCRIPT_ENGINE` does not have this problem because gradle passes it on the CMake command line, making it a cache variable visible in every scope, so `hostEngine` resolves correctly. I have left the pre-existing `hostPlatform` quirk alone as it is unrelated to this PR, but it is a live bug for any test that gates on `hostPlatform` on Android.

Core/AppRuntime/Source/AppRuntime_V8.cpp
```diff
@@ -2,25 +2,211 @@
...
+            {
+                std::function<void()> wake;
+                {
+                    std::scoped_lock lock{m_mutex};
+                    const auto entry = m_wakes.find(isolate);
+                    if (entry == m_wakes.end())
+                    {
+                        return;
+                    }
+                    wake = entry->second;
```
bghgary: Make unregistering synchronize with a wake already copied here. `Wake` releases the mutex before invoking a callback that captures `AppRuntime*`, so `SetWake(isolate, nullptr)` can erase it, let the runtime thread exit and `m_impl` be destroyed, then this posting thread resumes and calls `Dispatch` through the dead object. V8 permits `TaskRunner` posts after isolate destruction.

Core/AppRuntime/Source/AppRuntime_V8.cpp
```diff
@@ -2,25 +2,211 @@
...
+            void PostDelayedTask(std::unique_ptr<v8::Task> task, double delayInSeconds) override
+            {
+                m_inner->PostDelayedTask(std::move(task), delayInSeconds);
+                m_platform.Wake(m_isolate);
+            }
+
+            void PostNonNestableDelayedTask(std::unique_ptr<v8::Task> task, double delayInSeconds) override
+            {
+                m_inner->PostNonNestableDelayedTask(std::move(task), delayInSeconds);
+                m_platform.Wake(m_isolate);
```
bghgary: Both delayed paths need a wake at the deadline, not only when the task is posted. The immediate no-op pumps before the deadline and clears `wakePending`; with no render loop, nothing pumps when due. V8 11.9 uses delayed foreground tasks for its memory reducer and metrics, so those tasks remain stalled in the idle case this wrapper is meant to fix.

Core/AppRuntime/Source/AppRuntime_V8.cpp
```diff
@@ -108,14 +327,33 @@ namespace Babylon
         }
 
         // Destroy the isolate.
+        Module::Instance().Platform().SetWake(isolate, nullptr);
```
bghgary: Call `v8::platform::NotifyIsolateShutdown(&Module::Instance().DefaultPlatform(), isolate)` before `isolate->Dispose()`. Libplatform requires this whenever its process-global platform outlives an isolate; without it, the inner platform retains this raw isolate key and its queued tasks. A later `AppRuntime` whose isolate reuses that address can then pump foreground work from the destroyed runtime.

Tests/UnitTests/CMakeLists.txt
```diff
@@ -46,6 +46,7 @@ endif()
 
 add_executable(UnitTests ${SOURCES} ${SCRIPTS} ${TYPE_SCRIPTS} ${ASSETS})
 target_compile_definitions(UnitTests PRIVATE JSRUNTIMEHOST_PLATFORM="${JSRUNTIMEHOST_PLATFORM}")
+target_compile_definitions(UnitTests PRIVATE JSRUNTIMEHOST_ENGINE="${NAPI_JAVASCRIPT_ENGINE}")
```
bghgary: Please use `NAPI_JAVASCRIPT_ENGINE` as the compiler definition name here and in the Android test target, and read that same name in `Shared.cpp`. `JSRUNTIMEHOST_ENGINE` only aliases the existing CMake value; keeping one name avoids an unnecessary translation and matches the existing `JSRUNTIMEHOST_PLATFORM` pattern.

Core/AppRuntime/Source/AppRuntime_V8.cpp
```diff
@@ -2,25 +2,211 @@
...
+        private:
+            std::shared_ptr<v8::TaskRunner> WrapForegroundTaskRunner(v8::Isolate* isolate);
+
+            std::unique_ptr<v8::Platform> m_inner;
+            std::mutex m_mutex;
+            std::map<v8::Isolate*, std::function<void()>> m_wakes;
+            std::map<v8::Isolate*, std::shared_ptr<v8::TaskRunner>> m_taskRunners;
+        };
+
+        class WakingTaskRunner final : public v8::TaskRunner
```
bghgary: Please replace the wake-and-pump bridge with a foreground task runner that dispatches the V8 task itself through AppRuntime. Chromium's `V8ForegroundTaskRunner` uses this model by posting `v8::Task::Run` directly to its host task runner. Immediate tasks can use the existing dispatcher; delayed tasks should reuse the timer queue behind `setTimeout`, after extracting its native scheduling core, so each task is dispatched when its deadline arrives. This removes the second foreground queue, the no-op wake, and `PumpMessageLoop`.

Tests/UnitTests/CMakeLists.txt
bghgary: Please use `NAPI_JAVASCRIPT_ENGINE` as the compiler definition name here and in the Android test target, and read that same name in `Shared.cpp`. `JSRUNTIMEHOST_ENGINE` only aliases the existing CMake value; keeping one name avoids an unnecessary translation and matches the existing `JSRUNTIMEHOST_PLATFORM` pattern.

bghgary (changes requested):

Concerns inline.

Core/AppRuntime/Source/AppRuntime_V8.cpp
bkaradzic-microsoft: Done in 7bab0cd. `DispatchingTaskRunner` posts `v8::Task::Run` through `AppRuntime::Dispatch`, matching the Chromium foreground-runner model.

Delayed posts go through `DeadlineScheduler`, extracted from `TimeoutDispatcher`, so setTimeout/setInterval and V8 delayed tasks share one native queue. `PumpMessageLoop` and the no-op wake are gone. Interval re-arm still happens on the JS thread after the callback so the setInterval starvation fix from #220 is preserved.

Tests/UnitTests/CMakeLists.txt
bkaradzic-microsoft: Done in 7bab0cd. The UnitTests and Android JNI targets now define `NAPI_JAVASCRIPT_ENGINE="${NAPI_JAVASCRIPT_ENGINE}"`, and `Shared.cpp` reads that same name.

Core/AppRuntime/Source/AppRuntime_V8.cpp
```diff
@@ -1,26 +1,277 @@
...
+            void PostDelayed(std::unique_ptr<v8::Task> task, double delayInSeconds)
+            {
+                auto host = m_platform.GetHost(m_isolate);
+                if (host.runtime == nullptr || host.scheduler == nullptr)
+                {
+                    return;
+                }
+
+                auto shared = std::shared_ptr<v8::Task>(std::move(task));
+                auto delay = std::chrono::duration_cast<std::chrono::milliseconds>(std::chrono::duration<double>(delayInSeconds));
```
bghgary: Please compute the absolute deadline from `steady_clock::now() + delayInSeconds`, then round that deadline upward to `DeadlineScheduler::TimePoint`'s microsecond precision. The current millisecond truncation can run delayed V8 tasks before their requested deadline.

Core/AppRuntime/Source/AppRuntime_V8.cpp
```diff
@@ -1,26 +1,277 @@
...
+                }
+
+                std::scoped_lock lock{m_mutex};
+                const auto id = host.scheduler->Schedule(delay, [runtime = host.runtime, isolate = m_isolate, shared]() {
+                    runtime->Dispatch([shared, isolate](Napi::Env) {
+                        v8::Isolate::Scope isolate_scope{isolate};
+                        shared->Run();
+                    });
+                });
+                m_pending.push_back(Pending{host.scheduler, id});
```
bghgary: Please remove each `Pending` entry when its delayed callback fires without making the callback depend on the task runner's lifetime, including when the callback fires before `Schedule` returns and its ID is recorded. The vector currently retains one stale scheduler ID for every completed V8 delayed task until destruction, so it grows without bound.

Core/JsRuntime/Include/Babylon/DeadlineScheduler.h
```diff
@@ -0,0 +1,34 @@
...
+#include <cstdint>
+#include <functional>
+#include <memory>
+
+namespace Babylon
+{
+    /// Native deadline queue used by setTimeout/setInterval and by the V8
+    /// foreground task runner's delayed posts. Callbacks run on the scheduler
+    /// thread; callers that need the JavaScript thread Dispatch themselves.
+    class DeadlineScheduler final
```
bghgary: Would `DelayedTaskScheduler` or `TimerScheduler` be clearer here? A deadline usually means work must finish by a time, while this class schedules callbacks not to run before a time, so `DeadlineScheduler` suggests a different contract.

Core/JsRuntime/Include/Babylon/JsRuntime.h
```diff
@@ -37,17 +40,22 @@ namespace Babylon
         // later -- an instance of an inheriting type, for example. The dispatch function
         // must be safely callable as soon as it is passed to the JsRuntime constructor.
         static JsRuntime& BABYLON_API CreateForJavaScript(Napi::Env, DispatchFunctionT);
+        static JsRuntime& BABYLON_API CreateForJavaScript(Napi::Env, DispatchFunctionT, DeadlineScheduler&);
```
bghgary: Could we keep delayed scheduling out of `JsRuntime`? The two-argument overload now creates a scheduler thread for every standalone runtime even when the host never initializes the scheduling polyfill. AppRuntime's shared scheduler could be associated with the environment for `Scheduling::Initialize(env)` to borrow, while `TimeoutDispatcher` continues owning its scheduler when none is registered. This would preserve the existing initialization API without adding timer ownership to the core runtime.

Core/AppRuntime/Source/AppRuntime_V8.cpp
```diff
@@ -1,26 +1,277 @@
...
 {
     namespace
     {
+        // V8 hands work that finishes off-thread - most visibly asynchronous WebAssembly
+        // compilation - back to the isolate by posting a task to the platform's *foreground*
+        // task runner. Chromium runs those as `v8::Task::Run` on the host task runner. The
+        // host JavaScript thread is AppRuntime's dispatcher, so this platform's foreground
+        // runner posts the task itself instead of parking it on libplatform's queue and
+        // pumping later. Delayed posts reuse the DeadlineScheduler behind setTimeout.
+        class DispatchingPlatform final : public v8::Platform
```
bghgary: Could we move the V8 platform adapter and foreground task runner into dedicated files, with the version-specific forwarding separated from AppRuntime task routing? Keeping `AppRuntime_V8.cpp` focused on isolate creation, host registration, execution, unregistration, and disposal would make the lifetime behavior easier to verify. `SetHost(..., nullptr, nullptr)` could also become explicit registration and unregistration operations.

bghgary (changes requested):

Two correctness concerns and three design suggestions inline.

Core/AppRuntime/Source/AppRuntime_V8.cpp
bkaradzic-microsoft: Fixed in 7192ac0. V8 delayed posts now compute the absolute time from steady_clock::now() plus the requested delay, round that time point upward with chrono::ceil<microseconds>, and call Schedule(TimePoint, ...). The millisecond conversion is gone.

Added deterministic sub-microsecond/fractional-millisecond rounding checks and a real-scheduler regression asserting dispatch is not earlier than the requested time. Nonpositive delays remain immediately eligible.

Core/AppRuntime/Source/AppRuntime_V8.cpp
bkaradzic-microsoft: Fixed in 7192ac0. Pending records now live in shared callback state, not on the runner itself. Completion removes its record; if it occurs before Schedule returns, the completed record is not inserted afterward. Callbacks never capture the runner's this pointer.

Shutdown gates dispatch and cancels only outstanding records, including when V8 retains the runner. Regressions cover 1,000 completed posts without stale cancellation IDs, synchronous completion before Schedule returns, an extracted callback invoked after runner destruction, and shutdown racing with dispatch.

Core/JsRuntime/Include/Babylon/DeadlineScheduler.h
bkaradzic-microsoft: Renamed to DelayedTaskScheduler in 7192ac0 and moved the native scheduler into Foundation. The header explicitly documents that callbacks run on the scheduler thread and callers dispatch to the JavaScript thread themselves. All includes and build references use the new name.

Core/JsRuntime/Include/Babylon/JsRuntime.h
bkaradzic-microsoft: Implemented in 7192ac0. JsRuntime is back to its original two-argument API with no scheduler ownership, accessor, or extra thread. AppRuntime owns the scheduler and associates it with the environment through a separate registration helper. Scheduling::Initialize(env) is unchanged; TimeoutDispatcher borrows the registered scheduler or owns a fallback when none is registered.

The shared host scheduler is stopped and joined before environment teardown, and timeout callbacks use shared lifetime state instead of capturing the dispatcher. Added registration, fallback, cancellation, and shutdown regressions; the scheduling tests and JavaScript suites pass on both V8 and Chakra.

Core/AppRuntime/Source/AppRuntime_V8.cpp
bkaradzic-microsoft: Implemented in 7192ac0. V8Platform.{h,cpp} contains the platform adapter/version forwarding and host registration; V8ForegroundTaskRunner.{h,cpp} contains immediate/delayed routing and pending-task state. These sources are built only for direct V8.

AppRuntime_V8.cpp now makes the lifecycle explicit: Allocate, RegisterHost, Initialize, Run with an UnregisterHost shutdown hook, detach, dispose. Registration precedes initialization, eliminating the thread-local construction fallback. Unregistration deactivates retained runners before scheduler shutdown and dispatcher clearing, so queued tasks are released before isolate disposal.

Built and ran against desktop V8 11.9; also syntax-compiled the three V8 implementation files against the actual Android V8 11.0 headers. Android device execution remains for CI.

bkaradzic-microsoft:
CI follow-up fixed in 3014935: the extracted V8 headers included v8-platform.h directly, bypassing the existing napi/env.h warning guards. Android then treated V8 11.0's unused parameters as errors. Both headers now use the existing wrapper; project warning settings are unchanged.

Reproduced the failure with Android NDK Clang and verified all three V8 implementation files compile with -Werror after the fix. Desktop V8's eight runner regressions and the JavaScript suite also pass. The five implemented review threads are resolved, and @bghgary's review has been re-requested. Monitoring the new CI run.

Core/Foundation/Source/DelayedTaskScheduler.cpp
```diff
@@ -0,0 +1,217 @@
...
+
+        Id NextId()
+        {
+            while (true)
+            {
+                ++m_lastId;
+                if (m_lastId <= 0)
+                {
+                    m_lastId = 1;
+                }
```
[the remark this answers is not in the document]
bkaradzic-microsoft: Fixed in 5da35cc. The native scheduler now uses a shared IncrementTimerId helper that checks INT32_MAX before adding and wraps directly to 1; it still skips IDs that are currently in use.

Added boundary coverage for 0, INT32_MAX - 1, INT32_MAX, and the next post-wrap ID. The maximum-value case also has a static_assert, so signed overflow would fail constant-expression evaluation rather than being hidden by machine wraparound.

Polyfills/Scheduling/Source/TimeoutDispatcher.cpp
```diff
@@ -39,169 +30,87 @@ namespace Babylon::Polyfills::Internal
...
-
-            while (!m_timeMap.empty())
-            {
-                nextTimePoint = m_timeMap.begin()->second->time;
-                if (nextTimePoint <= Now())
+                ++lastTimeoutId;
+                if (lastTimeoutId <= 0)
                 {
-                    break;
+                    lastTimeoutId = 1;
```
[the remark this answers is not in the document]
bkaradzic-microsoft: Fixed in 5da35cc. TimeoutDispatcher now uses the same max-before-increment helper as the native scheduler, so it never increments a signed value past INT32_MAX. The existing live-timer collision check is unchanged.

The shared helper has runtime and compile-time boundary regressions; the timer, foreground-runner, and JavaScript regression suites pass.

Core/AppRuntime/Source/AppRuntime_V8.cpp
```diff
@@ -95,7 +100,9 @@ namespace Babylon
             }
 #endif
 
-            Run(env);
+            Run(env, [&platform, isolate] {
```
bghgary: Could `Run` call an engine-selected shutdown method, following the `RunEnvironmentTier` and `DrainMicrotasks` pattern, instead of accepting a one-off shutdown callback? The V8 implementation could unregister foreground-task routing while the other engines remain no-ops. This would keep teardown consistent with the existing engine boundary and make the shutdown order easier to follow.

Core/Foundation/Source/DelayedTaskScheduler.h
```diff
@@ -0,0 +1,40 @@
...
+#include <cstdint>
+#include <functional>
+#include <memory>
+
+namespace Babylon
+{
+    /// Native delayed-work queue used by setTimeout/setInterval and by host
+    /// task runners. Callbacks run on the scheduler thread; callers that need
+    /// the JavaScript thread must dispatch there themselves.
+    class DelayedTaskScheduler final
```
bghgary, line 15: Could this remain one internal Foundation abstraction instead of adding two public-facing classes? `DelayedTaskScheduler` and `DelayedTaskSchedulerRegistration` are exposed through public include paths, although only AppRuntime, Scheduling, and their tests consume them. Registration and lookup could live with the scheduler in Foundation and use a dedicated Babylon global property instead of depending on JsRuntime's `_native` object. A `FoundationInternal` interface target could expose the internal header only to those consumers.

Core/Foundation/Include/Babylon/Internal/TimerId.h
```diff
@@ -0,0 +1,12 @@
+#pragma once
+
+#include <cstdint>
+#include <limits>
+
+namespace Babylon::Internal
+{
+    constexpr int32_t IncrementTimerId(int32_t id)
```
bghgary: Could we keep the wrap logic local to each ID allocator instead of introducing `TimerId.h`? The scheduler task ID and JavaScript timeout ID have different meanings and lifetimes; they only share a small max-before-increment check. Inlining that check in `NextId` and `NextTimeoutId` would be clearer than coupling the two implementations through a shared helper whose test does not exercise either allocator.

bghgary (changes requested):

Three design concerns inline.

Core/AppRuntime/Source/AppRuntime_V8.cpp
bkaradzic-microsoft: Implemented and pushed in 864fb3babb1b62c6e6b619036a053d9071ab7154. `Run` now calls the engine-selected `ShutdownEnvironment(Napi::Env)` hook instead of accepting a shutdown callback. V8 unregisters foreground-task routing there; Chakra, Hermes, JSI, JavaScriptCore, and QuickJS implement no-ops.

The shutdown order remains explicit in `Run`: stop engine routing, unregister the scheduler, join it, then clear queued dispatches, all before environment detach/isolate disposal.

Core/Foundation/Source/DelayedTaskScheduler.h
bkaradzic-microsoft: Implemented and pushed in 864fb3babb1b62c6e6b619036a053d9071ab7154. Scheduling and environment registration/lookup now live together in `Babylon::Internal::DelayedTaskScheduler`, with its header under `Core/Foundation/Source`. The public scheduler/registration headers and separate JsRuntime registration implementation are removed.

`FoundationInternal` exposes the internal header to AppRuntime, Scheduling, and their tests without making it part of Foundation's public include path. The non-owning environment association uses the dedicated `_BabylonDelayedTaskScheduler` global property, not JsRuntime's `_native` object. The registration regression now exercises lookup, unregister, and re-register with `_native` absent.

Core/Foundation/Include/Babylon/Internal/TimerId.h
bkaradzic-microsoft: Implemented and pushed in 864fb3babb1b62c6e6b619036a053d9071ab7154. Removed `TimerId.h`; `NextId` and `NextTimeoutId` now each perform their own max-before-increment check while retaining their existing live-ID collision checks.

Replaced the helper-only tests with tests that allocate through the real `Schedule` and `TimeoutDispatcher::Dispatch` paths across `INT32_MAX - 1`, `INT32_MAX`, `1`, and `2`. Private test access only seeds each allocator's counter; the wrapping and allocation under test are the production implementations.

Core/Foundation/Source/DelayedTaskScheduler.h
```diff
@@ -0,0 +1,51 @@
...
+
+        DelayedTaskScheduler();
+        ~DelayedTaskScheduler();
+
+        DelayedTaskScheduler(const DelayedTaskScheduler&) = delete;
+        DelayedTaskScheduler& operator=(const DelayedTaskScheduler&) = delete;
+
+        // Environment association is non-owning and accessed on the JavaScript
+        // thread. The registered scheduler must outlive its borrowers.
+        void Register(Napi::Env env);
```
bghgary: Could all three environment-association operations be static and use names that reflect their non-owning behavior? `Register` is currently an instance method while `Unregister` and `Get` are static, although none owns the scheduler lifecycle. `SetForJavaScript(env, scheduler)`, `ClearFromJavaScript(env)`, and `GetFromJavaScript(env)` would make the API consistent and distinguish it from `JsRuntime::CreateForJavaScript`, which creates and transfers ownership.

bghgary (approved):

One minor comment inline.

Core/Foundation/Source/DelayedTaskScheduler.h
bkaradzic-microsoft: Implemented and pushed in 5b356fb29c66652a5a41d1d0b6428efd857bbb4f. All three association operations are now static: `SetForJavaScript(env, scheduler)`, `ClearFromJavaScript(env)`, and `GetFromJavaScript(env)`.

Updated AppRuntime, Scheduling, and the association regression to use the new API. The setter only stores a non-owning reference; neither setting nor clearing the association creates, destroys, or transfers ownership of the scheduler.

PR #233, opened 2026-08-27
Forward native standard streams to platform diagnostics

bkaradzic-microsoft:
Captures process-level `stdout` and `stderr` so native and third-party diagnostics reach each platform's normal diagnostic channel without routing through JavaScript.

## What

- Adds an opt-in, N-API-independent `Babylon::StandardStreamLogger` in `Core/Foundation`.
- Windows tees both streams to their original destinations and `OutputDebugString`.
- Android tees both streams to their original destinations and logcat.
- iOS and macOS tee both streams to their original destinations and unified logging.
- Linux and other Unix platforms leave standard streams unchanged because terminals, CI pipes, journald, and containers already consume them.
- Replaces the Android unit-test host's `AndroidExtensions::StdoutLogger` with the shared implementation.

`Start()` and `Stop()` are idempotent. The host opts in before initializing third-party libraries, and `Stop()` restores the original destinations and drains pending output.

## Why

The original `JsConsoleLogger` design only captured migrated call sites, required a live `Napi::Env`, and could not capture libraries writing directly through `printf` or `fprintf`. Redirecting the process streams at the application boundary covers both first-party and third-party native output while preserving existing terminal, file, parent-process, and CI destinations.

## Tests

- Win32/Chakra: 12 C++ tests and 225 JavaScript assertions passing.
- New lifecycle test covers repeated start/stop calls, exact tee behavior, and a final line without a trailing newline.
- The Foundation source also compiles through its non-Windows Unix path.


Tests/UnitTests/Android/app/src/main/cpp/JNI.cpp
```diff
@@ -14,20 +14,24 @@ Java_com_jsruntimehost_unittests_Native_javaScriptTests(JNIEnv* env, jclass claz
...
     Babylon::DebugTrace::SetTraceOutput([](const char* trace) { printf("%s\n", trace); fflush(stdout); });
 
     auto testResult = RunTests();
 
-    android::StdoutLogger::Stop();
+    const bool loggerStopped = Babylon::StandardStreamLogger::Stop();
 
     java::websocket::WebSocketClient::DestructJavaWebSocketClass(env);
-    return testResult;
+    return loggerStopped ? testResult : -1;
```
bghgary: `Stop()` failing discards `testResult` and returns `-1` with nothing logged, so an Android red caused by the 2-second drain timeout would be indistinguishable from a test regression. The `Start()` failure path above logs before it returns `-1`.

Core/Foundation/Source/StandardStreamLogger.cpp
```diff
@@ -0,0 +1,546 @@
...
+        if (::fcntl(fds[0], F_SETFD, FD_CLOEXEC) != 0)
+        {
+            const int error = errno;
+            (void)::close(fds[0]);
+            (void)::close(fds[1]);
+            errno = error;
+            return -1;
+        }
+        return 0;
+    }
```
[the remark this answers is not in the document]
bkaradzic-microsoft: Agreed — fixed on the POSIX path. `CreatePipe` now sets `FD_CLOEXEC` on both ends before returning. The Windows path already creates non-inheritable ends via `CreatePipe` + `bInheritHandle = FALSE`.

Note that after `dup2` into fd 1/2 the process standard streams remain intentionally inheritable; `Stop()` already bounds the drain wait and detaches if a child keeps the redirected writer alive.

Tests/UnitTests/Shared/StandardStreamLogger.cpp
```diff
@@ -0,0 +1,197 @@
...
+
+    const bool started = Babylon::StandardStreamLogger::Start();
+    const bool isStarted = Babylon::StandardStreamLogger::IsStarted();
+    const bool secondStart = Babylon::StandardStreamLogger::Start();
+
+    std::fputs("StandardStreamLogger stdout test", stdout);
+    std::fflush(stdout);
+    std::fputs("StandardStreamLogger stderr test\n", stderr);
+    std::fflush(stderr);
+
```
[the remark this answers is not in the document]
bkaradzic-microsoft: Already addressed in f8b8619. The unasserted stderr write is gone; the test only exercises the stdout tee path it asserts.

We specifically cannot write to stderr in this suite on iOS: the CI harness does `xcrun simctl launch ... 2> /tmp/exitCode` and then `exit $(cat /tmp/exitCode)`, so any app stderr corrupts the numeric exit-code handshake (that is what broke `iOS_Xcode264` on the first CI run).

bkaradzic-microsoft (commented):
Re-requesting Balanced Copilot review after CLOEXEC fix and replies.

Core/Foundation/Source/StandardStreamLogger.cpp
```diff
@@ -0,0 +1,584 @@
...
+                channel.Reader.join();
+            }
+            else
+            {
+                channel.Reader.detach();
+                restored = false;
+            }
+        }
+        channel = {};
+        return restored;
```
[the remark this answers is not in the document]
bkaradzic-microsoft: Agreed. Stop() still detaches after the 2s drain timeout so shutdown cannot hang forever, but the Completion future is now retained in g_outstandingDrains. Start() reaps those futures (waiting up to another 2s) and refuses to install a new tee while a detached drain is still alive, so you cannot get concurrent drains / duplicated platform logging after a timed-out Stop().

Core/Foundation/Source/StandardStreamLogger.cpp
```diff
@@ -0,0 +1,584 @@
...
+                else if (pending.size() >= MAX_PLATFORM_LINE_SIZE)
+                {
+                    EmitLine(stream, pending.substr(0, MAX_PLATFORM_LINE_SIZE));
+                    pending.erase(0, MAX_PLATFORM_LINE_SIZE);
+                }
+                else
+                {
+                    break;
+                }
+            }
```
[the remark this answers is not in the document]
bkaradzic-microsoft: Fixed. Drain() now walks complete lines with a start index and only erases the consumed prefix once per read batch, so we no longer memmove the whole pending buffer on every newline.

Core/Foundation/Source/StandardStreamLogger.cpp
```diff
@@ -0,0 +1,584 @@
...
+    {
+        if (!line.empty() && line.back() == '\r')
+        {
+            line.pop_back();
+        }
+        WritePlatform(stream, line);
+    }
+
+    void Drain(int readFd, int originalFd, Stream stream)
+    {
```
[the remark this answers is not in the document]
bkaradzic-microsoft: Added a comment. 3800 is intentionally under the ~4 KiB practical caps for OutputDebugStringA, Android logcat, and Apple os_log payloads so long unbroken writes are chunked before the platform truncates them.

Tests/UnitTests/Android/app/src/main/cpp/JNI.cpp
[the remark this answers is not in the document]
bkaradzic-microsoft: Done. JNI now writes an ANDROID_LOG_ERROR when Stop() fails (restore or drain timeout) before returning -1, so the failure is visible in logcat/CI.

Core/Foundation/Include/Babylon/StandardStreamLogger.h
```diff
@@ -0,0 +1,28 @@
...
+    bool BABYLON_API Start();
+
+    /**
+     * Flushes pending output, restores the original streams, and stops forwarding.
+     *
+     * Returns false if an original stream could not be restored or pending output
+     * could not be drained before the shutdown timeout. Repeated calls are idempotent.
+     */
+    bool BABYLON_API Stop();
+
```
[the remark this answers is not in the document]
bkaradzic-microsoft: Documented. IsStarted() is the logical Start/Stop flag under the mutex — not a live fd probe. On plain Linux (and other hosts that leave streams alone) Start() still succeeds and IsStarted() is true with no redirection installed.

Core/Foundation/Include/Babylon/StandardStreamLogger.h
```diff
@@ -0,0 +1,37 @@
...
+         * Returns whether Start() has successfully begun process-wide forwarding and
+         * Stop() has not yet completed.
+         *
+         * This is the logical started flag, not a live probe of the underlying file
+         * descriptors. On platforms that leave stdout/stderr unchanged (plain Linux
+         * and other non-Android Unix hosts), Start() still succeeds and IsStarted()
+         * reports true even though no redirection was installed.
+         */
+        bool BABYLON_API IsStarted();
+    }
```
ryantrem: Does this have extra indentation?

Core/Foundation/Source/StandardStreamLogger.cpp
ryantrem, line 1: In many other cases, we have a shared cpp (if there is enough sharing across platforms) plus platform specific files, rather than lots of ifdefs in the code. Was it an explicit choice to deviate and go with inline ifdefs for this case? If yes it would be good to comment on why, otherwise I'd say align with existing patterns?

Core/Foundation/Include/Babylon/StandardStreamLogger.h
bkaradzic-microsoft: Yep — leftover over-indent from the IsStarted() doc comment. Fixed in ac36b14 so it matches Start()/Stop().

Core/Foundation/Source/StandardStreamLogger.cpp
bkaradzic-microsoft: Not an intentional deviation — just the smallest first cut. Aligned with the AppRuntime pattern in ac36b14:

- `StandardStreamLogger.cpp` — shared Start/Stop/IsStarted + process-wide mutex
- `StandardStreamLogger_Windows.cpp` — Win32/UWP (`OutputDebugString` + CRT/`CreatePipe`)
- `StandardStreamLogger_Android.cpp` / `_Apple.cpp` — platform sink + shared POSIX tee body via `StandardStreamLogger_Posix.inl`
- `StandardStreamLogger_Unix.cpp` — successful no-op (streams already reach the environment)

CMake picks the platform TU from `JSRUNTIMEHOST_PLATFORM` the same way AppRuntime does.

Core/Foundation/Source/StandardStreamLogger_Windows.cpp
ryantrem, line 1: It looks like there is still a fair bit of duplication between Windows and Posix. Could it make sense to share more of the code via lower level platform abstractions?

bkaradzic-microsoft: Agreed — the Windows/Posix bodies were still mostly the same tee loop. In the latest commit the shared path is `StandardStreamLogger_Shared.inl` (Drain / StartChannel / StopChannel / outstanding-drain tracking), and each platform TU only supplies lower-level ops:

- fd/pipe/null helpers (`OsDuplicate`, `OsCreatePipe`, `OsOccupyTarget`, …)
- diagnostic sink (`OsWritePlatform`)
- optional handle hooks (`OsOnStartChannel` / `OsOnRedirected` / `OsOnRestore`) — real work on Windows for `GetStdHandle`/`SetStdHandle`, no-ops on POSIX

Android and Apple also share the POSIX fd helpers via `StandardStreamLogger_PosixOps.inl`.

Core/Foundation/CMakeLists.txt
```diff
@@ -2,19 +2,44 @@ set(SOURCES
...
+    "Source/StandardStreamLogger_Shared.inl")
+
+# Shared tee body (StandardStreamLogger_Shared.inl) + thin OS-ops platform TUs.
+# Match AppRuntime: one JSRUNTIMEHOST_PLATFORM implementation is compiled in.
+if(JSRUNTIMEHOST_PLATFORM STREQUAL "Win32" OR JSRUNTIMEHOST_PLATFORM STREQUAL "UWP")
+    list(APPEND SOURCES "Source/StandardStreamLogger_Windows.cpp")
+elseif(JSRUNTIMEHOST_PLATFORM STREQUAL "Android")
+    list(APPEND SOURCES
+        "Source/StandardStreamLogger_Android.cpp"
+        "Source/StandardStreamLogger_PosixOps.inl")
```
ryantrem: Any reason Android.cpp (and other variants) don't #include PosixOps.inl (similar to how they already do this for the Shared.inl)? Can you either change that for consistency, or add a comment explaining why not?

bkaradzic-microsoft: They already do — `StandardStreamLogger_Android.cpp` / `_Apple.cpp` both `#include "StandardStreamLogger_PosixOps.inl"` (inside the anonymous namespace, same pattern as `#include "StandardStreamLogger_Shared.inl"` at the bottom).

What you were looking at in CMake was only an IDE/`source_group` listing so the .inl shows up next to the .cpp in the project view — it is not a separate TU. That was easy to misread, so I tightened CMake to:

- keep the .inl files in `SOURCES` purely for IDE visibility
- set `HEADER_FILE_ONLY` on them
- comment that the platform .cpp files are what actually `#include` them

ryantrem (approved):

Core/Foundation/Source/StandardStreamLogger_Shared.inl
```diff
@@ -0,0 +1,367 @@
...
+    {
+        channel.Target = target;
+        if (!OsOnStartChannel(channel.Platform, target, stream == Stream::Error))
+        {
+            channel = {};
+            return false;
+        }
+
+        errno = 0;
+        channel.Original = OsDuplicate(target);
```
bghgary, line 185: Please make each platform's `OsDuplicate()` return a non-inheritable descriptor. POSIX `dup()` clears `FD_CLOEXEC`, so hidden copies can survive `exec`; Windows `_dup()` creates an inheritable handle, so inheriting process creation can retain them after the logger releases its copies.

Core/Foundation/Source/StandardStreamLogger_PosixOps.inl
```diff
@@ -0,0 +1,83 @@
...
+}
+
+int64_t OsWrite(int fd, const void* data, size_t size)
+{
+    return ::write(fd, data, size);
+}
+
+int OsCreatePipe(int fds[2])
+{
+    if (::pipe(fds) != 0)
```
bghgary, line 35: Please use atomic close-on-exec pipe creation where the platform supports it, such as `pipe2(..., O_CLOEXEC)` on Android. `pipe()` followed by `fcntl()` still leaves a concurrent fork/exec window; Darwin lacks `pipe2`, so that fallback remains racy.

Core/Foundation/Source/StandardStreamLogger_Shared.inl
```diff
@@ -0,0 +1,367 @@
...
+            }
+
+            pending.append(buffer.data(), size);
+
+            // Consume complete lines via a start index so we only memmove once
+            // per read batch instead of on every newline.
+            size_t start = 0;
+            for (;;)
+            {
+                const size_t newline = pending.find('\n', start);
```
bghgary, line 119: Please apply `MAX_PLATFORM_LINE_SIZE` before emitting complete lines too. A line completed in the latest 1,024-byte read can exceed 3,800 bytes, and this branch sends it to the platform sink unchunked.

Core/Foundation/Source/StandardStreamLogger_Shared.inl
```diff
@@ -0,0 +1,367 @@
...
+        }
+        return true;
+    }
+
+    bool StopChannel(Channel& channel)
+    {
+        bool restored{true};
+        if (channel.Original >= 0)
+        {
+            restored = OsDuplicateTo(channel.Original, channel.Target) == 0;
```
bghgary, line 269: Please preserve and restore the target descriptor's original inheritance flag. POSIX `dup2()` clears `FD_CLOEXEC`, and Windows `_dup2()` creates an inheritable target, so an originally non-inheritable stdout or stderr can be left inheritable after `Stop()`.

Core/Foundation/Source/StandardStreamLogger_Apple.cpp
```diff
@@ -0,0 +1,23 @@
...
+
+namespace
+{
+// POSIX fd helpers (dup/pipe/CLOEXEC/devnull); sink is OsWritePlatform below.
+#include "StandardStreamLogger_PosixOps.inl"
+
+    void OsWritePlatform(bool isError, const std::string& line)
+    {
+        const os_log_type_t type = isError ? OS_LOG_TYPE_ERROR : OS_LOG_TYPE_DEFAULT;
+        os_log_with_type(OS_LOG_DEFAULT, type, "%{public}s", line.c_str());
```
bghgary, line 19: Please chunk Apple's dynamic `%s` content to the selected log type's persisted limit before calling `os_log_with_type()`: 1,024 bytes for `OS_LOG_TYPE_DEFAULT` and 256 for `OS_LOG_TYPE_ERROR`. The persisted copy of the shared 3,800-byte chunks is truncated.

bghgary (commented):

Concerns inline.

Core/Foundation/Source/StandardStreamLogger_Shared.inl
bkaradzic-microsoft: Addressed in https://github.com/BabylonJS/JsRuntimeHost/pull/236. Each channel snapshots the original descriptor's inheritance flags and reapplies them both after redirection and during restoration, including rollback. Windows also preserves the existing Win32 standard-handle restoration behavior and handles unavailable GUI CRT streams (`-2`) as unavailable descriptors. Lifecycle coverage checks both inheritable and non-inheritable targets while forwarding and after `Stop()`.

Core/Foundation/Source/StandardStreamLogger_Shared.inl
bkaradzic-microsoft: Addressed in https://github.com/BabylonJS/JsRuntimeHost/pull/236. The shared emitter now bounds complete lines before emission as well as unterminated output and EOF flushes. The 4,095-byte line completed by the fourth 1,024-byte read is covered, along with exact-limit/CRLF boundaries, blank lines, and UTF-8 boundaries. Chunking does not alter the original-stream tee.

Core/Foundation/Source/StandardStreamLogger_Shared.inl
bkaradzic-microsoft: Addressed in https://github.com/BabylonJS/JsRuntimeHost/pull/236. POSIX private duplicates now use `F_DUPFD_CLOEXEC`. Windows clears kernel handle inheritance on `_dup` results and marks pipe/NUL descriptors non-inheritable, while retaining CRT text/binary/Unicode and other descriptor state.

The Windows `_dup`/flag-update sequence still has a concurrent process-creation window; the API now explicitly requires serialization of child creation with `Start()`/`Stop()`. This limitation is documented rather than replacing CRT descriptors with incomplete state.

Core/Foundation/Source/StandardStreamLogger_PosixOps.inl
bkaradzic-microsoft: Addressed in https://github.com/BabylonJS/JsRuntimeHost/pull/236: Android now uses `pipe2(O_CLOEXEC)`. Darwin retains the checked `pipe()`/`fcntl()` fallback because it has no `pipe2`; its remaining race is explicitly documented, with callers required to serialize concurrent child-process creation with `Start()`/`Stop()`.

Core/Foundation/Source/StandardStreamLogger_Apple.cpp
bkaradzic-microsoft: Addressed in https://github.com/BabylonJS/JsRuntimeHost/pull/236. Apple now selects a per-log-type limit: 1,023 content bytes for default logs and 255 for error logs, reserving the string terminator within the documented 1,024/256-byte persisted dynamic-content budgets. Those limits apply to complete lines, partial chunks, and EOF output, with UTF-8 boundaries preserved. Windows and Android retain the existing 3,800-byte limit.

PR #236, opened 2026-09-10
Fix standard-stream inheritance and platform log chunking

bkaradzic-microsoft:
Follow-up to #233, addressing [Gary's original review](https://github.com/BabylonJS/JsRuntimeHost/pull/233#pullrequestreview-5044014805) and [the follow-up review](https://github.com/BabylonJS/JsRuntimeHost/pull/236#pullrequestreview-5173940758).

## Changes

- Make persistent private saved/reader descriptors non-inheritable: `F_DUPFD_CLOEXEC` on POSIX; non-inheritable `DuplicateHandle` plus `_open_osfhandle(_O_BINARY | _O_NOINHERIT)` on Windows, excluding private entries from CRT `_spawn*` serialization as well as kernel inheritance.
- Use `pipe2(O_CLOEXEC)` on Android. Keep the checked `pipe`/`fcntl` fallback on Apple, which has no `pipe2`.
- Save and restore target inheritance flags during redirection, Stop, and rollback. Preserve Windows standard-handle behavior, including unavailable GUI descriptors (`_get_osfhandle` returning `-2`).
- Preserve Windows CRT translation modes for targets with valid handles during active redirection and restoration. Tee already-translated bytes through binary private descriptors, without a second translation. Decode UTF-16 only for platform diagnostics, retaining incomplete code units/surrogate pairs between reads and replacing malformed input.
- Enforce byte limits for complete and unterminated lines: 3,800 bytes on Windows/Android, 1,023 bytes for Apple default output, and 255 bytes for Apple error output. Preserve UTF-8 boundaries, CRLF, and intentional blank lines; platform chunking does not alter original tee bytes.
- Add lifecycle, byte-preservation, line-boundary, UTF-8/UTF-16, POSIX flag, Windows mode/restoration, and CRT child-startup metadata regressions. Windows child probes run before GTest initialization and are excluded from UWP.

## Windows CRT-state limitation

**CRT append semantics are not preserved.** Private Windows descriptors are reconstructed as binary, non-inheritable descriptors. Translation modes are restored explicitly, but other CRT-only flags, including `_O_APPEND`, are not retained in those private copies or restored targets. This affects both tee writes and writes after Stop(); applications requiring CRT append semantics must not use Windows forwarding. This replaces the previous description's overly broad claim of full CRT-state preservation.

## Process-creation limitation

Applications must serialize child-process creation with `Start()`/`Stop()`. Apple's `pipe`/`fcntl` fallback, Windows' short-lived CRT mode probe, and descriptor replacement followed by inheritance-flag restoration cannot be atomic with child creation. Persistent private descriptors are non-inheritable while forwarding is active.

## Validation

- Windows x64 QuickJS RelWithDebInfo: 16/16 focused logger tests passed across three shuffled runs.
- Both new Windows regressions fail against the previous head, `9b44d12`: active modes become binary, and private descriptors are serialized in CRT child-startup metadata.
- CI for `04b808f546cca3256d3b03f886329ad5dc891041` is fully green: **25/25 checks passed** ([run 35155569100, attempt 2](https://github.com/BabylonJS/JsRuntimeHost/actions/runs/35155569100/attempts/2)). The initial Android QuickJS failure was an Android Platform 33 SDK archive installation error before compilation; the retry passed without source changes.


Core/Foundation/Source/StandardStreamLogger_PosixOps.inl
```diff
@@ -32,6 +34,11 @@ int64_t OsWrite(int fd, const void* data, size_t size)
 
 int OsCreatePipe(int fds[2])
 {
+#if defined(__linux__)
+    return ::pipe2(fds, O_CLOEXEC);
```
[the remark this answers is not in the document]
bkaradzic-microsoft: Addressed in 9b44d12c03291abc0b0796079899197fe78c10ea, now pushed. `pipe2(O_CLOEXEC)` is gated on `__ANDROID__`; desktop POSIX tests use the checked `pipe`/`fcntl` fallback, without adding `_GNU_SOURCE` to the test target. Android retains its atomic close-on-exec path.

The local GCC toolchain predefines `_GNU_SOURCE`, so its default C++20 mode did not reproduce the missing declaration. I additionally compiled the desktop POSIX test with `pipe2` explicitly forbidden: it failed before this change and passes afterward. The POSIX descriptor/line tests pass, and the Android implementation still compiles with NDK Clang.

Also rebased the PR onto current `main` (`f47991dd5729fac1d6d477e66ed8a843da1d4832`), retaining both the incoming scheduler tests and this PR's stream-logger tests in the desktop and Android CMake source lists.

Core/Foundation/Source/StandardStreamLogger_Windows.cpp
```diff
@@ -68,18 +158,19 @@ namespace
...
+        fds[0] = ::_open_osfhandle(reinterpret_cast<intptr_t>(readHandle), _O_BINARY | _O_NOINHERIT);
         if (fds[0] < 0)
         {
             (void)::CloseHandle(readHandle);
             (void)::CloseHandle(writeHandle);
             return -1;
         }
 
-        fds[1] = ::_open_osfhandle(reinterpret_cast<intptr_t>(writeHandle), _O_BINARY);
+        fds[1] = ::_open_osfhandle(reinterpret_cast<intptr_t>(writeHandle), _O_BINARY | _O_NOINHERIT);
```
bghgary, line 204: Please preserve the target descriptor's CRT mode while it is redirected. Because `_dup2` copies the pipe's `_O_BINARY` mode onto stdout/stderr, `_O_U8TEXT` wide writes reach the pipe as UTF-16LE rather than the expected UTF-8/CRLF, so platform logging parses the wrong bytes. Preserve the mode on the redirected target, avoid applying its translation again when teeing those bytes to the original handle, and add Windows coverage for active text/Unicode modes and restoration.

Core/Foundation/Source/StandardStreamLogger_Windows.cpp
```diff
@@ -25,14 +29,100 @@ namespace
...
+        // text mode), unlike rebuilding it with _open_osfhandle.
         const auto previousHandler = ::_set_thread_local_invalid_parameter_handler(IgnoreInvalidParameter);
         const int duplicated = ::_dup(fd);
         (void)::_set_thread_local_invalid_parameter_handler(previousHandler);
+
+        if (duplicated < 0)
+        {
+            return -1;
+        }
+        if (!SetDescriptorInheritance(duplicated, false))
```
bghgary: Please also keep these private duplicates non-inheritable in the CRT descriptor table. `_dup` clears UCRT's `FNOINHERIT` bit, so clearing only `HANDLE_FLAG_INHERIT` still makes `_spawn*` serialize the descriptors; children can then see stale open entries whose handles were not inherited. Add a `_spawn*` regression test.

bghgary (commented):

Concerns inline.

Core/Foundation/Source/StandardStreamLogger_Windows.cpp
bkaradzic-microsoft: Fixed in 04b808f546cca3256d3b03f886329ad5dc891041. The original CRT translation mode is captured without changing the live target, then applied to the private source before `_dup2`, so it is installed together with the target handle during both redirection and restoration. Reader-side duplicates stay binary: already-translated bytes are teed unchanged, without a second newline/Unicode conversion. UTF-16 diagnostic bytes are decoded to UTF-8 before bounded line splitting, including split surrogate pairs and truncated input.

The new Windows regression checks active/restored modes and exact before/during/after bytes for stdout and stderr in `_O_BINARY`, `_O_TEXT`, `_O_U8TEXT`, `_O_U16TEXT`, and `_O_WTEXT`. It fails against the previous head and passes with this change. The complete focused logger suite passes 16/16 across three shuffled runs.

One explicit limitation is now documented in the API and PR description: reconstructing the private CRT descriptors does not preserve other CRT flags such as `_O_APPEND`, either for teeing or for the restored target. Translation-mode preservation is not a claim to preserve all CRT state.

Core/Foundation/Source/StandardStreamLogger_Windows.cpp
bkaradzic-microsoft: Fixed in 04b808f546cca3256d3b03f886329ad5dc891041. Long-lived saved/reader descriptors now use `DuplicateHandle(..., FALSE, ...)` followed by `_open_osfhandle(..., _O_BINARY | _O_NOINHERIT)`, so both the kernel handle and the CRT entry are non-inheritable. The remaining `_dup` is only a short-lived mode probe during Start(), covered by the documented requirement to serialize child creation with Start()/Stop().

Added a bounded `_wspawnv` regression whose child inspects the raw UCRT startup descriptor metadata before GTest initializes. It detects serialized private entries even when the child CRT would discard their non-inherited handles. The test fails against 9b44d12 and passes with the fix. The PR description also removes the obsolete claim that `_dup` preserves all CRT state and explicitly calls out the reconstructed descriptors' append-mode limitation.

Core/Foundation/Source/StandardStreamLogger_Windows.cpp
```diff
@@ -25,22 +30,138 @@ namespace
...
-    int OsDuplicateTo(int source, int target)
+    int OsDuplicateTo(int source, int target, const ChannelPlatformState* state = nullptr)
     {
+        // Set the private source's mode first so _dup2 installs the correct mode
+        // atomically with the target handle, on redirect as well as restoration.
+        if (state != nullptr && ::_setmode(source, state->OriginalMode) < 0)
+        {
+            return -1;
+        }
         return ::_dup2(source, target);
```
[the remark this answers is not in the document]
bkaradzic-microsoft: Addressed in 37cf0fa. For originally non-inheritable standard handles, redirect and restore now rebind fd 1/2 through `_open_osfhandle(..., _O_NOINHERIT)` instead of relying on `_dup2`, which clears the CRT flag. The existing kernel inheritance restoration remains in place. I also extended the `_wspawnv` probe to check fds 1 and 2 both while forwarding and after `Stop()`. Validation: macOS JavaScriptCore build and 40 native tests passed (the Windows path cannot run locally); Win32 x64 Chakra, x64 QuickJS, and x86 Chakra CI builds/tests for this commit passed.

PR #249, opened 2026-09-16
Tests: serve the network tests from a loopback HTTP server

matthargett:
## Problem

The XMLHttpRequest and fetch suites hit `https://github.com/` and `raw.githubusercontent.com` to check status codes, `statusText`, percent-encoding of non-ASCII paths and abort-in-flight behaviour. Any hiccup on those hosts fails the run — the recent upstream reruns of #116 went red on external 504s (httpbin.org, github.com) with the same commit green elsewhere — and a runner without outbound network cannot run the suite at all.

## Change

The test host starts a small loopback HTTP/1.1 server (`Tests/UnitTests/Shared/TestHttpServer.{h,cpp}`, plain sockets, Winsock on Windows) on `127.0.0.1` with an ephemeral port and hands its origin to the scripts as the `hostTestServer` global. Routes, matched after percent-decoding the request-target:

| path | response |
|---|---|
| `/` | 200 `OK` |
| `/assets/<name>` | 200 only when `<name>` decodes to `στρογγυλεμένος % κύβος.glb` — the client's own encoding is what gets exercised; anything else 404 |
| `/delay/<ms>` | 200 after the delay, so the abort tests always abort a request that is genuinely in flight |
| anything else | 404 `Not Found` |

Every response is `text/plain` with `Connection: close`. `tests.ts` keeps its assertions and just takes the URLs from a `testServer` table; the connection-refused tests still use `127.0.0.1:1`. The Android test library (`UnitTestsJNI`) links the server too. No polyfill code changes.

Verified on macOS (JavaScriptCore): 225 passing / 3 pending in the JavaScript suite, `UnitTests` 23/23. Fork twin: rebeckerspecialties/JsRuntimeHost#33.


matthargett:
Fork twin caught the Android side: UrlLib there goes through `HttpURLConnection`, and with targetSdk 33 the platform blocks cleartext `http://127.0.0.1` unless the app opts in, so the loopback requests failed as transport errors (`fetch failed`, status 0). Added `android:usesCleartextTraffic="true"` to the test app's manifest (2d32a1d); the rest of the matrix was already green.

PR #251, opened 2026-09-17
Blob/URL: identify Blobs by our own constructor; make URL's static methods writable (WebIDL)

matthargett:
## Problem

`Blob::TryGetData` (behind `URL.createObjectURL`) decided whether an object is one of our Blobs by walking its prototype chain against **the current global `Blob`**. A page or harness that installs its own `Blob` class — BabylonNative's Dawn validation harness does, for image loading — then made every instance of *that* class pass the check, and `ObjectWrap::Unwrap` on an object that wraps nothing threw a bare `Error: Invalid argument` out of `createObjectURL`; meanwhile a genuine native Blob was rejected as "not a Blob".

The same harness assigns its own `URL.createObjectURL`. Node-API's `napi_default` makes static methods read-only, so that assignment failed silently and the native method kept receiving the foreign Blob. In browsers WebIDL static operations are writable/enumerable/configurable, so the override would have worked.

Symptom in the wild: the three SOG Gaussian-splat scenes (and "Iridescence NME") of the Dawn validation catalog failing with `Unable to load … .sog: Failed to parse SOG zip data.` whose `cause` is `Invalid argument`.

## Change

- Keep the Blob constructor on the runtime's native object at `Initialize` and identify instances against **that**; a null `Unwrap` yields `nullopt` (→ the existing `TypeError: … argument is not a Blob`).
- Declare the `URL` statics (`canParse`, `parse`, `createObjectURL`, `revokeObjectURL`) with `napi_writable | napi_enumerable | napi_configurable`, matching WebIDL.
- Two regression tests in the `URL.createObjectURL` suite (both fail on main: the real Blob is rejected / the assignment is silently ignored).

Verified on macOS (JavaScriptCore): JavaScript suite green, `UnitTests` 23/23; the four Playground scenes above pass again on the BabylonNative fork with this build. Fork twin: rebeckerspecialties/JsRuntimeHost#35.


matthargett:
Twin caught a test-only issue on Win32_x86_Chakra: ChakraCore has no `globalThis`, so the look-alike test threw a ReferenceError before asserting anything. The test now reaches the global object via `Function("return this")()` (b374eb4); no polyfill change.

matthargett:
cc @CedricGuillemet since this is another issue blocking splats on WebGPU in BN

PR #254, opened 2026-09-17
Keep runtime schedulers safe through teardown

matthargett:
## Summary
- Share synchronized dispatch state so scheduler copies safely outlive their runtime.
- Close dispatch before AppRuntime discards pending work; retain finalizer cleanup and release dispatch captures outside the lock.
- Accept mutable and `Napi::Env` callbacks while preserving existing const/zero-argument overload behavior. Add ten bounded native tests with owned timeout state.

## Validation
macOS 27.0 arm64, SDK 27.0, LLVM 23.1.1, real system JavaScriptCore. A native-only target compiled the committed scheduler test file and runtime, reusing dependency sources read-only:
- `JSC_collectContinuously=1 SchedulerTests --gtest_repeat=20`: 200/200 passed.
- Same run with `ENABLE_THREAD_SANITIZER=ON`, `TSAN_OPTIONS=halt_on_error=1`: 200/200 passed, no TSan reports. System JSC itself is not instrumented.
- A bounded ASan/UBSan subprocess against pristine upstream `839b352` reproduced post-destruction dispatch failure: `mutex lock failed: Invalid argument`. No ASan UAF diagnostic was emitted.
- `git diff --check` passed. No npm/polyfill suite or other engines were tested.

## Provenance
Adapted from Matt Hargett's integration commits `7093c83d3a29020fc52f070859b7f7514ed3195f` and `0a5d570d156d75f5fc01162f5fe6a7ccc891e3c3`. This standalone commit preserves the original author/date and adds explicit host shutdown ordering, unlocked capture release, callable compatibility, and bounded tests on upstream `main`.

Note: this latent issue was discovered by the ChatGPT DayBreak model when I was doing adversarial testing on my webgpu branch.


matthargett:
Also validated this change in the integrated browser-polyfill stack on M4 Max/system JavaScriptCore with `JSC_collectContinuously=1`: all 55 native tests passed, including 362 JavaScript cases and the worker WPT subset (three existing JavaScript skips). The standalone TSan results remain recorded above.

